PR #4, opened 2016-09-17
Add try catch to runtime data decode step

AllanYangZhou:
Fixes fake-runtime crash bug.


kevinmasd (approved):
Note to self: Find out why console.log sends only to stdout and not to DevTools.

AllanYangZhou:
@kevinmasd It depends on whether you `console.log` from the main process (which goes to `stdout`) or from the renderer process (which goes to the dev tools)

PR #6, opened 2016-09-18
Update runtime.proto, previously named runtime_proto.proto

brandonxxlee:
-runtime.proto now includes TCP message to handle console outputs
-Change naming and some numbering, Dawn is notified of changes


ansible-protos/runtime.proto
```diff
@@ -5,13 +5,17 @@ message RuntimeData {
...
-
+		AUTO = 4;
 	}
 	message SensorData {
-		string type = 1;
-		float value =2;
-		string id = 3;
+		string device_type = 1;
+		string device_name = 2;
+		float value = 3;
```
nikitakit, line 13: FYI you're renumbering the fields here, in addition to renaming them.

bodiddles (approved):
Changed fields should never happen again but since no parts of the stack have actually utilized these proto's, it should be fine.

PR #14, opened 2016-09-22
Removing SSH keys 

rmandla:
change the install scripts to no longer contain setup for SSH key authentication


nikitakit (changes requested):
You want to merge into master, not out of master

PR #15, opened 2016-09-22
Removing ssh keys

rmandla:
remove ssh keys from the install files for frankfurter


DevOps/frankfurter/scripts/install/master_frank_setup.sh~
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
nikitakit: You need to remove this temp file before this can be merged

nikitakit (changes requested):

vdonato:
We probably want to go ahead and take out the gpg verification of updates while we're at it.

(Actually, nevermind let's just make that another change.)

nikitakit (approved):
LGTM now

PR #17, opened 2016-09-22
Update global gitignore

nikitakit:
I've loaded up on boilerplate gitignores.

While this many may be overkill, a central repo needs to cover a number of programming languages and text editors.


vdonato (approved):
lgtm

PR #22, opened 2016-09-24
Fix #18: Cross platform dawn start script

### LINKED ISSUE #18: Dawn start script not cross-platform, only works on Mac
In `dawn/package.json`, a start script is defined which currently only works on Mac, by calling the `Electron.app` binary on our `build/main.js` script. We should make it work at least on linux also, where the command should be very similar (without the `.app`). In general, we just need a way to change the start script being run depending on the platform.

AllanYangZhou:
See #18.


AllanYangZhou:
@kevinmasd I don't actually have a mac to test this on, so can you test this on your machine first before merging?

kevinmasd (approved):
Works on my Mac.

PR #25, opened 2016-09-24
[RUNTIME] Adds basic UDP integration to the current implementation of runtime

bodiddles:
Creates 4 processes: packager, sender, receiver, unpackager to handle
data and send over it over UDP sockets. These currently don't use
google protos but converts "TEST" into a bytearray and sends over sockets.

To handle async messages, there is a class called two_buffer which just
stores and sends data in two separate indices of a list which are atomic
operations which should be ok for async.

It uses the process factory in runtime.py to generate said processes and
I have added them into the process names enum class. I have also added
additional SM_COMMANDS SEND and RECEIVE. SEND is sent to SM when the
packager is done with packaging one set of data and can accept another.
This should be fixed in the future to constantly send over the state,
potentially using another instance of the two_buffer.
The RECEIVE command just stores whatever we send to state manager into
the state itself which will also need to fixed to handle storing gamepad
data which we expect from Dawn over UDP.

### DIFF
--- a/runtime/testy/Ansible.py
+++ b/runtime/testy/Ansible.py
@@ -10,41 +10,46 @@ sendPort = 1235
 recvPort = 1236
 
 #Custom buffer for handling states. Holds two states, updates one and sends the other.
-class two_buffer(): 
+class TwoBuffer(): 
     def __init__(self):
         self.data = [0, 0]
         self.put_index = 0
         self.get_index = 1
+
     def replace(self, item):
         self.data[self.put_index] = item
-        self.put_index = (self.put_index + 1) % 2
-        self.get_index = (self.get_index + 1) % 2
+        self.put_index = 1 - self.put_index
+        self.get_index = 1 - self.get_index
+
     def get(self):
         return self.data[self.get_index]
 
-sendBuffer = two_buffer()
-recvBuffer = two_buffer()
+sendBuffer = TwoBuffer()
+recvBuffer = TwoBuffer()
 raw_fake_data = [[0]]
 packed_fake_data = [[0]]
 dawn_buffer = [0]
+
 ###Protobuf handlers###
 #Function for handling unpackaging of protobufs from Dawn
 def unpackage(data):
     return data #will be changed when we fully support protos
+
 #Handles packaging and sending state to dawn in the form of protobuf we define
 def package(state, badThingsQueue):
-    s = "TEST"#will be changed when we fully support protos
+    s = "TEST" #will be changed when we fully support protos
     b = bytearray()
     b.extend(map(ord, s))
-    return b
+    return b 
+
 ###Start Ansible Thread Chain###
 def packageData(badThingsQueue, stateQueue, pipe):
-    while(True):
+    while True:
         try:
             rawState = pipe.recv() #Pull state from the pipe
-            if (rawState == SM_COMMANDS.READY):
+            if rawState == SM_COMMANDS.READY:
                 stateQueue.put([SM_COMMANDS.SEND, 1])
-            elif(rawState):
+            elif rawState:
                 packState = package(rawState, badThingsQueue)
                 sendBuffer.replace(pack_state) ##Used list mutation as it's an atomic operation
         except Exception:
@@ -53,46 +58,33 @@ def packageData(badThingsQueue, stateQueue, pipe):
 def udpSender(badThingsQueue, stateQueue, pipe):
     host = socket.gethostname()
     with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:#DGRAM for UDP sockets
-        while(True): #constantly send the state to Dawn
+        while True: #constantly send the state to Dawn
             try:
                 msg = sendBuffer.get()
-                if(msg != 0): #check if the data inside the buffer has changed
+                if msg != 0: #check if the data inside the buffer has changed
                     s.sendto(msg, (host, sendPort))
             except Exception:
                 badThingsQueue.put(BadThing(sys.exc_info(), None))
+
 def udpReceiver(badThingsQueue, stateQueue, pipe):
     #same thing as the client side from python docs
     host = socket.gethostname()
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM) #DGRAM for UDP sockets
     s.bind((host, recvPort))
-    while(True):
+    while True:
         try:
             data = s.recv(2048)
             recvBuffer.replace(data)
         except Exception: 
             badThingsQueue.put(BadThing(sys.exc_info(), None))
+
 def unpackageData(badThingsQueue, stateQueue, pipe):
     ready = False;
-    while(True):
+    while True:
         try:
             ready = pipe.recv()
-            if(ready):
+            if ready:
                 unpackagedData = unpackage(recvBuffer.get())
                 stateQueue.put([SM_COMMANDS.STORE, unpackagedData])
         except Exception:
             badThingsQueue.put(BadThing(sys.exc_info(), None))
-
-
-
-
-        
-
-
-    
-
-
-
-
-  
-
-
--- a/runtime/testy/Ansible.py
+++ b/runtime/testy/Ansible.py
@@ -1,0 +1,98 @@
+import socket
+import threading
+import time
+import fake_dawn
+import random
+import sys
+from runtimeUtil import *
+data = [0] #for testing purposes
+sendPort = 1235
+recvPort = 1236
+
+#Custom buffer for handling states. Holds two states, updates one and sends the other.
+class two_buffer(): 
+    def __init__(self):
+        self.data = [0, 0]
+        self.put_index = 0
+        self.get_index = 1
+    def replace(self, item):
+        self.data[self.put_index] = item
+        self.put_index = (self.put_index + 1) % 2
+        self.get_index = (self.get_index + 1) % 2
+    def get(self):
+        return self.data[self.get_index]
+
+sendBuffer = two_buffer()
+recvBuffer = two_buffer()
+raw_fake_data = [[0]]
+packed_fake_data = [[0]]
+dawn_buffer = [0]
+###Protobuf handlers###
+#Function for handling unpackaging of protobufs from Dawn
+def unpackage(data):
+    return data #will be changed when we fully support protos
+#Handles packaging and sending state to dawn in the form of protobuf we define
+def package(state, badThingsQueue):
+    s = "TEST"#will be changed when we fully support protos
+    b = bytearray()
+    b.extend(map(ord, s))
+    return b
+###Start Ansible Thread Chain###
+def packageData(badThingsQueue, stateQueue, pipe):
+    while(True):
+        try:
+            rawState = pipe.recv() #Pull state from the pipe
+            if (rawState == SM_COMMANDS.READY):
+                stateQueue.put([SM_COMMANDS.SEND, 1])
+            elif(rawState):
+                packState = package(rawState, badThingsQueue)
+                sendBuffer.replace(pack_state) ##Used list mutation as it's an atomic operation
+        except Exception:
+            badThingsQueue.put(BadThing(sys.exc_info(), None))
+
+def udpSender(badThingsQueue, stateQueue, pipe):
+    host = socket.gethostname()
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:#DGRAM for UDP sockets
+        while(True): #constantly send the state to Dawn
+            try:
+                msg = sendBuffer.get()
+                if(msg != 0): #check if the data inside the buffer has changed
+                    s.sendto(msg, (host, sendPort))
+            except Exception:
+                badThingsQueue.put(BadThing(sys.exc_info(), None))
+def udpReceiver(badThingsQueue, stateQueue, pipe):
+    #same thing as the client side from python docs
+    host = socket.gethostname()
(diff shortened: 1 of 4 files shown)


runtime/testy/runtimeUtil.py
```diff
@@ -14,6 +14,10 @@ class PROCESS_NAMES(Enum):
   STUDENT_CODE        = "studentProcess"
   STATE_MANAGER       = "stateProcess"
   RUNTIME             = "runtime"
+  UDP_PACKAGER        = "udpPackager"
+  UDP_SENDER          = "udpSender"
+  UDP_RECEIVER        = "udpReceiver"
+  UDP_UNPACKAGER       = "udpUnpackager"
```
brandonxxlee: Get rid of extra space.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
+import socket
+import threading
+import time
+import fake_dawn
+import random
+import sys
+from runtimeUtil import *
+data = [0] #for testing purposes
```
nikitakit, line 8: This variable is not used anywhere. (To use it inside a function, you should have a `global data` line at the top of the function)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+sendPort = 1235
+recvPort = 1236
+
+#Custom buffer for handling states. Holds two states, updates one and sends the other.
+class two_buffer():
```
nikitakit: Name this `TwoBuffer` so that it follows the proper naming convention for classes. (You can consult with PEP8 for style advice)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+    def replace(self, item):
+        self.data[self.put_index] = item
+        self.put_index = (self.put_index + 1) % 2
+        self.get_index = (self.get_index + 1) % 2
+    def get(self):
+        return self.data[self.get_index]
+
+sendBuffer = two_buffer()
+recvBuffer = two_buffer()
+raw_fake_data = [[0]]
```
nikitakit, line 29: This variable is not used anywhere. (It does not look like an export to outside of this module, and there is no `global raw_fake_data` declaration anywhere)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+packed_fake_data = [[0]]
```
nikitakit, line 30: Looks unused (see also above)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+dawn_buffer = [0]
```
nikitakit, line 31: Looks unused (see also above)

runtime/testy/stateManager.py
```diff
@@ -16,7 +16,6 @@ def __init__(self, badThingsQueue, inputQueue, runtimePipe):
     self.input = inputQueue
     # map process names to pipes
     self.processMapping = {PROCESS_NAMES.RUNTIME: runtimePipe}
-
   def initRobotState(self):
```
yizow, line 20: add newline between functions

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+def unpackage(data):
+    return data #will be changed when we fully support protos
+#Handles packaging and sending state to dawn in the form of protobuf we define
+def package(state, badThingsQueue):
+    s = "TEST"#will be changed when we fully support protos
+    b = bytearray()
+    b.extend(map(ord, s))
+    return b
+###Start Ansible Thread Chain###
+def packageData(badThingsQueue, stateQueue, pipe):
```
nikitakit, line 46: This deserves a docstring that explains that this is run in a thread (or is it a process?)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+            rawState = pipe.recv() #Pull state from the pipe
+            if (rawState == SM_COMMANDS.READY):
+                stateQueue.put([SM_COMMANDS.SEND, 1])
+            elif(rawState):
+                packState = package(rawState, badThingsQueue)
+                sendBuffer.replace(pack_state) ##Used list mutation as it's an atomic operation
+        except Exception:
+            badThingsQueue.put(BadThing(sys.exc_info(), None))
+
+def udpSender(badThingsQueue, stateQueue, pipe):
```
nikitakit, line 58: This deserves a docstring that explains that this is run in a thread

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+    host = socket.gethostname()
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:#DGRAM for UDP sockets
+        while(True): #constantly send the state to Dawn
+            try:
+                msg = sendBuffer.get()
+                if(msg != 0): #check if the data inside the buffer has changed
+                    s.sendto(msg, (host, sendPort))
+            except Exception:
+                badThingsQueue.put(BadThing(sys.exc_info(), None))
+def udpReceiver(badThingsQueue, stateQueue, pipe):
```
nikitakit, line 69: This deserves a docstring that explains that this is run in a thread

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+    host = socket.gethostname()
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM) #DGRAM for UDP sockets
+    s.bind((host, recvPort))
+    while(True):
+        try:
+            data = s.recv(2048)
+            recvBuffer.replace(data)
+        except Exception: 
+            badThingsQueue.put(BadThing(sys.exc_info(), None))
+def unpackageData(badThingsQueue, stateQueue, pipe):
```
nikitakit, line 81: This deserves a docstring that explains that this is run in a thread

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+    while(True):
+        try:
+            ready = pipe.recv()
+            if(ready):
+                unpackagedData = unpackage(recvBuffer.get())
+                stateQueue.put([SM_COMMANDS.STORE, unpackagedData])
+        except Exception:
+            badThingsQueue.put(BadThing(sys.exc_info(), None))
+
+
```
nikitakit: So many blank lines here seems excessive

runtime/testy/stateManager.py
```diff
@@ -16,7 +16,6 @@ def __init__(self, badThingsQueue, inputQueue, runtimePipe):
     self.input = inputQueue
     # map process names to pipes
     self.processMapping = {PROCESS_NAMES.RUNTIME: runtimePipe}
-
```
nikitakit, line 19: This irrelevant whitespace change pollutes the diff

nikitakit (changes requested):
I made a few stylistic changes in the line-by-line.

But the big issue as I see it is that the send and receive buffers are each accessed from two different processes. Are you sure this works? I don't see any cross-process synchronization primitives.

I suspect that when a TwoBuffer is called from two different process, each process actually gets a clone of the instance and they don't actually share any data!

runtime/testy/runtime.py
```diff
@@ -9,6 +9,7 @@
 import studentAPI
 
 from runtimeUtil import *
+from Ansible import *
```
yizow: Generally I think we'd like to avoid "import *". Pollutes namespace, leading to annoying bugs.
runtimeUtil should be the only one; it should also only be importing classes.

runtime/testy/Ansible.py
yizow: Namely, python classes should be CamelCase, with the first letter Capitalized.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+sendPort = 1235
+recvPort = 1236
+
+#Custom buffer for handling states. Holds two states, updates one and sends the other.
+class two_buffer(): 
+    def __init__(self):
+        self.data = [0, 0]
+        self.put_index = 0
+        self.get_index = 1
+    def replace(self, item):
```
yizow, line 19: put newlines between functions.

runtime/testy/Ansible.py
yizow: If its for testing purposes, hopefully the testing has been completed before you pull requested, and this should be removed.

Re: Nikita
you want the `global data` line only if you plan on reassigning the value of data to point to, say, a different object. However, as lists are mutable objects in python, you can change the contents of data without using global.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+        self.data[self.put_index] = item
+        self.put_index = (self.put_index + 1) % 2
+        self.get_index = (self.get_index + 1) % 2
+    def get(self):
```
yizow, line 24: add newline

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+
+#Custom buffer for handling states. Holds two states, updates one and sends the other.
+class two_buffer(): 
+    def __init__(self):
+        self.data = [0, 0]
+        self.put_index = 0
+        self.get_index = 1
+    def replace(self, item):
+        self.data[self.put_index] = item
+        self.put_index = (self.put_index + 1) % 2
```
yizow: another way of doing this when you only have two options, `self.put_index = 1 - self.put_index`

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+sendBuffer = two_buffer()
+recvBuffer = two_buffer()
+raw_fake_data = [[0]]
+packed_fake_data = [[0]]
+dawn_buffer = [0]
+###Protobuf handlers###
+#Function for handling unpackaging of protobufs from Dawn
+def unpackage(data):
+    return data #will be changed when we fully support protos
+#Handles packaging and sending state to dawn in the form of protobuf we define
```
yizow, line 38: newline between functions

runtime/testy/Ansible.py
yizow, line 46: newline between functions

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,98 @@
...
+    s = "TEST"#will be changed when we fully support protos
+    b = bytearray()
+    b.extend(map(ord, s))
+    return b
+###Start Ansible Thread Chain###
+def packageData(badThingsQueue, stateQueue, pipe):
+    while(True):
+        try:
+            rawState = pipe.recv() #Pull state from the pipe
+            if (rawState == SM_COMMANDS.READY):
```
yizow: Im working on a commit that will hide SM_COMMANDS.READY from spawned process. Continue as is, I will deal with merge conflicts.

runtime/testy/Ansible.py
yizow, line 69: newline between functions

runtime/testy/Ansible.py
yizow, line 81: newline between functions

yizow (changes requested):
I agree with nikita's assesment: using multiple processes like this means they don't share data; i.e. sendBuffer and recvBuffer are separate for each process you spawn. 
So currently, you have 16(!) total buffers. I think this is unnecessary and can be reduced. 

You probably want to move to a model where the entire ansible is a single process. This single process may be multithreaded. 

Also, I think this is written in a functional style. The rest of the code we've written so far has been object oriented i think. 
You would make an Ansible class, and it would have a start() method that we would call as a new process. 
Do you think it's worth doing it this way, for consistency?

bodiddles:
The only issue I see with having a single Ansible Object is that sending and receiving should be separate as they are definitely independent and don't require the other. Thus, if we implement something like multi-threading, I think the issue would be that we may have to kill the entire Ansible process if one part of it needs to be restarted. I think something that may work would be making sending and packaging and receiving and unpackaging their own class and probably TCP stuff in a separate class as well.

runtime/testy/stateManager.py
```diff
@@ -56,5 +56,9 @@ def start(self):
       elif cmd_type == SM_COMMANDS.HELLO:
         print("HELLO")
       # TODO: Add better error description
+      elif cm_type == SM_COMMANDS.SEND:
```
brandonxxlee, line 59: I think the command `SEND` is too generally named since it specifically refers to sending to the UDPpackager.

yizow:
I imagine that a lot of the logic, for both sending and receiving, will overlap. You can probably make a single object, then instatiate it with arguments that tell it whether its sending or receiving. 

As for restarts, ideally Ansible should _never_ need to restart, right? That ability is there only to make it harder to crash runtime, as a whole. I'm treating an Ansible process as a duplex channel. If one side (sending, or receiving) of it goes down, i think it makes sense to just pull the plug on the whole thing and setup a new process. 

Regardless, I think 4 process for the purpose of just one UDP duplex channel is definitely overkill.

bodiddles:
I don't know about the logic overlapping for sending and receiving since it is already barebones and utilizing different protos for sending and receiving means that packaging and unpackaging are intrinsically different. 
;
Obviously it shouldn't restart but if we are building for resiliency, we should have the ability to just restart one part of the channel, especially if dawn is using this as a heartbeat. I think we could have 3 different processes, one for sending, one for receiving, and one for tcp and just have packaging and upackaging as separate threads on their respective processes. I think this makes sense as it reduces the number of processes but still allows for individual restarts if something unseen does go wrong. 

In terms of classes, there aren't any issues with separate classes for each process as each one only interacts with its own set of data.

PR #27, opened 2016-09-24
[RUNTIME] stateManager bugfix

yizow:
stateManager was responding on a pipe to studentAPI, when the studentAPI
was not expecting it. This caused a backlog of read/writes on the pipes,
leading to weird behaviour.

This commit fixes this bug by making studentAPI read from the pipe and
return the value, just like stateManager is doing.


greenkee (approved):
Should fix the clogged pipe

nikitakit:
LGTM as a bugfix, but the presence of this bug in the first place suggests that you're relying on the request-response across two queues to not go out of sync. This seems prone to introducing bugs like this one in the future.

nikitakit (approved):

PR #28, opened 2016-09-25
[RUNTIME] Implement getters and setters to access nested dictionaries in StateManager.state

jiangdaniel:
- Change Robot.getValue and Robot.setValue to accept a variable number of arguments
- Change StateManager.getValue and StateManager.setValue to be able to access nested dictionaries in StateManager.state
- Add StateManager.dictErrorMessage to create informative error messages when StateManager.getValue and StateManager.setValue error


runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
     self.processMapping[processName] = pipe
     pipe.send(RUNTIME_CONFIG.PIPE_READY.value)
 
-  def getValue(self, key):
-    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state[key])
+  def getValue(self, keys):
+    result = self.state
+    i = 0
```
yizow: The pythonic way of doing iterations like this is to use a for loop rather than a while loop. 
Cleaner, more readable, less likely to have an error when you forget `i+=1`

In fact, in general it's better to use a for loop over a while loop when possible. 

for key in keys: 
  result = result[key]

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+    try:
+      while (i < len(keys)):
+        result = result[keys[i]]
+        i += 1
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(result)
+    except:
+      self.badThingsQueue.put(BadThing(sys.exc_info(), self.dictErrorMessage(i, keys, result), event = BAD_EVENTS.BAD_EVENT, printStackTrace = False))
```
yizow: Probably should create a new BAD_EVENTS in runtimeUtil 
"stateManager key error" perhaps?

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+    currDict = self.state
+    i = 0
+    try:
+      while (i < len(keys) - 1):
+        currDict = currDict[keys[i]]
+        i += 1
+      currDict[keys[i]] = value
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(value)
+    except:
+      self.badThingsQueue.put(BadThing(sys.exc_info(), self.dictErrorMessage(i, keys, currDict), event = BAD_EVENTS.BAD_EVENT, printStackTrace = False))
```
yizow: same as above. New event type?

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+        result = result[keys[i]]
+        i += 1
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(result)
+    except:
+      self.badThingsQueue.put(BadThing(sys.exc_info(), self.dictErrorMessage(i, keys, result), event = BAD_EVENTS.BAD_EVENT, printStackTrace = False))
+
+  def setValue(self, keys, value):
+    currDict = self.state
+    i = 0
+    try:
```
yizow, line 58: Same as above. Use for loop instead.

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
     self.processMapping[processName] = pipe
     pipe.send(RUNTIME_CONFIG.PIPE_READY.value)
 
-  def getValue(self, key):
-    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state[key])
+  def getValue(self, keys):
```
yizow, line 45: I think we should put the `*args` logic in these functions inside statemanger. 
That way, the code in the `start()` function will be very clean. Since there are always going to be 2 element lists, we can just do 
cmd_type, args = received
if cmd_type == thing:
  cmd_function(args)

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+    i = 0
+    try:
+      while (i < len(keys)):
+        result = result[keys[i]]
+        i += 1
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(result)
+    except:
+      self.badThingsQueue.put(BadThing(sys.exc_info(), self.dictErrorMessage(i, keys, result), event = BAD_EVENTS.BAD_EVENT, printStackTrace = False))
+
+  def setValue(self, keys, value):
```
yizow: See above regarding putting *args logic here.
keys  = args[:-1]
value = args[-1]

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+    try:
+      while (i < len(keys) - 1):
+        currDict = currDict[keys[i]]
+        i += 1
+      currDict[keys[i]] = value
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(value)
+    except:
+      self.badThingsQueue.put(BadThing(sys.exc_info(), self.dictErrorMessage(i, keys, currDict), event = BAD_EVENTS.BAD_EVENT, printStackTrace = False))
+
+  def dictErrorMessage(self, erroredIndex, keys, currDict):
```
yizow, line 69: The fact that you wrote this function for proper error messages is amazing. I love it.

runtime/testy/stateManager.py
```diff
@@ -32,12 +32,52 @@ def addPipe(self, processName, pipe):
...
+    keyChain = ""
+    i = 0
```
yizow, line 71: for loop > while loop
enumerate(iterable) if you need numberings.

runtime/testy/studentAPI.py
```diff
@@ -7,16 +7,17 @@ def __init__(self, toManager, fromManager):
     # Wait for ack from SM before continuing
     self.fromManager.recv()
 
-  def getValue(self, key):
+  def getValue(self, *args):
     """Returns the value associated with key
     """
-    # TODO: Actually use key
-    self.toManager.put([SM_COMMANDS.GET_VAL, key])
+    keys = args
```
yizow: line unnecessary. Just pass in args.

runtime/testy/studentAPI.py
```diff
@@ -7,16 +7,17 @@ def __init__(self, toManager, fromManager):
...
+    self.toManager.put([SM_COMMANDS.GET_VAL, keys])
     return self.fromManager.recv()
 
-  def setValue(self, key, value):
+  def setValue(self, *args):
     """Sets the value associated with key
     """
-    # TODO: Implement
-    self.toManager.put([SM_COMMANDS.SET_VAL, key, value])
+    value = args[-1] if len(args) > 0 else None
```
yizow: Should do a sanity check to make sure len(args) > 1

yizow (changes requested):
Good job catching the [RUNTIME] before I did. :+1: 

Did you test that this works? I see that you haven't modified studentCode.py at all.

runtime/testy/stateManager.py
jiangdaniel: Having the index i is nice because it gets passed to dictErrorMessage in order to provide an accurate error message

jiangdaniel: I just noticed your comment about the enumerate function.  Did not know that that function existed.  I'll try it out

runtime/testy/studentAPI.py
jiangdaniel: What do you think should be done if len(args) <= 1

jiangdaniel:
@yizow I did test my code using studentCode.py.  I just didn't include the changes since the existing code in studentCode.py still worked

runtime/testy/studentAPI.py
yizow: In theory everything making it to stateManager should be parsed already.
Since this is studentAPI, we should throw an exception. The exception is being worked on by Carlos, Justin, Legeng.

One could probably also rewrite the argument list to be (value, key1, [additional keys])

yizow (approved):

PR #44, opened 2016-09-29
[RUNTIME] Changed stateManager expected input format

Angelina-Wang:
- stateManager's input is now a list of length 2, the first being the command and the second being the list of arguments
- a dictionary maps the commands to the functions


runtime/testy/stateManager.py
```diff
@@ -29,6 +39,7 @@ def initRobotState(self):
     }
 
   def addPipe(self, processName, pipe):
+    #print("add")
```
yizow: delete

runtime/testy/stateManager.py
```diff
@@ -44,17 +55,15 @@ def start(self):
...
-        self.getValue(request[1])
-      elif cmd_type == SM_COMMANDS.SET_VAL:
-        self.setValue(request[1], request[2])
-      elif cmd_type == SM_COMMANDS.HELLO:
-        print("HELLO")
-      # TODO: Add better error description
-      else:
+
+      if(len(request) != 2):
+        self.badThingsQueue.put(BadThing(sys.exc_info(), "Wrong input size, need list of size 2" % (request,), event = BAD_EVENTS.UNKNOWN_PROCESS, printStackTrace = False))
```
yizow: I don't think you tested this line. 
`request = "abc";
"Wrong input size, need list of size 2" % (request,)`
errors out because request is not used inside the string format expression. Probably add a `\n%s`

Side note: "code coverage" is the vocab word that describes this

runtime/testy/stateManager.py
```diff
@@ -44,17 +55,15 @@ def start(self):
     # And that there are the correct number of elements
     while True:
       request = self.input.get(block=True)
-      cmd_type = request[0]
```
yizow, line 47: add this back in, along with `arguments = request[1]`. 

We should probably also rename it to python style `cmdType`

Makes the code a lot more readable and clear. Self-documenting code best code

runtime/testy/stateManager.py
```diff
@@ -44,17 +55,15 @@ def start(self):
...
-      elif cmd_type == SM_COMMANDS.SET_VAL:
-        self.setValue(request[1], request[2])
-      elif cmd_type == SM_COMMANDS.HELLO:
-        print("HELLO")
-      # TODO: Add better error description
-      else:
+
+      if(len(request) != 2):
+        self.badThingsQueue.put(BadThing(sys.exc_info(), "Wrong input size, need list of size 2" % (request,), event = BAD_EVENTS.UNKNOWN_PROCESS, printStackTrace = False))
+      elif(not isinstance(request[0], SM_COMMANDS)):
```
yizow: request[0] should be cmdType

question. why don't you just use `if cmdType not in self.commandMapping`?

yizow (changes requested):
This looks really good, really clean. Good job! (Have you had a software engineering internship before?)

Minor issues, should be pretty quick to fix. Readability is important. Remember, the amount of time people spend reading your code >> 10x the amount of time you spend writing it.

yizow (approved):

PR #45, opened 2016-09-29
[RUNTIME] Add StudentAPI exceptions class.

cjsanchez:
- Adds exceptions class to studentAPI.py
- Raises exception when key for getValue doesn't exist.
- Catch statement in runtime.py specifically for student exceptions

Some questions I had:
- are new keys allowed to be added to the stateManager dictionary?
- do we just want one except statement to catch all studentAPI errors?
- should the exception class be renamed to StudentAPIError?


runtime/testy/stateManager.py
```diff
@@ -37,6 +40,7 @@ def getValue(self, key):
 
   def setValue(self, key, value):
     self.state[key] = value
+    print("NEW STATE:", self.state)
```
yizow: remove. We want to avoid debugging print statements in code we push to master because
1. clutter
2. it should be debugged before we push

runtime/testy/stateManager.py
```diff
@@ -28,6 +28,9 @@ def initRobotState(self):
      "string1" : "abcde"
     }
 
+  def getRobotState(self):
+    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state)
```
yizow: oooh, this is dangerous. self.state may contain info we dont want students to access. 

Was there a reason to make this new function? why not just use the getValue function?

runtime/testy/stateManager.py
```diff
@@ -53,6 +57,8 @@ def start(self):
         self.getValue(request[1])
       elif cmd_type == SM_COMMANDS.SET_VAL:
         self.setValue(request[1], request[2])
+      elif cmd_type == SM_COMMANDS.GET_STATE:
```
yizow: See above about whether or not we want to remove this function

runtime/testy/studentAPI.py
```diff
@@ -7,16 +7,29 @@ def __init__(self, toManager, fromManager):
     # Wait for ack from SM before continuing
     self.fromManager.recv()
 
+  def getState(self):
```
yizow: See above for whether or not we want to remove this function.

runtime/testy/studentAPI.py
```diff
@@ -7,16 +7,29 @@ def __init__(self, toManager, fromManager):
...
+    self.toManager.put([SM_COMMANDS.GET_STATE])
+    return self.fromManager.recv()
+
   def getValue(self, key):
     """Returns the value associated with key
     """
     # TODO: Actually use key
+    if key not in self.getState().keys():
```
yizow: For debugging purposes, you can just raise the error without any if statements.

yizow (changes requested):
Pretty good. 

You should think of running `git rebase -i` and squashing a lot of these commits before you push. it makes it a lot easier on me.

Before i accept pull request, i will also need a good, consolidated commit message about it. 

Regarding your questions: 
yes, we probably need a new command to add keys. you should add this to the todo list on slack (`/todo help')
Yes, we probably want just one except. Thats why we have the class inheritance; we just need to catch the parent exception
Yes. renaming is probably gonna be good.

yizow:
Don't forget to run rebase so that merging is no longer blocked. 
you're in a race with angelina and daniel. whoever get their pull request approved first wont need to do merge conflicts : P

cjsanchez:
Implemented the changes requested. Just waiting for master to be fixed.

runtime/testy/stateManager.py
```diff
@@ -38,16 +38,29 @@ def initRobotState(self):
...
+  def getRobotState(self):
+    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state)
+
   def addPipe(self, processName, pipe):
     self.processMapping[processName] = pipe
     pipe.send(RUNTIME_CONFIG.PIPE_READY.value)
 
   def getValue(self, key):
-    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state[key])
+    message = ""
```
yizow: unnecessary line

runtime/testy/stateManager.py
```diff
@@ -38,16 +38,29 @@ def initRobotState(self):
...
+    if key in self.state.keys():
+      message = self.state[key]
+    else:
+      message = StudentAPIValueError("{} is not a valid key.".format(key))
+    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(message)
 
   def setValue(self, key, value):
-    self.state[key] = value
-    self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(self.state[key])
+    message = ""
```
yizow: unncessary line

runtime/testy/studentAPI.py
```diff
@@ -9,12 +9,18 @@ def getValue(self, key):
     """Returns the value associated with key
     """
     # TODO: Actually use key
-    self.toManager.put([SM_COMMANDS.GET_VAL, [key]])
-    return self.fromManager.recv()
+    self.toManager.put([SM_COMMANDS.GET_VAL, key])
```
yizow: should be using the new paradigm of [SM_COMMAND, [command arguments]]

runtime/testy/studentAPI.py
```diff
@@ -9,12 +9,18 @@ def getValue(self, key):
...
+    message = self.fromManager.recv()
+    if type(message) is StudentAPIValueError:
```
yizow: is type() neccessary? I think there is a python function that checks is Object is an instance of Class
i think it might be simply
`if Object is Class:`

runtime/testy/studentAPI.py
```diff
@@ -9,12 +9,18 @@ def getValue(self, key):
...
+    return message
 
   def setValue(self, key, value):
     """Sets the value associated with key
     """
-    # TODO: Implement
-    self.toManager.put([SM_COMMANDS.SET_VAL, [key, value]])
-    return self.fromManager.recv()
+    #statemanager passes exception, then check to see if returned value is exception or not
+    self.toManager.put([SM_COMMANDS.SET_VAL, key, value])
```
yizow: Follow new paradigm. `[SM_COMMANDS.SET_VAL, [key, value]])`

runtime/testy/studentAPI.py
```diff
@@ -9,12 +9,18 @@ def getValue(self, key):
...
+    message = self.fromManager.recv()
+    if type(message) is StudentAPIValueError:
```
yizow: See above. Should be a marginally cleaner way of doing it.

yizow (changes requested):

yizow:
Still one unanswered comment from previous review. 
why do we have a getstate function? or, what does it do that getValue doesnt?

runtime/testy/runtimeUtil.py
```diff
@@ -4,10 +4,11 @@
 
 @unique
 class BAD_EVENTS(Enum):
-  BAD_EVENT               = "BAD THINGS HAPPENED"
-  STUDENT_CODE_ERROR      = "Student Code Crashed"
-  STUDENT_CODE_TIMEOUT    = "Student Code Timed Out"
-  UNKNOWN_PROCESS         = "Unknown State Manager process name"
+  BAD_EVENT             = "BAD THINGS HAPPENED"
```
yizow: Add whitespace to align up the equal signs

yizow (changes requested):
One last thing: 
Have you tested that the errors get raised correctly? 

I have a suspicion that studentCode might fail if they dont import studentAPIError. But maybe that gets bundled into Robot... 
You should test this :+1:

yizow (approved):

PR #46, opened 2016-10-01
Fix #34: Dawn/ansiblechanges

### LINKED ISSUE #34: Remove old Ansible.js
New ansible in `main/` folder is being written, we should remove the old `Ansible.js` in the `renderer/utils` folder. There may still be files importing the older Ansible file, they should be fixed first so nothing breaks when you remove it.

kevinmasd:
#34

Removes old Ansible.js from environment; comments out old Ansible commands to keep copy for future reference.


dawn/renderer/components/Editor.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ConsoleOutput from './ConsoleOutput';
-import Ansible from '../utils/Ansible';
+// import Ansible from '../../main/ansible/Ansible';
```
AllanYangZhou: I think you should just delete the line completely, rather than commenting it out. Committing commented out lines leads to clutter. If we need to go back, we can just use git to revert.

Same for all the commented out stuff below.

AllanYangZhou (changes requested):
Also I think we should just delete Ansible.js, not rename it. Again, if we need to recover it we can just use git.

kevinmasd:
Commented out.

AllanYangZhou (approved):

PR #51, opened 2016-10-08
[RUNTIME] Testing framework for studentCode.py

yizow:
This commit introduces a testing framework for functions inside
studentCode.py. Suppose we wish to write a test called 'testName':
1. In studentCode.py, implement {testName}_setup and {testName}_main
2. Runtime will run the setup function once, then the main function
   three times.
3. Runtime will kill the studentCode process, then repeat step 2 two
   more times.
4. Save the desired output of this series of events in a text file
   test_outputs/{testName}_output

You have now added a test to runtime. Congratulations!

Sample tests "test0" has been provided. It tests nothing other than that
the system works.

"mainTest" has been the most integrated test we have been running up
until now. Initial values in stateManager have been tweaked to allow
this test to actually check division by 0 errors.

A new function testTime() has been added to runtime.py. By default,
runtime will now execute this function. It searches for a list of tests
that follow the naming convention defined above, then executes as
described above.

At the end of this series of tests, helpful messages will be printed
detailing how many/ which tests failed.

A new bad event was added to signify the "graceful exit" of studentCode.
While in production this should never happen, this was necessary to
trigger runtime to restart studentCode in our testing environment.

Some minor refactoring:
1. stateManager.py: 'print' is a function in Python3, not just a
   keyword, so the if-else logic can be removed.
2. runtime.py: runtimeStudentCode defines a helper function to check if
   a function has timed out.
3. studentCode.py functions no longer need arguments passed to them. The
   relevant function calls in runtime have been amended.


runtime/testy/runtime.py
```diff
@@ -103,17 +117,111 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+  testNameRegex = re.compile(".*_setup")
+  testNames = [testName[:-len("_setup")] for testName in dir(studentCode) if testNameRegex.match(testName)]
+
+  failCount = 0
+  failedTests = []
+
+  for testName in testNames:
+    testFileName = "%s_output" % (testName,)
+    with open(testFileName, "w") as testOutput:
+      # Wrapper to flush after every write. This ensures file is written to in the same
```
bodiddles: Make this a docstring

runtime/testy/runtime.py
```diff
@@ -103,17 +117,111 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
       stateQueue.put([SM_COMMANDS.ADD, [processName, pipeToChild]], block=True)
       pipeFromChild.recv()
-    newProcess = multiprocessing.Process(target=helper, name=processName.value, args=(badThingsQueue, stateQueue, pipeFromChild))
+    newProcess = multiprocessing.Process(target=helper, name=processName.value, args=[badThingsQueue, stateQueue, pipeFromChild] + list(args))
     allProcesses[processName] = newProcess
     newProcess.daemon = True
     newProcess.start()
   return spawnProcessHelper
 
+def testTime():
```
bodiddles: Name this something more descriptive please, makes it seem like you're testing the time module

runtime/testy/runtime.py
```diff
@@ -103,17 +117,111 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+    print("All {0} tests passed.".format(len(testNames)))
+  else:
+    print("{0} of the {1} tests failed.".format(failCount, len(testNames)))
+    print("Output saved in {{test_name}}_output.")
+    print("Inspect with 'diff {{test_name}}_output {0}{{test_name}}_output".format(RUNTIME_CONFIG.TEST_OUTPUT_DIR.value))
+    for testName in failedTests:
+      print("    {0}".format(testName))
+
+
+def testSuccess(testFileName):
```
bodiddles, line 194: Can we not use the built in diff function instead of making our own?

runtime/testy/runtime.py
```diff
@@ -103,17 +117,111 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+      stateQueue = multiprocessing.Queue()
+      spawnProcess = processFactory(badThingsQueue, stateQueue, sys.stdout)
+      restartCount = 0
+
+      try:
+        spawnProcess(PROCESS_NAMES.STATE_MANAGER, startStateManager)
+        while True:
+          if restartCount >= 3:
+            break
+          spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode, testName, 3)
```
bodiddles, line 161: Potential issue with socket integration as it only starts two processes. We need to be able to toggle between runtime and "testTime" before I think we should merge this in.

runtime/testy/runtimeUtil.py
```diff
@@ -9,6 +9,7 @@ class BAD_EVENTS(Enum):
   STUDENT_CODE_TIMEOUT    = "Student Code Timed Out"
   UNKNOWN_PROCESS         = "Unknown State Manager process name"
   STATE_MANAGER_KEY_ERROR = "Error accessing key in State Manager"
+  END_EVENT               = "Process terminated" # Used for testing
```
bodiddles: description doesn't really signal that it ended gracefully. I would change the message

runtime/testy/runtime.py
yizow: I dont feel like there is anything preventing us from merging this. 

The old runtime also only spawned two processes. This new function is a strict upgrade of the previous: it allows us to test more, more thoroughly, and more independently. 

Adding socket integration needs to be part of the socket integration commit. 

The command line arguments to toggle are in a commit Christene Zheng is working on. The functionality can still be readily simulated by commenting out one line of code. Furthermore, i dont see a way the old runtime code will be useful, other than running infinitely (which we also dont want if we want a test to finish.)

runtime/testy/runtime.py
yizow: Fixed

runtime/testy/runtime.py

runtime/testy/runtime.py

runtime/testy/runtimeUtil.py
yizow: To what? its not exactly that the process ended gracefully either; or at least, it's only true in testing. An actual runtime should never send this event.

runtime/testy/runtime.py
```diff
@@ -103,17 +118,100 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+  testNameRegex = re.compile(".*_setup")
+  testNames = [testName[:-len("_setup")] for testName in dir(studentCode) if testNameRegex.match(testName)]
+
+  failCount = 0
+  failedTests = []
+
+  for testName in testNames:
+    testFileName = "%s_output" % (testName,)
+    with open(testFileName, "w") as testOutput:
+      class Unbuffered:
```
nikitakit: If you're on Python 3, have you tried the following: https://docs.python.org/3/library/io.html#raw-i-o

Perhaps this solution to the buffering problem obsoletes the mailing list solution from 2003.

runtime/testy/runtime.py
```diff
@@ -64,33 +66,46 @@ def runtime():
...
+def runStudentCode(badThingsQueue, stateQueue, pipe, testName = "", maxIter = 0):
   try:
     import signal
-    def timed_out_handler(signum, frame):
+    def timedOutHandler(signum, frame):
       raise TimeoutError("studentCode timed out")
-    signal.signal(signal.SIGALRM, timed_out_handler)
+    signal.signal(signal.SIGALRM, timedOutHandler)
+
+    def checkTimedOut(func, *args):
```
nikitakit, line 75: I think this function is misnamed. It doesn't really check for timeout at all, it just schedules an alarm and then runs something.

runtime/testy/runtime.py
```diff
@@ -103,17 +118,100 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+      try:
+        spawnProcess(PROCESS_NAMES.STATE_MANAGER, startStateManager)
+        while True:
+          if restartCount >= 3:
+            break
+          spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode, testName, 3)
+          while True:
+            newBadThing = badThingsQueue.get(block=True)
+            print(newBadThing.event)
+            if (newBadThing.event == BAD_EVENTS.STUDENT_CODE_ERROR) or \
```
nikitakit: Past some point it might be cleaner to say `if variable in (VALUE1, VALUE2, VALUE3...)`

runtime/testy/runtimeUtil.py
```diff
@@ -34,10 +35,11 @@ def __new__(cls):
   SET_VAL             = ()
 
 class RUNTIME_CONFIG(Enum):
-  STUDENT_CODE_TIMEOUT = 3
-  STUDENT_CODE_HZ    = 5 # Number of times to execute studentCode.main per second
-  DEBUG_DELIMITER_STRING  = "****************** RUNTIME DEBUG ******************"
-  PIPE_READY = ["ready"]
+  STUDENT_CODE_TIMEOUT        = 3
```
nikitakit, line 41: Why is there a diff here when this line hasn't changed?

nikitakit (commented):
+1 for having a testing framework!

I have to say that I'm suspicious about your choice to evaluate the tests by comparing stdout. Specifically, the output appears to be coming from multiple processes, which could easily cause nondeterministic ordering of the print statements. Do you have good reason to believe that you won't have a high spurious failure rate due to variation in scheduling?

Possible alternatives for getting deterministic tests are:
- Use assert statements in the student code only. If you want to test some part of runtime, have the student code invoke some special debug interface that forces the runtime to trigger those codepaths.
- Use multiple streams per thread (or annotations about which thread prints what), and only demand correct ordering within one stream.

(Not assigning a approve/change here because it doesn't appear that I'm the primary reviewer)

runtime/testy/runtime.py
yizow: Right. It checks if the function we run is going to time out, right?

runtime/testy/runtime.py
yizow: Fixed. 
Apparently there's a "buffering" keyword arg we can pass in to open()
I set it to line buffering, which is what we want.

runtime/testy/runtime.py

runtime/testy/runtimeUtil.py
yizow: I made the whitespace line up. I was modifiying the code here so figured i might as well make it look prettier. 
Should i remove this into a separate mini-refactoring commit?

yizow:
This framework was created with the purpose of testing student code in isolation. 
In that context, I think the current method is fine. Even though multiprocessing has the possibility of out of order print statements, the line buffering solves that problem. The main runtime process only wakes when a BadThing has happened. Actually, the process as a whole just acts like a really fancy semaphore... 
For now, I believe this is sufficient to guarantee deterministic print statement orderings.

As for assert statements: The studentcode functions are designed to be called multiple times. Sometimes we need to test persistence/ change _across_ calls. Assert statements won't allow us to do this.

When we add new processes, what we can do is add multiple files to the test. So we would have test1_output_studentCode, test1_output_Hibike, etc. 

That's IF we continue to use this framework to test the other parts of runtime.

I think it should be fine/ doable.

nikitakit:
So I do want to point out that print statements from the state manager are being captured in the log (as far as I can tell). That's where the non-determinism will come in, and not with the main runtime thread.

I challenge you to add a `print(request)` after line 97 of `stateManager.py`. I believe that should be sufficient to cause 100% test failure rate. This is not exactly non-determinism, but it does create a potential surprising behavior for staff working on the project.

On the other hand, state manager doesn't really do any printing besides the hello command. If you think that will remain the case in the future, this isn't that big of a problem.

Other than that this PR looks good.

yizow:
Statemanager shouldnt be printing in the future. The current hello is a holdover from the early days when i needed to see what was going on. 

Re: adding the print statement
`print(request)` will always fail because one of the things that is printed is the memory address of the connection object, which changes every time. 
`print(request[0])`, or anything that doesnt include that memory address, is deterministic as far as my testing can show. 
This is because every other process waits on statemanager to reply before proceeding.

nikitakit:
I see. LGTM

yizow:
Updated to fix merge conflicts with recent master commit. 
runtimeUtil.py 
studentCode.py

nikitakit (approved):

PR #52, opened 2016-10-10
[RUNTIME] Adds createKey methods to both studentAPI and stateManager.

cjsanchez:
-   Accepts a list of keys for creating nested dictionaries and creating
  keys for nested dictionaries.
-   Raises an error when attempting to create a key for a value that is
  not a dictionary.
-   Modifies setValue to no longer be able to create new keys.
-   Creates testing function createKey_main to test functionality.


runtime/testy/studentCode.py
```diff
@@ -25,11 +25,38 @@ def nestedDict_setup():
...
+    except StudentAPIKeyError:
+      print("Success!")
+    else:
+      print("ERROR: setValue can create keys :(")
+
+  print("Creating key 'Klefki' and setting to value 707")
+  Robot.createKey("Klefki")
+  Robot.setValue(707, "Klefki")
+  print("Success!")
+
```
yizow, line 64: You never actually check using `getValue` that you successfully set it. 
Might be good to include that just in case.

runtime/testy/stateManager.py
```diff
@@ -42,6 +44,22 @@ def addPipe(self, processName, pipe):
...
+        if key not in currDict:
+          currDict[key] = {}
+        currDict = currDict[key]
+      except TypeError:
+        error = StudentAPIKeyError(
+          "key '{}' is defined, but does not contain a dictionary.".format(key))
+        self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
+        return
+    # Returns new state as a string
+    message = json.dumps(self.state, indent=4)
```
yizow: We dont want to return the state for two reasons. 
1. We dont want students having access to the entire state.
2. That's a lot of data to send, and we wont use nearly all of it.

If this error occurs, it seems like something internal in runtime broke. I think we should consider it a BadThing.

Would also mean we won't need to import the json library.

yizow (changes requested):
OH MY GOD YOU USED THE TESTING FRAMEWORK YOURE AMAZING.

Other than that, just minor fixes. On the whole looks good to me.

yizow (approved):
LGTM

PR #53, opened 2016-10-12
[RUNTIME] Updated README.MD

yizow:
Updated README.MD to include information on:
- Writing and formatting commit messages.
- Viewing only runtime commits with git log
- Squashing commits
- Reordering commits
- Running runtime


akumaraguru (approved):
SGTM. Very clever with the --grep trick

PR #55, opened 2016-10-12
[RUNTIME] runtime counts studentCode.main executions

yizow:
Runtime now counts the number of times that studentCode.main runs. The
counter is incremented at the end of each successful execution of the
student function.

A new dictionary has been added to stateManager with the key
"runtime_meta". This dictionary should be used to store all runtime
relevant meta information.

SM_COMMANDS.HELLO has been removed, and replaced with
SM_COMMANDS.STUDENT_MAIN_OK. It takes no arguments, and simply
increments the counter.

The speed of student code execution has been increased to 20 Hz. It was
previously slowed so that we could observe the execution of runtime, but
now we have tests.


brandonxxlee (approved):
LGTM

PR #61, opened 2016-10-22
[RUNTIME] Adds UDP communication capabilities with Dawn

bodiddles:
Refactors UDP processes into two processes (send and receive). Each process
then spawns 2 threads, one for packaging/unpackaging, and one for sending/receiving
respectively.

The send and receive processes are now encapsulated in classes that 
the shared buffers between the package and send and the receive and unpackage
threads respectively. Each process class inherits from a super class that handles
thread creation and starting said threads.

Currently the code just sends a test string and adds the bytes received from
the receive port into the state dictionary with the key of "bytes". Protobuf
integration and still needs to be accomplished as well as handling gamepad
data from Dawn.

Adds 2 command types and corresponding functions for state manager to send
the state to the send process as well as adding the data received from Dawn
into the dictionary.

Creates new process name enums and adds thread enums for state manager
to keep track of all programs running in Ansible.

The current Ansible has been tested with fake_dawn.py.


runtime/testy/runtimeUtil.py
```diff
@@ -19,6 +21,15 @@ class PROCESS_NAMES(Enum):
...
+  UDP_SEND_PROCESS    = "udpSendProcess"
+  UDP_RECEIVE_PROCESS = "udpReceiveProcess"
+
+@unique
+class THREAD_NAMES(Enum):
+  UDP_PACKAGER        = "udpPackager"
+  UDP_SENDER          = "udpSender"
+  UDP_RECEIVER        = "udpReceiver"
+  UDP_UNPACKAGER      = "udpUnpackager"
   HIBIKE              = "hibike"
```
yizow, line 26: Hibike isn't a thread.
my guess is somewhere in the merging you mixed this up.

runtime/testy/runtimeUtil.py
```diff
@@ -19,6 +21,15 @@ class PROCESS_NAMES(Enum):
   STUDENT_CODE        = "studentProcess"
   STATE_MANAGER       = "stateProcess"
   RUNTIME             = "runtime"
+  UDP_SEND_PROCESS    = "udpSendProcess"
+  UDP_RECEIVE_PROCESS = "udpReceiveProcess"
+
+@unique
+class THREAD_NAMES(Enum):
```
yizow: Is there a reason to put this in runtimeUtil.py and not Ansible.py? 
I've been using runtimeUtil for things that are used across files. It looks like the thread names are only used within Ansible.py though. 

If so, i think putting it in there will make things cleaner and more maintaneable. Better abstraction, etc.

runtime/testy/runtime.py
```diff
@@ -208,4 +219,4 @@ def startHibike(badThingsQueue, stateQueue, pipe):
     badThingsQueue.put(BadThing(sys.exc_info(), str(e)))
 
 if __name__ == "__main__":
-  runtimeTest()
+  runtime()
```
yizow, line 223: I think the suite of tests that runtimeTest runs should be default behaviour.

runtime/testy/runtime.py
```diff
@@ -118,6 +115,20 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
     SM.start()
   except Exception as e:
     badThingsQueue.put(BadThing(sys.exc_info(), str(e)))
+    
+def startUDPSender(badThingsQueue, stateQueue, smPipe):
+  try:
+    sendClass = Ansible.UDPSendClass(badThingsQueue, stateQueue, smPipe)
+    sendClass.start()
+  except Exception:
+    badThingsQueue.put(BadThing(sys.exc_info(), None))
```
yizow: Please put additional information in this bad thing. 

`BadThing(sys.exc_info(), "String description", event=BAD_EVENTS.EVENT_TYPE)

runtime/testy/runtime.py
```diff
@@ -118,6 +115,20 @@ def startStateManager(badThingsQueue, stateQueue, runtimePipe):
...
+
+def startUDPReceiver(badThingsQueue, stateQueue, smPipe):
+  try:
+    recvClass = Ansible.UDPRecvClass(badThingsQueue, stateQueue, smPipe)
+    recvClass.start()
+  except Exception:
+    badThingsQueue.put(BadThing(sys.exc_info(), None))
```
yizow: See above

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
+import socket
+import threading
+import queue
+import time
+data = [0]
```
yizow, line 6: add newline between imports and global variables.

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
...
+	host = socket.gethostname()
+	with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+		while True:
+			msg = None
+			string = "From Dawn" 
+			b = bytearray()
+			b.extend(map(ord, string))
+			msg = bytes(b)
+			s.sendto(msg, (host, send_port))
+def receiver(port, receive_queue):
```
yizow, line 26: add newline between functions

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
...
+		msg, addr = s.recvfrom(2048)
+		receive_queue[0]=msg
+
+sender_thread = threading.Thread(target = sender, name = "fake dawn sender", args = (send_port, data))
+recv_thread = threading.Thread(target = receiver, name = "fake dawn receiver", args = (recv_port, data))
+sender_thread.daemon = True
+recv_thread.daemon = True
+recv_thread.start()
+sender_thread.start()
+while True:
```
yizow, line 42: Newline before thread initialization and 'main' function

yizow, line 42: This is busy waiting. This is also generally really bad for performance. 

I suppose if you are using this to simulate Dawn, it should be fine. Since it will be replaced, right?

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
...
+	print(data)
+	time.sleep(1)
```
yizow: Add newline at end of file.

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
+import socket
+import threading
+import queue
+import time
+data = [0]
+send_port = 1236
+recv_port = 1235
+
+def sender(port, send_queue):
+	host = socket.gethostname()
```
yizow: Can probably move `host` assignment to outside both of these functions. The host should never change while we are running the threads.

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,35 @@
+import socket
+import threading
+import queue
+import time
+data = [0]
+send_port = 1236
+recv_port = 1235
+
```
yizow, line 10: Why create two different sockets? Shouldn't we be using the same socket to both send and receive information, since it's the same connection? 

After some internet digging, it seems using SOCK_DGRAMS means that the send and recv on the socket are atomic. I think it should be fine.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+                                  args = (self, self.badThingsQueue, self.stateQueue, self.pipe))
+        thread.daemon = True
+        return thread
+
+    def start(self):
+        packagerThread = self.threadMaker(self.packagerFunc, self.packagerName)
+        socketThread = self.threadMaker(self.socketFunc, self.socketName)
+        packagerThread.start()
+        socketThread.start()
+        while True:
```
yizow: This is busy waiting. 
This is bad.

Use `packagerThread.join(); socketThread.join()`

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+                         UDPSendClass.udpSender, badThingsQueue, stateQueue, pipe)
+
+    def packageData(self, badThingsQueue, stateQueue, pipe):
+        """Function run as a thread that packages data to be sent.
+
+        The robot's current state is received from the StateManager via the pipe and packaged 
+        by the package function, defined internally. The packaged data is then placed 
+        back into the TwoBuffer replacing the previous state.
+        """
+        def package(state, badThingsQueue):
```
yizow: You can remove the badThingsQueue object. It is already bound in the scope of the parent object. (Assuming there is only one badThingsQueue object being shared within the scope of packageData)

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+        return self.data[self.get_index]
+
+
+class AnsibleHandler():
+    """Parent class for UDP Processes that spawns threads 
+
+    Initializes generalized instance variables for both UDP sender and receiver, and creates a callable
+    method to initialize the two threads per UDP process and start them. 
+    """
+    def __init__(self, packagerName, socketName, packager, socketThread, badThingsQueue, stateQueue, pipe):
```
yizow: might make more sense to organize the argument list to be `packagerName, packagerThread, socketName, socketThread` to improve clarity.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+        back into the TwoBuffer replacing the previous state.
+        """
+        def package(state, badThingsQueue):
+            """Helper function that packages the current state.
+
+            Currently this function converts the input into bytes. This will
+            eventually be implemented to package the rawState into protos.
+            """
+            s = "TEST" 
+            b = bytearray()
```
yizow: The extend function can be combined with initialization. Both take an iterable of integers in the range [0, 256) as arguments.
`b = bytearray(map(ord, s))`

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+            b.extend(map(ord, s))
+            return b 
+
+        while True:
+            try:
+                stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                rawState = pipe.recv()
+                if rawState == RUNTIME_CONFIG.PIPE_READY:
```
yizow: These two lines will never be received by you. Deprecated behaviour, please remove.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+        def unpackage(data):
+            """Function that takes a packaged proto and unpackages item
+
+            Currently simply returns the original data. Needs to be implemented
+            """
+            return data 
+        while True:
+            try:
+                unpackagedData = unpackage(self.recvBuffer.get())
+                stateQueue.put([SM_COMMANDS.RECV_ANSIBLE, [unpackagedData]])
```
nikitakit: This loop will keep unpacking the same data repeatedly, wasting CPU time. Don't do this.

nikitakit (changes requested):
Sender class looks fine, but the receiver feels wrong.

You shouldn't need multiple thread for the receiver. Just do socket read, unpackage, followed by adding to statemanager all in the same thread. Also don't use a TwoBuffer for receive.

If you're worried about not being able to keep up with received data, you can do that by asking the kernel how much data is in the socket, and dropping some packets if the kernel reports a backlog. But this is an optimization that can be done in a later PR (it might be unneeded anyway).

The use of multiple threads (a) adds latency at each cross-thread hop, (b) currently uses a busy-loop that will steal all of the BBB's CPU time, and (c) is actually more complex than the single-threaded approach.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+            b.extend(map(ord, s))
+            return b 
+
+        while True:
+            try:
+                stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                rawState = pipe.recv()
+                if rawState == RUNTIME_CONFIG.PIPE_READY:
+                    stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                elif rawState:
```
yizow: Assuming rawState is a dictionary:
An empty dictionary returns a false value. Are we sure than an empty dictionary will _never_ have semantic meaning? 

Does this if check guard against anything? Even if we receive an empty dictionary, could there be a case where we actually do want to send an empty dictionary? 
Consider just removing the if statement altogether. we're calling `pipe.recv()`, meaning that stateManager has deliberately put something on the pipe for us to see.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+            b = bytearray()
+            b.extend(map(ord, s))
+            return b 
+
+        while True:
+            try:
+                stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                rawState = pipe.recv()
+                if rawState == RUNTIME_CONFIG.PIPE_READY:
+                    stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
```
yizow: Do we perhaps want to limit how fast this runs? Currently, we are asking for new info to send as soon as we send the previous one. Could we flood the queue so that nothing else gets processed? 

If you want to rate limit this, look in runtime.py for how i'm rate limiting the frequency of studentCode.main() calls.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def udpSender(self, badThingsQueue, stateQueue, pipe):
+        """Function run as a thread that sends a packaged state from the TwoBuffer
+        
+        The current state that has already been packaged is gotten from the 
+        TwoBuffer, and is sent to Dawn via a UDP socket.
+        """
+        host = socket.gethostname() #TODO: determine host in runtime-dawn comm
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+            while True:
```
yizow, line 122: This loop needs a time.sleep in it somewhere. Currently, this is just constantly sending without any rate limiting. 

time.sleep at some set interval configurable in runtime.util would be my recommendation.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,176 @@
...
+        while True:
+            try:
+                unpackagedData = unpackage(self.recvBuffer.get())
+                stateQueue.put([SM_COMMANDS.RECV_ANSIBLE, [unpackagedData]])
+            except Exception as e:
+                    badThingsQueue.put(BadThing(sys.exc_info(), 
+                    "UDP sender thread has crashed with error:",  
+                    event = BAD_EVENTS.UDP_RECV_ERROR, 
+                    printStackTrace = True))
+
```
yizow: missing newline

yizow (changes requested):
Sorry for the delay. Here's a host of comments.

I agree with nikita's comment's on receiver, so didn't code review those parts.

cbartondock:
MJ MN CV

runtime/testy/Ansible.py
yizow: Do we need to have the badThingsQueue as an argument?

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,188 @@
...
+            """
+            s = "TEST" 
+            b = bytearray(map(ord, s))
+            return b 
+
+        while True:
+            try:
+                stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                rawState = pipe.recv()
+                packState = package(rawState, badThingsQueue)
```
yizow: Dont need to pass in badThingsQueue. It's already in the scope of the package function.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,188 @@
...
+                self.sendBuffer.replace(packState)
+                time.sleep(1.0/self.packagerHZ)
```
yizow: Not accurate. Time varies depending on how long we block and how long the package function takes to execute. And on when the scheduler decides to run this process.
runtime.py has an example to tie the sleep duration to the amount of time the it takes to run.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,188 @@
...
+        TwoBuffer, and is sent to Dawn via a UDP socket.
+        """
+        host = '127.0.0.1' #TODO: determine host in runtime-dawn comm
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+            while True: 
+                try:
+                    msg = self.sendBuffer.get()
+                    if msg != 0: 
+                        s.sendto(msg, (host, UDPSendClass.SEND_PORT))
+                    time.sleep(1.0/self.socketHZ)
```
yizow: See comments above. Not necessarily accurate timing.

runtime/testy/runtime.py
yizow: I still think runtimeTest should be run here.

yizow (changes requested):

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, packagerName, packagerThread, socketName, socketThread,
+                 badThingsQueue, stateQueue, pipe, packagerHZ, socketHZ):
+        self.packagerFunc = packagerThread
+        self.socketFunc = socketThread
+        self.badThingsQueue = badThingsQueue
+        self.stateQueue = stateQueue
+        self.pipe = pipe
+        self.packagerName = packagerName
+        self.socketName = socketName
+        self.packagerHZ = packagerHZ
```
yizow: Should these two be the same, or different? Will we ever want a case where one of these threads is running faster than the other?

runtime/testy/fake_dawn.py
```diff
@@ -0,0 +1,39 @@
...
+import time
+
+data = [0]
+send_port = 1236
+recv_port = 1235
+
+def sender(port, send_queue):
+	host = '127.0.0.1'
+	with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+		while True:
```
yizow, line 14: Isn't this going to flood the receiving process?

yizow (changes requested):

runtime/testy/fake_dawn.py
bodiddles: Done

yizow (approved):
Things look good to me.

yizow:
Oh, make sure to squash and write a new commit message.
May want to mention brandon's name in the commit message as well, so people know there are two people who can answer questions on this code.

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,199 @@
...
+class TwoBuffer():
+    """Custom buffer class for handling states.
+
+    Holds two states, one which is updated and one that is sent. A list is used because
+    the get and replace methods are atomic operations. Replacing swaps indices of the 
+    get and put index.
+
+    """
+    def __init__(self):
+        self.data = [0, 0]
```
nikitakit, line 25: I believe Python convention is to use None instead of 0

runtime/testy/Ansible.py
```diff
@@ -0,0 +1,199 @@
...
+            return b 
+
+        while True:
+            try:
+                nextCall = time.time()
+                stateQueue.put([SM_COMMANDS.SEND_ANSIBLE, []])
+                rawState = pipe.recv()
+                packState = package(rawState)
+                self.sendBuffer.replace(packState)
+                nextCall += 1.0/self.packagerHZ
```
nikitakit, line 105: This type of logic is not resilient against a one-time delay. If, say, package takes forever a single time, it is possible that this loop will never sleep again in the immediate aftermath.

nikitakit (approved):
Major concerns have been addressed, and this now looks ready to merge!

I haven't actually run the code, so I trust that you guys actually tested that it works.

At this point integration takes priority over perfection. However, there are a few minor nitpicks I pointed out that you might want to consider at some point in the future.

runtime/testy/Ansible.py
bodiddles: Doess the initialization of nextCall each loop not catch this as it only times how long each iteration of the loop takes? If it does take too long, it would never sleep but that would be the correct behavior as we only want to limit the maximum rate it goes right?

nikitakit: You're right, I misread the code.

PR #62, opened 2016-10-29
[RUNTIME] Add timestamps to StateManager.state

jiangdaniel:
- Modify each element el in StateManager.state to be [el, timestamp] in order
  to keep track of when elements are updated
- Modify setValue and createKey in StateManager to update the timestamp of the
  element being modified and all of the timestamps of the element's parent
  elements
- Modify getValue in StateManager to work with the new timestamp format
- Create function StateManager.getTimestamp to retrieve timestamps from
  StateManager.state. Parameters are exactly the same as in
  StateManager.getValue
- Add test for timestamps


runtime/testy/stateManager.py
```diff
@@ -64,24 +66,29 @@ def addPipe(self, processName, pipe):
...
+          currDict[key] = [{}, 0]
+        path.append(currDict[key])
+        currDict = currDict[key][0]
       except TypeError:
         error = StudentAPIKeyError(
           "key '{}' is defined, but does not contain a dictionary.".format(key))
         self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
         return
+    currTime = time.time()
+    for item in path:
```
yizow, line 84: This means we iterate through the path twice. We can accomplish the same results in just a single pass, by updating the timestamp as we iterate through the first time. This will allow you to remove the `path` instance variable you created, as well. 

So, call `currTime = time.time()` before iterating through keys. Before the `currDict = currDict[key][0]` line, add `currDict[key][1] = currTime`. 
Or something like that; I may have an off by one error.

runtime/testy/stateManager.py
```diff
@@ -91,21 +98,40 @@ def setValue(self, value, keys):
...
+        currDict = currDict[key[1]][0]
       if len(keys) > 1:
         i += 1
       if keys[i] not in currDict:
         raise Exception
-      currDict[keys[i]] = value
+      path.append(currDict[keys[i]])
+      currDict[keys[i]][0] = value
+      currTime = time.time();
+      for item in path:
```
yizow, line 114: Same comment as above. We can accomplish the same result in a single pass by updating timestamps as we search into the dictionary.

runtime/testy/stateManager.py
```diff
@@ -91,21 +98,40 @@ def setValue(self, value, keys):
...
       self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(value)
     except:
       error = StudentAPIKeyError(self.dictErrorMessage(i, keys, currDict))
       self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
 
+  def getTimestamp(self, keys):
+    currDict = self.state
+    timestamp = 0
+    try:
+      for key in enumerate(keys):
```
yizow: we can get rid of the first line following for loop with
`for i, key in enumerate(keys):`

You should apply this to the other places you have `i=key[0]` as well.

runtime/testy/stateManager.py
```diff
@@ -91,21 +98,40 @@ def setValue(self, value, keys):
...
+        i = key[0]
+        timestamp = currDict[key[1]][1]
```
yizow: slightly cleaner way of doing this:
`timestamp, currDict = currDict[key[1]]`

runtime/testy/stateManager.py
```diff
@@ -91,21 +98,40 @@ def setValue(self, value, keys):
...
+        currDict = currDict[key[1]][0]
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(timestamp)
+    except:
+      error = StudentAPIKeyError(self.dictErrorMessage(i, keys, result))
+      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
+
   def studentCodeTick(self):
-    self.state["runtime_meta"]["studentCode_main_count"] += 1
+    self.state["runtime_meta"][0]["studentCode_main_count"][0] += 1
```
yizow, line 139: Ugh, this is getting ugly. ( Not your fault)
Let's start thinking of ways we might be able to get rid of this though in a future commit

yizow (changes requested):
A few readability changes.
Otherwise, looks good!

runtime/testy/stateManager.py
jiangdaniel: @yizow The reason I chose to update timestamps by iterating over a second time was that in the case of an error being thrown while attempting to create a key or set a value, I figured that the timestamps should not be updated.  So I iterate through the first time to create a key or set a value; and then after that loop is done, I know that it was successful, so I then update the timestamps.  If I were setting the timestamps while doing whatever the main operation is, wouldn't I have to write some weird timestamp reversal function in case of an error occurring?

runtime/testy/stateManager.py
jiangdaniel: The reason I initialize the variable `i` outside of the for loops is that in the case of an error, I use `i` in the `except` block to help me create the error message.

`error = StudentAPIKeyError(self.dictErrorMessage(i, keys, result))`

runtime/testy/stateManager.py
yizow: Oh good catch, I hadn't even thought of that yet.

runtime/testy/stateManager.py
yizow: @jiangdaniel this is a simple readability fix to make

yizow (changes requested):

yizow (approved):
LGTM

PR #63, opened 2016-10-29
Updated actions to match reducers

kev-chien:
- Made new actions to match those implemented in reducers, specifically Peripherals, Info, and Gamepads.
- Also reordered actions to match the order in reducers.


kevinmasd (approved):
Looks good.

PR #64, opened 2016-11-01
[HIBIKE] create hibike simulator for runtime integration

karthik-shanmugam:
- rewrite `hibike_simulator.py` and add a function `hibike_process`, which takes in various Queues and Pipes and simulates the behavior described in `ATALANDA_INTEGRATION.md`

- rewrite `README.md`, `hibikeDevices.json`, and `hibike_message.py` with protocol changes

- add an `__init__.py` so hibike can be imported from runtime


hibike/ATALANTA_INTEGRATION.md
```diff
@@ -0,0 +1,83 @@
+# Hibike Runtime Integration
```
yizow: Rename the file to just RUNTIME_INTEGRATION. 

Atalanta is a deprecated name

hibike/ATALANTA_INTEGRATION.md
```diff
@@ -0,0 +1,83 @@
...
+- switch2: bool read only
+
+Servo
+
+- servo0_torqued: bool read/write
+- servo0_pos: int read/write
+
+MetalDetector
+
+- callibrate: bool write only
```
yizow: typo

nikitakit:
I see that I have been assigned. Reviewing this now, but given the size of the change it may take a while.

hibike/README.md
```diff
@@ -46,75 +104,118 @@ Each Smart Device will be assigned an 88-bit UID with the following data.
...
 
-Device Type - 16-bit ID specifying the type device the Smart Device controller is attached to.
-              Device types are enumerated in Section 4
+- Device Type
+    - 16-bit ID specifying the Type of a Smart Device
+    - Device types are enumerated in Section 5
+
+- Year
+  - 8-bit ID corresponding to the competition year that the Smart Device was manufactured for. 
+  - The 2015-2016 season will correspond to 0x00
```
nikitakit: The spec should be updated with a new year, `0x01`.

Additionally, is possible to detect last year's hibike devices and warn about protocol incompatibility to students? I would not be at all surprised if students try to plug in last year's sensors to their robots. Breaking compatibility is fine as long this does not become a yearly thing, but having old sensors crash the robot would be unfortunate.

hibike/README.md
```diff
@@ -46,75 +104,118 @@ Each Smart Device will be assigned an 88-bit UID with the following data.
...
-    +------------------------------------+
-    |  0x08   |    Description Request   |
-    +------------------------------------+
-    |  0x09   |   Description Response   |
-    +------------------------------------+
-    |  0xFF   |           Error          |
-    +------------------------------------+
+|   ID    |       Message Type       |
+|---------|--------------------------|
+|  0x10   |          Ping            |
```
nikitakit, line 151: Have you thought carefully about whether future kits will be able to extend hibike in a backward-compatible way? Breaking changes this year is fine, but ideally we could use a few years of experience building hibike to then lock in a stable system.

hibike/README.md
```diff
@@ -46,75 +104,118 @@ Each Smart Device will be assigned an 88-bit UID with the following data.
...
+|         |                | 5            | total      | float      | yes   | no     |
+|         |                | 6            | callibrate | bool       | no    | yes    |
+|  0x05   | TeamFlag       | 0            | mode       | bool       | yes   | yes    |
+|         |                | 1            | blue       | bool       | yes   | yes    |
+|         |                | 2            | yellow     | bool       | yes   | yes    |
+|         |                | 3            | s1         | bool       | yes   | yes    |
+|         |                | 4            | s2         | bool       | yes   | yes    |
+|         |                | 5            | s3         | bool       | yes   | yes    |
+|         |                | 6            | s3         | bool       | yes   | yes    |
+|  0x06   | Grizzly        |              |            |            |       |        |
```
nikitakit: What about Yogi Bear? Does it get an ID?

hibike/README.md
```diff
@@ -124,187 +225,152 @@ Note: As of now, Grizzlies are not supported by Hibike (pyGrizzly should
...
-    +---------+---------------+
-    | Status  |    Meaning    |
-    +-------------------------+
-    |   0xFB  |    Invalid    |
-    |         | Message Type  |
-    +-------------------------+
-    |   0xFC  |   Malformed   |
-    |         |    Message    |
-    +-------------------------+
-    |   0xFD  |  Invalid UID  |
```
nikitakit, line 136: I'm seeing some of these error types disappear. Is that intentional?

hibike/README.md
```diff
@@ -46,75 +104,118 @@ Each Smart Device will be assigned an 88-bit UID with the following data.
...
-ID          - Randomly generated 64-bit number that will uniquely identify each Smart Device within
-              a specific device type and year. With 64-bit IDs, the probability of a hash collision 
-              with 1000 of 1 type of device per year is roughly 0.05%
 
-## Section 3: Enumerations
+
+
+    ```
+    +-------------------+-----------------------------+     +-----------------------------+
+    |      Params       |         Value 0             | ... |         Value 15            |
```
nikitakit, line 135: Let's say you only request 2 params via the bitmask. Does the value section only contain data for two values, or will there be 16 value sections (with most of them set to zero)? I think this aspect of the spec can be clarified.

hibike/__init__.py
```diff
@@ -0,0 +1,3 @@
+__all__ = ['hibike_simulator', 'hibike_message']
+from . import hibike_simulator
+from . import hibike_message
```
nikitakit: Note the little red symbol at the last line of this diff. It means you don't have a newline at the end of file. This is typically discouraged style, because it pops up in the diff like this.

hibike/hibikeDevices.json
```diff
@@ -1,288 +1,83 @@
...
-                "test3",
-                "test4",
-                "test5"
-            ]
-        }
-    ]
-}
\ No newline at end of file
+[
+    {"id": 0,     "name": "LimitSwitch",   "params": [
```
nikitakit, line 2: BTW, the formatting here is super awkward in diff view. Is there any change you could adjust the whitespace in this file?

hibike/hibike_message.py
```diff
@@ -1,55 +1,74 @@
...
-messagePayloadLengths = {
-  messageTypes["SubscriptionRequest"] :  2,
-  messageTypes["SubscriptionResponse"] : 13,
-  # TODO: find a clean way to not hardcode this
-  messageTypes["DataUpdate"] :           1,
-  messageTypes["DeviceUpdate"] :         5,
-  messageTypes["DeviceStatus"] :         5,
-  messageTypes["DeviceResponse"] :       5,
-  messageTypes["Error"] :                1
+  "LimitSwitch"   :       0x00,
```
nikitakit: Is there a reason these device IDs are inlined into this file, when the JSON could be sourced instead?

hibike/hibike_message.py
```diff
@@ -106,57 +125,203 @@ def checksum(data):
...
   payload = bytearray(temp_payload)
-  message = HibikeMessage(messageTypes["DeviceUpdate"], payload)
+  message = HibikeMessage(messageTypes["DeviceRead"], payload)
   return message
 
-def make_device_status(param):
-  """ Makes and returns DeviceStatus message."""
-  temp_payload = struct.pack('<B', param)
+def decode_params(device_id, params):
+#     Decodes an inputted set of parameters that is in binary form
```
nikitakit: Docstring instead of comment for the docs

hibike/ATALANTA_INTEGRATION.md
```diff
@@ -0,0 +1,83 @@
...
+- reading: int read only
+
+
+
+## StateManager -> Hibike
+
+`["ready"]`
```
nikitakit: I would recommend renaming this to "init_hibike" or similar.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,144 @@
...
+__all__ = ["hibike_process"]
+
+fake_uids = [0 << 72, 7 << 72]
+
+
+uid_to_index = {}
+
+
+def hibike_process(badThingsQueue, stateQueue, pipeFromChild):
+    while pipeFromChild.recv()[0] != "ready":
```
nikitakit: (repeating my recommendation to rename "ready")

hibike/hibike_message.py
```diff
@@ -106,57 +125,203 @@ def checksum(data):
     chk ^= data[i]
   return chk
 
+
```
nikitakit, line 111: Note that I didn't actually review the checksum for correctness.

nikitakit (commented):
I've left my comments inline.

I think it would be useful to have someone who worked a lot with last year's hibike protocol spec review this year's update. I never really delved into the nitty-gritty details, so I may have missed something.

hibike/README.md
karthik-shanmugam: It will only contain data for the two requested via bitmask

hibike/hibike_simulator.py
yizow: The ready command is actually deprecated and this line should be removed.

yizow (changes requested):
Haven't fully reviewed. 
Just a couple things I noticed on a first cursory pass.
Will do another review.

hibike/README.md
karthik-shanmugam: Yes, they were never actually implemented last year.

hibike/hibike_message.py
karthik-shanmugam: There was no reason; they've been removed now

hibike/__init__.py
karthik-shanmugam: Resolved

hibike/README.md
karthik-shanmugam: Future kits can add new message types; this year's devices will simply ignore them

hibike/hibikeDevices.json
yizow: This is a json file right, whitespace doesn't have semantic meaning? Just do indents normally. Currenyl, the "number": 0 stuff is waaaaay indented in.

hibike/hibikeDevices.json
```diff
@@ -1,288 +1,83 @@
...
-                "test4",
-                "test5"
-            ]
-        }
-    ]
-}
\ No newline at end of file
+[
+    {"id": 0,     "name": "LimitSwitch",   "params": [    
+                                                          {"number": 0  , "name": "switch0"    , "type": "bool"    , "read": true , "write": false },
```
yizow, line 3: If we're interested in efficiency, read/ write can be represented as a single integer 0, 1, 2, 3, representing none, either, or both. None doesn't really make sense though.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
-            return 1
-
-    def writeValue(self, UID, param, value):
-        if param >= len(self.subscribedTo[UID]):
-            return 1
-        self.subscribedTo[UID][param] = value
-        return 0
+import threading
+import multiprocessing
+from . import hibike_message as hm
```
yizow, line 4: Im still confused why we need the .

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+__all__ = ["hibike_process"]
+
+fake_uids = [0 << 72, 7 << 72]
+
+
+uid_to_index = {}
+
+
+def hibike_process(badThingsQueue, stateQueue, pipeFromChild):
+    ports = ['0', '1']
```
yizow, line 16: why not just delete ports, and use serials = range(2) for everythinG?

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+    serials = [int(port) for port in ports]
+
+    # each device has it's own write thread, with it's own instruction queue
+    instruction_queues = [queue.Queue() for _ in ports]
```
yizow, line 20: How many things are going to be writing to this queue? I thought hibike had a centralized model, meaning there is only one thread putting things on here. A pipe would be faster.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+import multiprocessing
+from . import hibike_message as hm
+import queue
+
+__all__ = ["hibike_process"]
+
+fake_uids = [0 << 72, 7 << 72]
+
+
+uid_to_index = {}
```
yizow, line 12: You can get rid of wordiness by just making instruction_queues a dictionary

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+        instruction, args = pipeFromChild.recv()
+        if instruction == "enumerate_all":
+            for instruction_queue in instruction_queues:
+                instruction_queue.put(("ping", []))
+        elif instruction == "subscribe_device":
+            uid = args[0]
+            if uid in uid_to_index:
+                instruction_queues[uid_to_index[uid]].put(("subscribe", args))
+
+class DeviceWriteThread(threading.Thread):
```
yizow, line 47: There is no need to extend the threading class.[http://stackoverflow.com/questions/660961/overriding-python-threading-thread-run]

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+    serials = [int(port) for port in ports]
+
+    # each device has it's own write thread, with it's own instruction queue
+    instruction_queues = [queue.Queue() for _ in ports]
+
+    # since this is a simulation and there are no real devices, this is how the write threads interface with the read threads
+    fake_device_queues = [queue.Queue() for _ in ports]
+
+    # each device has one write thread that receives instructions from the main thread and writes to devices
+    write_threads = [DeviceWriteThread(ser, iq, fake_device_queue) for ser, iq, fake_device_queue in zip(serials, instruction_queues, fake_device_queues)]
```
yizow, line 26: iq is a really hard to parse name. (so is ser)

Extra characters in variable names costs nothing but improves readability.

hibike/hibikeDevices.json
nikitakit: (I would still prefer this to be changed)

nikitakit:
No more big issues jump out to me on a second read-through. However, I think the possibility of plugging in last year's devices into the robot should not be ignored. Is there any way we can at least ensure that the protocol mismatch won't crash hibike? (Even better would be to report something like an "incompatible device" type to dawn, to make it very clear what the problem is)

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+                self.fake_subscription_thread.quit = True
+                self.fake_subscription_thread.join()
+                self.fake_subscription_thread = FakeSubscriptionThread(uid, delay, params, self.fake_device_queue)
+                self.fake_subscription_thread.start()
+            elif instruction == "device_values":
+                res = [instruction, args]
+
+            self.stateQueue.put(res)
+
+class FakeSubscriptionThread(threading.Thread):
```
yizow, line 96: No need to extend threading class.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+        self.fake_device_queue = fake_device_queue
+        self.quit = False
+        super().__init__()
+
+    def run(self):
+        if self.delay != 0:
+            while True:
+                if self.quit:
+                    return
+                time.sleep(self.delay / 1000.0)
```
yizow, line 111: potentially inaccurate, as the number of processes/ threads running on the beaglebone scales up.
May not be an issue at all in practice, though.

Using signals wouldnt work here. Signals + threads leads to bad things in python.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+                param_types = [hm.paramMap[hm.uid_to_device_id(self.uid)][param][1] for param in self.params]
+                params_and_values = {}
+                for param, param_type in zip(self.params, param_types):
+                    if param_type in ("bool", ):
+                        params_and_values[param] = random.choice([True, False])
```
yizow, line 116: So the fake device thread return random values? 
How is this even tested for correctness?

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,142 @@
...
+    def __init__(self, uid, delay, params, fake_device_queue):
+        self.uid = uid
+        self.delay = delay
+        self.params = params
+        self.fake_device_queue = fake_device_queue
+        self.quit = False
+        super().__init__()
+
+    def run(self):
+        if self.delay != 0:
```
yizow, line 107: Unsubscribing is not tested, right?

It looks like subscribing to a device starts a new thread that runs until told to quit. It doesn't look like it handles polling.

hibike/hibike_simulator.py
```diff
@@ -1,133 +1,144 @@
...
 if __name__ == "__main__":
-    hi = Hibike()
-    hi.subToDevices([('0x000100FFFFFFFFFFFFFFFF', 1)])
+    pipeToChild, pipeFromChild = multiprocessing.Pipe()
+    badThingsQueue = multiprocessing.Queue()
+    stateQueue = multiprocessing.Queue()
+    newProcess = multiprocessing.Process(target=hibike_process, name="hibike_sim", args=[badThingsQueue, stateQueue, pipeFromChild])
+    newProcess.daemon = True
+    newProcess.start()
+    pipeToChild.send(["ready", []])
```
yizow: remove

yizow (changes requested):
This commit is massive. It should be broken up into many different commits, on many different branches, and you should make many different pull requests. There are very clear feature distinctions that should make the separate branches logically clear. 

Please make sure to write a full commit message. Possibly squash a lot of your bug fixing commits together before you pull request.

hibike/README.md
```diff
@@ -46,265 +104,271 @@ Each Smart Device will be assigned an 88-bit UID with the following data.
...
+|         |                | 3            | enable1    | bool       | yes   | yes    |
+|         |                | 4            | servo2     | uint8_t    | yes   | yes    |
+|         |                | 5            | enable2    | bool       | yes   | yes    |
+|         |                | 6            | servo3     | uint8_t    | yes   | yes    |
+|         |                | 7            | enable3    | bool       | yes   | yes    |
+|  0x08   | LinearActuator |              |            |            |       |        |
+|  0x09   | ColorSensor    |              |            |            |       |        |
+|  0x10   | DistanceSensor |              |            |            |       |        |
+|  0x11   | MetalDetector  |              |            |            |       |        |
+|  0xFFFF | ExampleDevice  | 0            | kumiko     | bool       | yes   | yes    |
```
vdonato, line 202: kyaa~

hibike/hibikeDevices.json
yizow: Yeah, same. My comment was a suggestion on how to change it.

hibike/hibike_simulator.py
karthik-shanmugam: This is not an IPC Queue; it's just a thread safe Queue data structure.

hibike/hibike_simulator.py
karthik-shanmugam: hibike_message isn't in the python path when you try importing this from the runtime directory

hibike/hibike_simulator.py
karthik-shanmugam: This only supports subscribing. Unsubscribing in hibike is just subscribing with a delay of 0, which this also supports.

PR #65, opened 2016-11-02
[RUNTIME] Beaglebone Compatability fixes.

cjsanchez:
This corrects some halting that was occuring when running runtime.py on
the Beaglebone due to it having only 1 CPU. This also fixes a bug that
caused runStudentCode to fail when the studentCode ran longer than the
set frequency for the code to rerun. Also adds files "disable_CPUs.sh"
and "enable_CPUs.sh" for simulating the Beaglebone when testing.

In order to check that your code would run on the Beaglebone, run
disable_CPUs.sh in order to leave only one CPU running on your machine
to simulate the Beaglebone or run it on the Beaglebone. You can then
reenable them by running enable_CPUs.sh whenever you want. This only
works on Linux as far as I know.


runtime/testy/runtime.py
```diff
@@ -47,11 +47,14 @@ def runtime():
       print("Starting studentCode attempt: %s" % (restartCount,))
       spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode)
       while True:
-        newBadThing = badThingsQueue.get(block=True)
-        print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
-        print(newBadThing)
-        if newBadThing.event in restartEvents:
-          break
+        try:
```
nikitakit, line 51: Can you explain why this hunk was changed?

It looks like these two pieces of code should behave identically, with the old code being simpler. If that's not the case that means there's something interesting going on, that deserves careful thought.

runtime/testy/runtime.py
```diff
@@ -96,16 +99,18 @@ def checkTimedOut(func, *args):
       checkTimedOut(mainFunc)
       nextCall += 1.0/RUNTIME_CONFIG.STUDENT_CODE_HZ.value
       stateQueue.put([SM_COMMANDS.STUDENT_MAIN_OK, []])
-      time.sleep(nextCall - time.time())
-
+      try:
+        time.sleep(nextCall - time.time())
+      except ValueError:
+        nextCall = time.time()
```
nikitakit: Nice catch!

Compare with Ansible.py, which does not suffer from this error. (Though it might be a good idea to use the same style here and in Ansible.py, and also grep the code for any more instances of the buggy behavior)

runtime/testy/runtime.py
```diff
@@ -172,10 +177,13 @@ def runtimeTest():
             break
           spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode, testName, 3)
           while True:
-            newBadThing = badThingsQueue.get(block=True)
-            print(newBadThing.event)
-            if newBadThing.event in restartEvents:
-              break
+            try:
```
nikitakit, line 188: Again, could you explain why this code works but the old code doesn't?

runtime/testy/test_outputs/hibikeSubscribeDevice_output
```diff
@@ -1,14 +1,14 @@
-0
 1
 2
+3
```
nikitakit, line 3: I'm really confused by this. The old output must have been generated by actually running the code... as has this new output. But what has changed?

This points to some subtle non-determinism in the runtime. Best-case scenario is that the test-case framework is subtly flawed (@yizow: you should remember that I warned about non-determinism when this particular style of test framework was introduced). The worst-case scenario is that there is some subtle timing bug lurking in runtime.

runtime/testy/studentCode.py
```diff
@@ -80,4 +80,5 @@ def hibikeSubscribeDevice_setup():
 
 def hibikeSubscribeDevice_main():
   Robot.hibikeSubscribeDevice(1, 2, [3, 4])
+  time.sleep(0.01)
```
nikitakit: Is this statement necessary?

Using sleep statements to prevent race conditions is a code smell. Sleeping usually only masks the underlying issue, rather than actually fixing it.

runtime/testy/disable_CPUs.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
nikitakit, line 1: :+1: for having testing conditions that more closely approximate production!

nikitakit (commented):
Good job on testing with real hardware, and then figuring out how to replicate the same issues locally!

The `nextCall` fixes are definitely important. However, the remaining changes make me suspicious about what's going on. The runtime should have been designed with a single-core system in mind, so it's really suspicious that it would fail on the Beaglebone. (If the remaining changes were for debugging only, you should remove them from this PR and disregard the rest of this comment)

I strongly caution you guys to find the root cause of any timing issues, and address them in a principled way. You should make note of any tiny but crucial details in the code comments, or an associated design document. If you don't understand what's going on, it's really easy to sweep bugs under the rug only to have them resurface during competition. (Timing errors are notoriously hard to reproduce, so when they do come up the root cause should be carefully analyzed, and then the entire codebase should be checked for similar bugs.) And if you do understand exactly what's happening, there should be some associated docs such that the issues are not accidentally reintroduced by other people in future commits.

runtime/testy/studentCode.py
cjsanchez: Not necessary, just had it to check whether the value was supposed to change after a device was subscribed and whether it was a timing issue that kept the correct value from being read.

runtime/testy/runtime.py
cjsanchez: I found this a bit strange also. When running on the Beaglebone the main studentCode function would run two iterations and then hang forever. After trying to locate where the error was occurring it was happening after the second call to BadThingsQueue.put() where for some reason the object wasn't added. For some reason setting a delay when BadThingsQueue.get() was called fixed this issue. It may be some type of locking issue, though I'm not completely sure why it would only occur with only 1 CPU running.

nikitakit: This should be looked into more, to figure out the root cause. If no one understands _why_ one approach works while the other doesn't, there's a good chance that the blocking issue will reappear when students start using the kit.

cjsanchez:
Closing since new updated pull request has been opened here: https://github.com/pioneers/PieCentral/pull/67

PR #66, opened 2016-11-03
[RUNTIME] Add command line options for running tests

jiangdaniel:
- Import argparse and use it to parse -t and --test command line
  options
- Running "python3 runtime.py" will run runtime, "python3 runtime.py
  -t" will run all tests, "python3 runtime.py -t testName1 testName2"
will run the tests called testName1 and testName2
- Modify runtimeTest() to accept a list of test names


runtime/testy/runtime.py
```diff
@@ -218,6 +228,11 @@ def startHibike(badThingsQueue, stateQueue, pipe):
   except Exception as e:
     badThingsQueue.put(BadThing(sys.exc_info(), str(e)))
 
-if __name__ == "__main__":
-  runtimeTest()
+parser = argparse.ArgumentParser()
```
yizow: You'll want to put all this code inside an if **name** == "**main**" block
That ensures that this code will only be run if this is the main file being executed, and that if you import runtime.py, argument parsing will not run.

yizow (changes requested):
Clever code. I like it.

Just one fix: keep the **main** check in the code.

yizow (approved):

PR #67, opened 2016-11-04
Runtime/yizhe/beagle compatibility

yizow:


runtime/testy/runtime.py
```diff
@@ -47,11 +48,14 @@ def runtime():
       print("Starting studentCode attempt: %s" % (restartCount,))
       spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode)
       while True:
-        newBadThing = badThingsQueue.get(block=True)
```
nikitakit, line 51: This code is duplicated with `runtimeTest`... and this PR only fixes the bugs in the `runtimeTest` version. You should at least keep the two in sync, and preferably de-duplicate the code asap.

runtime/testy/runtimeUtil.py
```diff
@@ -59,7 +59,7 @@ def __new__(cls):
   GET_TIME            = ()
 
 class RUNTIME_CONFIG(Enum):
-  STUDENT_CODE_TIMEOUT        = 3
+  STUDENT_CODE_TIMEOUT        = 1
```
nikitakit: BTW the name `STUDENT_CODE_TIMEOUT` is too similar to `BAD_EVENTS.STUDENT_CODE_TIMEOUT`. Also, why is `RUNTIME_CONFIG` an Enum? That makes no sense. I strongly recommend cleaning this up in a future PR.

nikitakit (changes requested):

cjsanchez:
I just added the changes made to runtimeTest() to runtime() and tested that everything ran correctly on the beaglebone. Everything looks good.

runtime/testy/runtime.py
```diff
@@ -66,16 +73,23 @@ def runtime():
 def runStudentCode(badThingsQueue, stateQueue, pipe, testName = "", maxIter = 0):
   try:
     import signal
+
+    terminated = False
+    def sigTermHandler(signum, frame):
+      global terminated
```
bodiddles: Shouldn't this be nonlocal as we only want to change it in the parent frame and not global

runtime/testy/runtime.py
```diff
@@ -172,18 +185,29 @@ def runtimeTest():
...
+              if newBadThing.event in restartEvents:
+                break
+            except Exception as e:
+              print(e)
           stateQueue.put([SM_COMMANDS.RESET, []])
-          os.kill(allProcesses[PROCESS_NAMES.STUDENT_CODE].pid, signal.SIGKILL)
+          student_process = allProcesses[PROCESS_NAMES.STUDENT_CODE]
+          student_process.terminate()
+          time.sleep(.01) # Give the OS a chance to terminate the other process
+          if student_process.is_alive():
```
bodiddles: This code is a repeat of above, we should just have a function that handles this, even if it is not ideal.

bodiddles (changes requested):

runtime/testy/runtime.py
cjsanchez: fixed

nikitakit (approved):
LGTM

yizow:
Force pushed after rebasing. Nothing was changed; all tests still pass.

bodiddles (approved):

PR #57, opened 2016-10-13
[DevOps] Replacing SSH public key authentication

jonathan-j-lee:
- Replaced public key authentication with plain password authentication in the master SSHD config file.
- Wrote a script (to be run on staff machines) to: 
  - Copy the new config file to a connected BeagleBone.
  - Create a backup of the old config file.
  - Restart the SSH service.

PR #68, opened 2016-11-05
[DevOps] Reconfigure SSHD (Updated)

jonathan-j-lee:
Since #57 :

* Enables the user to pick the identity file


ArjunSridhar1 (approved):
Nice job!

PR #69, opened 2016-11-06
[Networking] Basic message server and driver install script

Alramech:
- currently requires manual ip input, defualt localhost

- client closes after sending message, to be changed to persistent

### DIFF
--- a/networking/client.py
+++ b/networking/client.py
@@ -1,10 +1,22 @@
 import socket
+import sys
+import argparse
 
+parser = argparse.ArgumentParser()
+parser.add_argument("-p", "--persistent", help="persistent sending mode",
+                    action="store_true")
+parser.add_argument("--ip", default='localhost',
+                    help="destination ip, default localhost")
+args = parser.parse_args()
+
+
+HOST = args.ip
 PORT = 3141
-HOST = 'localhost'
-
-with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-    s.connect((HOST, PORT))
-    msg = bytes(input('message: '), 'UTF8')
-    s.sendall(msg)
+while True:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.connect((HOST, PORT))
+        msg = bytes(input('message: '), 'UTF8')
+        s.sendall(msg)
+    if not args.persistent:
+        break
 
--- a/networking/client.py
+++ b/networking/client.py
@@ -1,0 +1,13 @@
+import socket
+
+s = socket.socket(
+    socket.AF_INET, socket.SOCK_STREAM)
+
+s.connect(('localhost', 3141))
+
+msg = raw_input('message: ')
+while msg != '':
+    sent = s.send(msg)
+    msg = msg[sent:]
+s.shutdown(socket.SHUT_RDWR)
+s.close()
--- a/networking/client.py
+++ b/networking/client.py
@@ -1,5 +1,4 @@
 import socket
-import sys
 import argparse
 
 parser = argparse.ArgumentParser()
@@ -7,11 +6,13 @@ parser.add_argument("-p", "--persistent", help="persistent sending mode",
                     action="store_true")
 parser.add_argument("--ip", default='localhost',
                     help="destination ip, default localhost")
+parser.add_argument("--port", default='3141',
+                    help="connection port, default 3141")
 args = parser.parse_args()
 
 
 HOST = args.ip
-PORT = 3141
+PORT = int(args.port)
 while True:
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
         s.connect((HOST, PORT))
--- a/networking/client.py
+++ b/networking/client.py
@@ -1,13 +1,10 @@
 import socket
 
-s = socket.socket(
-    socket.AF_INET, socket.SOCK_STREAM)
+PORT = 3141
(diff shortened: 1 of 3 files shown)


networking/client.py
```diff
@@ -0,0 +1,13 @@
+import socket
+
+s = socket.socket(
```
yizow: why split this into multiple lines?

networking/client.py
```diff
@@ -0,0 +1,13 @@
...
+    socket.AF_INET, socket.SOCK_STREAM)
+
+s.connect(('localhost', 3141))
```
yizow: if 3141 is going to be a constant, bind it to a global variable after your import statements.

networking/client.py
```diff
@@ -0,0 +1,13 @@
...
+
+msg = raw_input('message: ')
+while msg != '':
+    sent = s.send(msg)
+    msg = msg[sent:]
+s.shutdown(socket.SHUT_RDWR)
+s.close()
```
yizow: add newline at end of file

networking/server.py
```diff
@@ -0,0 +1,15 @@
+import socket
+
+s = socket.socket(
```
yizow: Same comment as above

networking/server.py
```diff
@@ -0,0 +1,15 @@
...
+    socket.AF_INET, socket.SOCK_STREAM)
+
+s.bind(('localhost', 3141))
```
yizow: Same comment as above
Use global costants.

networking/server.py
```diff
@@ -0,0 +1,15 @@
...
+s.listen(5)
```
yizow: Why choose 5? Leaving it blank lets the system choose a "reasonable value"

networking/server.py
```diff
@@ -0,0 +1,15 @@
...
+while 1:
+    (cs, addr) = s.accept()
+    msg = ''
+    rec = cs.recv(1024)
+    while rec:
+        msg += rec
+        rec = cs.recv(1024)
+    print(msg)
```
yizow: New line at end of file

networking/client.py
yizow: Using `with s as socket.socket(...):` autoamtically calls close at end of block

networking/server.py
```diff
@@ -0,0 +1,15 @@
+import socket
+
+s = socket.socket(
+    socket.AF_INET, socket.SOCK_STREAM)
+
+s.bind(('localhost', 3141))
+s.listen(5)
+while 1:
+    (cs, addr) = s.accept()
```
yizow: I recommend renaming `cs` to `conn`. More readable.

networking/client.py
```diff
@@ -0,0 +1,13 @@
+import socket
+
+s = socket.socket(
+    socket.AF_INET, socket.SOCK_STREAM)
+
+s.connect(('localhost', 3141))
+
+msg = raw_input('message: ')
```
yizow: we're using python3.4 currently. `raw_input` is deprecated and has been renamed to `input`

networking/client.py
```diff
@@ -0,0 +1,13 @@
...
+while msg != '':
+    sent = s.send(msg)
```
yizow: This will no longer work in python3. msg must support buffer interface, so use `bytes(msg)`

yizow (changes requested):
Main issue: Update to python 3.4

This is a lot more barebones that I was hoping for :-/

Have you tested this running on different machines? 
Or can you spin up more than just two processes: Simulate a (Driver station, field control, robot) network?
After making such a simulation, how would you verify it works? Then, write a test script that does that verfication.

networking/server.py
Alramech: That's a 3.5 thing, not a 3.4. 5 is usually the system maximum value.

networking/server.py
```diff
@@ -1,15 +1,16 @@
 import socket
 
-s = socket.socket(
-    socket.AF_INET, socket.SOCK_STREAM)
+HOST = 'localhost'
```
yizow: why are port and host in the opposite ordering here ._.

networking/server.py
```diff
@@ -1,15 +1,16 @@
...
-    print(msg)
\ No newline at end of file
+with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+    s.bind((HOST, PORT))
+    s.listen(5)
+    while 1:
+        conn, addr = s.accept()
+        msg = bytes()
+        rec = conn.recv(1024)
+        while rec != bytes():
```
yizow: i think `while rec` works, and is cleaner. You should check. I'm going off the python doc example which is `if not data: break`

yizow (changes requested):

networking/server.py
```diff
@@ -0,0 +1,16 @@
...
+            msg += rec
+            rec = conn.recv(1024)
+        print(msg.decode('UTF8'))
```
yizow, line 25: Do we have a newline at end of file?

yizow (changes requested):

yizow (approved):
Good work. In future patches, i'd like to see: 
1. Some sort of test. Send things, print things to a file, examine file. AUTOMATED.
2. More than 2 clients.
3. An interactive mode where we can send user-input messages. This will make debugging and testing the performance of runtime/ dawn super super useful. 

Merge when you're ready.

PR #70, opened 2016-11-09
Fake Runtime on Dawn Side

kevinmasd:
Testing currently consists of toggling a variable on every Runtime update, sending the toggle to Runtime back.


dawn/fake-runtime/FakeRuntime.js
```diff
@@ -8,15 +8,23 @@
 
 const dgram = require('dgram');
 const ProtoBuf = require('protobufjs');
+// const net = require('net');
```
yizow: Remove unnecessary commented out lines.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -8,15 +8,23 @@
...
 
-const builder = ProtoBuf.loadProtoFile('../ansible-protos/ansible.proto');
-const DawnData = builder.build('DawnData');
+const dawnBuilder = ProtoBuf.loadProtoFile('../ansible-protos/ansible.proto');
+const DawnData = dawnBuilder.build('DawnData');
+const runtimeBuilder = ProtoBuf.loadProtoFile('../ansible-protos/runtime.proto');
+const RuntimeData = runtimeBuilder.build('RuntimeData');
+// const TCPData = runtimeBuilder.build('TCPData');
```
yizow: Remove.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -8,15 +8,23 @@
...
 
-const clientPort = '12346'; // send port
-const serverPort = '12345'; // receive port
+const clientPort = 12346; // send port
+const serverPort = 12345; // receive port
+// const tcpPort = 12347;
```
yizow: Remove.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -8,15 +8,23 @@
...
 const hostname = 'localhost';
 const client = dgram.createSocket('udp4');// sender
 const server = dgram.createSocket('udp4'); // receiver
+// const tcp = net.connect({port: tcpPort}, () => {
```
yizow: Remove.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -40,73 +48,32 @@ const randomFloat = (min, max) => ((max - min) * Math.random() + min);
...
-      value: Math.round(randomFloat(0, 1)),
-      id: '2',
-    },
-  },
-  {
-    type: 'UPDATE_PERIPHERAL',
-    peripheral: {
-      name: 'Scalar Sensor 1',
-      peripheralType: 'SENSOR_SCALAR',
+    // type: 'UPDATE_STATUS',
```
yizow: Remove.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -40,73 +48,32 @@ const randomFloat = (min, max) => ((max - min) * Math.random() + min);
...
+    sensor_data: [{
+      device_type: 'MOTOR_SCALAR',
+      device_name: 'Motor 1',
       value: randomFloat(-100, 100),
-      id: '3',
     },
+      {
+        device_name: 'Limit Switch 1',
+        device_type: 'LimitSwitch',
+        value: Math.round(randomFloat(0, 1)),
```
yizow, line 55: Sending random floats makes testing really annoying.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -40,73 +48,32 @@ const randomFloat = (min, max) => ((max - min) * Math.random() + min);
...
-  // Don't know what Runtime protobufs look like so
-  // we are encoding with JSON for right now.
-  const encodedData = JSON.stringify(fakeData);
-  client.send(encodedData, clientPort, hostname);
-}, 500);
+  for (const item of fakeData) {
+    const udpData = new RuntimeData(item);
+    client.send(Buffer.from(udpData.toArrayBuffer()), clientPort, hostname);
+  }
+}, 1000);
```
yizow: Make the interval a constant somewhere.

dawn/main/ansible/Ansible.js
```diff
@@ -16,12 +19,31 @@ const StudentCodeStatus = DawnData.StudentCodeStatus;
 
 const runtimeBuilder = ProtoBuf.loadProtoFile(`${protoFolder}/runtime.proto`);
 const RuntimeData = runtimeBuilder.build('RuntimeData');
+const TCPData = runtimeBuilder.build('TCPData');
+
+let test = 0;
```
yizow, line 25: let declares test to be block scope, and automatically unbinds it when you leave the block.
1. This is global, not a block, so there's not much point of using let.
2. Is unbinding behaviour what you want?

yizow: Actually, more reading shows this wont work: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let
`At the top level of programs and functions, let, unlike var, does not create a property on the global object.`

yizow: And if youre not using test, this line should be removed.

dawn/main/ansible/Ansible.js
```diff
@@ -55,15 +77,40 @@ ipcMain.on('stateUpdate', (event, data) => {
...
+    },
+    peripheral_names: ['Test'],
+  });
+  const buffer = message.encode().toBuffer();
+  client.send(buffer, clientPort, hostname, (err) => {
+    if (err) {
+      console.error('UDP socket error on send:', err);
+    }
+  });
+}, 1000);
```
yizow: use a constant

dawn/main/ansible/Ansible.js
```diff
@@ -55,15 +77,40 @@ ipcMain.on('stateUpdate', (event, data) => {
...
 /**
  * Handler to receive messages from the robot Runtime
  */
 server.on('message', (msg) => {
   try {
-    const data = RuntimeData.decode(msg).toRaw();
-    console.log(`Dawn received: ${JSON.stringify(data)}\n`);
+    const data = RuntimeData.decode(msg);
+    console.log(`Dawn received Runtime Information at ${(new Date).toISOString()}`);
+    test = 1 - test;
```
yizow, line 106: i think test is unbound here

dawn/renderer/utils/sagas.js
```diff
@@ -195,6 +195,7 @@ function formatGamepads(newGamepads) {
  */
 function* updateGamepads() {
   while (true) {
+    console.log('Hello');
```
yizow: How often is this going to be printed? Do you want to rate limit this?

yizow (changes requested):
Full disclosure: I'm not entirely sure what i'm looking at :-/

My comments were focuses on cleaning up the code.

dawn/main/ansible/Ansible.js
AllanYangZhou: Our current linting rules prefer `let` over `var`, as its behavior is generally considered more predictable (that's subjective though).

dawn/main/ansible/Ansible.js
AllanYangZhou: The scope of `test` here is global (test was defined on global scope), so I think it should be fine.

dawn/renderer/utils/sagas.js
AllanYangZhou: This is already limited to 20 times/second due to delay on line 211/212.

AllanYangZhou: @kevinmasd Though you should remove this before merging.

dawn/main/ansible/Ansible.js
AllanYangZhou: I agree that the use of the `test` variable here is largely unnecessary, why not just print what you receive for this test?

AllanYangZhou (commented):
I see in Ansible.js there is code to dispatch received sensor data to the frontend. Is that whole flow (from FakeRuntime all the way to the frontend) work? Does that mean that toggling the test variable is no longer necessary?

dawn/main/ansible/Ansible.js
yizow: Isn't this top level though? According to that doc i linked, let literally will not bind `test` to anything.

dawn/main/ansible/Ansible.js
yizow: See above regarding let not binding things in the global scope

dawn/main/ansible/Ansible.js
AllanYangZhou: My understanding of the doc you linked is that `let` in global scope does not bind the variable to the _global object_. Note that all properties of the global object are also global variables, but not all global variables are necessarily properties of the global object.

This distinction is minor, the only issue is if you try to do `global.test` (or `this.test` when this refers to `global`), but nobody ever does this.

AllanYangZhou: But yeah the difference is super-confusing, I thought the same thing when I read your doc but I think there's a nice-ish explanation in section 9.7 of http://exploringjs.com/es6/ch_variables.html

EDIT: actually, I think since this is in a module that is imported everything here is actually run at module scope and not global scope, per the above link.

kevinmasd: The point of test was a proof of concept that Dawn can react to proto messages without going fully fledged into a fake controller. It'll be removed after more testing verifies that what Dawn does for communication is correct.

kevinmasd:
Haven't had time to check how Dawn actually deals with dispatch data. Currently waiting on UIDs to be re-added to the proto so I can actually map peripheral data to peripheral and checking to see whether the Sagas are actually running.

dawn/main/ansible/Ansible.js
AllanYangZhou: So will that (the removing part) be before this is merged in or will it be in a later PR?

yizow (approved):

PR #73, opened 2016-11-12
[DAWN] UIDs in Protos

kevinmasd:
Peripheral Names will be solely Dawn's domain. Instead, Runtime needs to send Dawn id's of the peripherals.


ansible-protos/runtime.proto
```diff
@@ -9,7 +9,7 @@ message RuntimeData {
 	}
 	message SensorData {
 		string device_type = 1;
-		string device_name = 2;
```
bodiddles, line 12: Keep device name for concurrency and backwards compatibility

bodiddles (changes requested):

bodiddles (approved):

PR #74, opened 2016-11-13
[RUNTIME] Adds Protobuf Integration

bodiddles:
Adds packaging and unpackaging functionality for the UDP send and receive
processes for google protobufs
    
Mirrors packaging and unpackaging code given by the protobuf python guide.
    
For packaging, creates protobuf objections, parses through the state, and adds
data to the correct fields of the objects, It then serializes the data to bytes
and gets sent by the socket.
    
Due to current fields in the state dictionary, we only serialize a single limit
switch but it can be expanded to all other types of sensors in the state.
    
For unpackaging, it reads and creates a protobuf object from serialized data and
then parses it and adds it to fields in a dictionary which is then added to the state.
    
fake_dawn also mirrors this schema but with the ansible proto instead.


runtime/testy/Ansible.py
```diff
@@ -167,11 +198,25 @@ def unpackageData(self):
         function is currently unimplemented.
```
yizow: Is it still unimplemented?

runtime/testy/Ansible.py
```diff
@@ -189,9 +234,15 @@ def start(self):
 
         try:
             while True:
+                nextCall = time.time()
                 event = sel.select()
                 self.udpReceiver()
                 self.unpackageData()
+                if (nextCall > time.time()):
+                    try:
+                        time.sleep(nextCall - time.time())
```
yizow: does `time.sleep(max(nextCall - time.time(), 0))` work?

runtime/testy/fake_dawn.py
```diff
@@ -2,34 +2,50 @@
...
-				time.sleep(next_call - time.time())
+    host = '127.0.0.1'
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+        while True:
+            next_call = time.time()
+            msg = None
+            msg = dawn_packager(0)
+            s.sendto(msg, (host, send_port))
+            next_call += 1.0/dawn_hz
+            if next_call > time.time():
```
yizow: `time.sleep(max(nextCall - time.time(), 0))` might be cleaner

runtime/testy/fake_dawn.py
```diff
@@ -40,5 +56,4 @@ def receiver(port, receive_queue):
 
 #Just Here for testing, should not be run regularly
 while True:
```
yizow: put this in a `if __name__ == "__main__"` block

runtime/testy/stateManager.py
```diff
@@ -59,7 +59,9 @@ def initRobotState(self):
      "list1" : [[[70, 0], ["five", 0], [14.3, 0]], 0],
      "string1" : ["abcde", 0],
      "runtime_meta" : [{"studentCode_main_count" : [0, 0]}, 0],
-     "hibike" : [{"device_subscribed" : [0, 0]}, 0]
+     "hibike" : [{"device_subscribed" : [0, 0]}, 0],
+     "studentCodeState": [2, 0],
+     "limit_switch": [["lmsw", .5, 123456], 0]
```
yizow: why not just write out `limit_switch` instead of `lmsw`?

nikitakit:
Woa what is up with this commit history? I'm going to review, but only the aggregate diff for the full PR

ansible-protos/runtime.proto
```diff
@@ -6,12 +6,14 @@ message RuntimeData {
...
 		AUTO = 4;
+		ESTOP = 5;
 	}
 	message SensorData {
 		string device_type = 1;
 		string device_name = 2;
 		float value = 3;
-		int64 uid = 4;
+		uint64 uid = 4;
+		uint32 int_type = 5;
```
nikitakit: What is "int_type"? It's unclear what this field does, and it also doesn't seem to be used anywhere

runtime/testy/Ansible.py
```diff
@@ -3,9 +3,19 @@
 import time
 import sys
 import selectors
+import runtime_pb2
+import ansible_pb2
 
 from runtimeUtil import *
 
+stateToEnum = {0:runtime_pb2.RuntimeData.STUDENT_CRASHED,
```
nikitakit: This dict is literally a no-op. `runtime_pb2.RuntimeData.STUDENT_CRASHED` should already be an int. Same with the rest of the values here.

See also `runtime_pb2.RuntimeData.State.items()`, which actually contains a list of (string, int) key-value pairs.

runtime/testy/Ansible.py
```diff
@@ -104,10 +129,13 @@ def package(state):
                 self.sendBuffer.replace(packState)
                 nextCall += 1.0/self.packagerHZ
                 if (nextCall > time.time()):
-                    time.sleep(nextCall - time.time())
-            except Exception:
+                    try:
+                        time.sleep(nextCall - time.time())
+                    except ValueError:
```
nikitakit: Why the check for `ValueError`? I'm guessing it's because the value can be negative. In that case, the `if (nextCall > time.time())` check above is redundant, and you can remove it and replace with a comment explaining that the try-catch handles that case.

runtime/testy/Ansible.py
```diff
@@ -123,11 +151,14 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
                 try:
                     nextCall = time.time()
                     msg = self.sendBuffer.get()
-                    if msg != 0: 
+                    if msg != 0 and msg != None:
```
nikitakit: FYI canonical style is `msg is not None`, not `msg != None`

runtime/testy/Ansible.py
```diff
@@ -123,11 +151,14 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
...
                     msg = self.sendBuffer.get()
-                    if msg != 0: 
+                    if msg != 0 and msg != None: 
                         s.sendto(msg, (host, UDPSendClass.SEND_PORT))
                     nextCall += 1.0/self.socketHZ
                     if (nextCall > time.time()):
-                        time.sleep(nextCall - time.time())
+                        try:
+                            time.sleep(nextCall - time.time())
+                        except ValueError:
```
nikitakit: See previous `ValueError` comment.

runtime/testy/Ansible.py
```diff
@@ -167,11 +198,25 @@ def unpackageData(self):
...
+            through the update method implemented in state manager.
             """
-            return data 
+            unpackaged_data = {}
+            received_proto = ansible_pb2.DawnData()
+            received_proto.ParseFromString(data)
+            unpackaged_data["student_code_status"] = received_proto.student_code_status
+            for gamepad in received_proto.gamepads:
+                gamepad_dict = {}
+                gamepad_dict["axes"] = {i : axis for i, axis in zip(range(0,4), gamepad.axes)}
```
nikitakit: Use `enumerate(gamepad.axes)` instead of `zip`.

Moreover, the entire line could be `gamepad_dict["axes"] = dict(enumerate(gamepad.axes))`.

But what's even less clear is why the gamepads are stored in a dict as opposed to a list in the first place. Could you explain?

runtime/testy/Ansible.py
```diff
@@ -167,11 +198,25 @@ def unpackageData(self):
...
+                gamepad_dict["buttons"] = {i : button for i, button in zip(range(0,20),gamepad.buttons)}
```
nikitakit: Same comment as for the axes

runtime/testy/Ansible.py
nikitakit: Same comment as `ValueError` above. (No need for redundant `if` followed by try-catch. Single try-catch or what @yizow said would work)

runtime/testy/fake_dawn.py
```diff
@@ -2,34 +2,50 @@
...
+import ansible_pb2
 
 data = [0]
 send_port = 1236
 recv_port = 1235
 dawn_hz = 10
 
+def dawn_packager(data):
+    proto_message = ansible_pb2.DawnData()
+    proto_message.student_code_status = 1
```
nikitakit: Use a meaningful name instead of int, e.g. `ansible_pb2.DawnData.TELEOP`

runtime/testy/ansible_pb2.py
```diff
@@ -0,0 +1,166 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
nikitakit, line 1: Why are you committing compiled protobuf files??? This should be removed and added to `.gitignore`

runtime/testy/runtime_pb2.py
```diff
@@ -0,0 +1,217 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
nikitakit, line 1: See above: this file is auto-generated and should be ignored.

nikitakit (changes requested):
See my comments below.

Overall functionality LGTM, but I haven't looked into dawn at all, so I haven't reviewed for any accidental incompatibilities between the two endpoints. I trust that you are going to integration test this extensively.

runtime/testy/ansible_pb2.py
bodiddles: So currently, I don't think devops has the protobuf compilers on the beaglebones so we are keeping these temporarily so runtime will actually run until that build system is figured out.

nikitakit: :cry: In that case, can you report a github issue, and make a note that compiled protobuf files will need to be removed when the issue is fixed?

runtime/testy/fake_dawn.py
bodiddles: Wouldn't this be the same as the no-op in Ansible?

runtime/testy/ansible_pb2.py
brandonxxlee: Done

runtime/testy/fake_dawn.py
nikitakit: Umm, what???

This change is for human readability. It's a rule of good style to always use the name of an enum, not its number, when hardcoding constants.

runtime/testy/Ansible.py
```diff
@@ -128,11 +119,10 @@ def package(state):
                 packState = package(rawState)
                 self.sendBuffer.replace(packState)
                 nextCall += 1.0/self.packagerHZ
-                if (nextCall > time.time()):
-                    try:
-                        time.sleep(nextCall - time.time())
-                    except ValueError:
-                        continue
+                try:
+                    time.sleep(nextCall - time.time())
```
yizow: `max(nextCall - time.time(), 0)`

runtime/testy/Ansible.py
```diff
@@ -238,11 +227,10 @@ def start(self):
                 event = sel.select()
                 self.udpReceiver()
                 self.unpackageData()
-                if (nextCall > time.time()):
-                    try:
-                        time.sleep(nextCall - time.time())
-                    except ValueError:
-                        continue
+                try:
+                    time.sleep(nextCall - time.time())
```
yizow, line 225: `max(nextCall - time.time(), 0)`

runtime/testy/stateManager.py
```diff
@@ -129,6 +129,7 @@ def send_ansible(self):
 
   def recv_ansible(self, new_data):
     self.state.update(new_data)
+    print(new_data)
```
yizow: this will mess up testing. do you really need to print, or was this only for debug purposes?

yizow (changes requested):

yizow (approved):

runtime/testy/Ansible.py
```diff
@@ -141,7 +155,7 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
         self.recvBuffer = TwoBuffer()        
         packName = THREAD_NAMES.UDP_UNPACKAGER
         sockRecvName = THREAD_NAMES.UDP_RECEIVER
-        host = '127.0.0.1' #TODO: determine host between dawn-runtime comm
+        host = "" #TODO: determine host between dawn-runtime comm
```
nikitakit: Why is empty string an acceptable value for host?

runtime/testy/studentCode.py
```diff
@@ -129,7 +129,6 @@ def infiniteMainLoop_main():
 def emergencyStop_setup():
   print("E-Stop setup")
 
-
```
nikitakit, line 132: pointless whitespace change

runtime/testy/fake_dawn.py
```diff
@@ -2,34 +2,52 @@
...
-			msg = bytes(b)
-			s.sendto(msg, (host, send_port))
-			next_call += 1.0/dawn_hz
-			if next_call > time.time():
-				time.sleep(next_call - time.time())
+    host = '127.0.0.1'
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+        while True:
+            next_call = time.time()
+            msg = None
```
nikitakit: This line should be removed (because it's redundant with the next line as it is)

runtime/testy/fake_dawn.py
```diff
@@ -2,34 +2,52 @@
...
 import time
+import runtime_pb2
+import ansible_pb2
 
 data = [0]
 send_port = 1236
 recv_port = 1235
 dawn_hz = 10
 
+def dawn_packager(data):
```
nikitakit: Why are you taking a data argument and not using it?

runtime/testy/fake_dawn.py
```diff
@@ -2,34 +2,52 @@
...
-			s.sendto(msg, (host, send_port))
-			next_call += 1.0/dawn_hz
-			if next_call > time.time():
-				time.sleep(next_call - time.time())
+    host = '127.0.0.1'
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+        while True:
+            next_call = time.time()
+            msg = None
+            msg = dawn_packager(0)
```
nikitakit: Why pass 0 as an argument here?

nikitakit (changes requested):

runtime/testy/Ansible.py
nikitakit: Nevermind, `""` is same as `0.0.0.0` but is apparently the preferred style.

nikitakit: However, the `TODO` comment on this line is probably a mistake. I think you want a comment that says "binds on all interfaces" without a TODO

nikitakit (approved):

PR #78, opened 2016-11-17
[RUNTIME] Emergency stop implemented

yizow:
Implemented handling for emergency stop in statemanager.py. Added
boolean in runtime_meta of self.state to keep track of e_stop state.
When in e_stop, any commands to hibike are overridden into an
emergency stop command. E_stop state is initiated through a command
to state manager.
    
In order to implement e_stop, updated runtimeUtil.py by adding
commands into HIBIKE_COMMANDS and SM_COMMANDS for e_stop. Also
added a BadEvent specifically for emergency stop.

For testing, added test function to send the stop command in
studentAPI.py and wrote a test function in studentCode.py

Added boolean in runtime.py to signify whether the robot
has been e_stopped and if so, student code will not attempt to
restart.

Finally changed the terminate_process method in runtime.py to wait for a longer amount of time.


greenkee (commented):
Recommitting to fix commit message

greenkee (approved):

PR #79, opened 2016-11-17
[DAWN] New Actions, Protobuf-Based Peripheral Display

kevinmasd:
Due to how protobuf.js interpreted longs, various variable references were changed. Actions also replace various ad-hoc versions, and minor visual changes elsewhere. Color sensors disabled. Runtime IP specified as new variable.


dawn/main/ansible/Ansible.js
```diff
@@ -4,8 +4,15 @@ import RendererBridge from '../RendererBridge';
...
+import {
+    ansibleConnect,
+    ansibleDisconnect,
+    updateStatus,
+} from '../../renderer/actions/InfoActions';
+import { updatePeripheral } from '../../renderer/actions/PeripheralActions';
 
 const hostname = 'localhost';
+const runtimeIP = 'localhost';
 const clientPort = 12345; // send port
```
brandonxxlee: Noticed while testing yesterday, these ports do not match the ports we have on the runtime side. We are sending to 1235 and we are receiving from 1236 (so you should be the mirror of that).

dawn/renderer/components/FinalCompPeripheralList.js
```diff
@@ -22,11 +22,11 @@ const FinalCompPeripheralList = (props) => {
         _.toArray(props.peripherals),
         (peripheral) => (
           <Peripheral
-            key={peripheral.id}
-            id={peripheral.id}
-            name={peripheral.name}
+            key={peripheral.uid.low}
```
nikitakit: Throwing away the upper bits of the device UID seems extremely suspect.

If the UID is ever displayed in the UI, it will be truncated and therefore not the same thing that staff would see on any debug tools.

If it's ever used for non-display purposes (e.g. name-to-uid mapping), then throwing away those bits may have even worse consequences in terms of making the mapping unusable for runtime/hibike.

dawn/renderer/components/FinalCompPeripheralList.js
```diff
@@ -22,11 +22,11 @@ const FinalCompPeripheralList = (props) => {
...
+            id={peripheral.uid.low}
```
nikitakit: Again, throwing away bits seems suspect

dawn/renderer/reducers/info.js
```diff
@@ -28,7 +28,6 @@ const info = (state = {}, action) => {
     case 'UPDATE_STATUS':
       return {
         ...state,
-        isRunningCode: (action.status.value === 1),
```
AllanYangZhou, line 31: I assume that sometime in the future this functionality will need to be added back right? We will need some way of keeping track of if code is running or not eventually I assume.

AllanYangZhou (changes requested):

dawn/renderer/reducers/info.js
kevinmasd: Yes. It is currently modified as otherwise, we wouldn't be able to show peripheral display.

AllanYangZhou: Ok, sounds good.

AllanYangZhou (approved):

PR #82, opened 2016-11-28
[RUNTIME] Deduplicated runtime code

yizow:
The main code for runtime is now deduplicated. Previously, runtime() and
runtimeTest() had 2 different copies of largely the same code. Those
lines have been merged into the new runtime() function.

This commit also fixes a few bugs that were causing tests to fail:
1. E-stop test did not actually have an output file
2. Ansible had a typo with parenthesis
3. Ansible processes are terminated in between tests to release their
port.


runtime/testy/runtime.py
```diff
@@ -42,31 +45,37 @@ def runtime():
     spawnProcess(PROCESS_NAMES.UDP_RECEIVE_PROCESS, startUDPReceiver)
     spawnProcess(PROCESS_NAMES.HIBIKE, startHibike)
     while True:
-      if restartCount >= 5 or emergency_stopped:
-        print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
-        print("Too many restarts, terminating")
+      if restartCount >= 3 or emergency_stopped:
+        if not testMode:
+          print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
+          print("Too many restarts, terminating")
```
nikitakit, line 51: Just saw the `if emergency_stopped` condition above this print statement. At some point, you should make the printed messages match up with what's actually happening (i.e. estop is not the same as too many restarts)

runtime/testy/runtime.py
```diff
@@ -30,7 +30,10 @@
 
 allProcesses = {}
 
-def runtime():
+def runtime(testName=""):
+  testMode = testName != ""
+  maxIter = 3 if testMode else 0
```
nikitakit: Again unrelated to the current changeset, but `maxIter` of 0 meaning unlimited is highly misleading. While this is standard practice in C due to the limits of the type system, for Python you may as well use `None` to make it clear that this is a signalling value.

runtime/testy/runtime.py
```diff
@@ -42,31 +45,37 @@ def runtime():
...
-        print(newBadThing.data)
+        if not testMode:
+          print(newBadThing.data)
         if newBadThing.event in restartEvents:
-          if(not emergency_stopped and newBadThing.event is BAD_EVENTS.EMERGENCY_STOP):
+          if (not emergency_stopped and newBadThing.event is BAD_EVENTS.EMERGENCY_STOP):
             emergency_stopped = True #somehow kill student code using other method? right now just restarting on e-stop
           break
       stateQueue.put([SM_COMMANDS.RESET, []])
-      terminate_process(allProcesses[PROCESS_NAMES.STUDENT_CODE])
```
nikitakit, line 61: Wait... the old code on this line would crash really badly, right?

runtime/testy/runtime.py
```diff
@@ -42,31 +45,37 @@ def runtime():
...
+    if not testMode:
+      print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
     print("Funtime Runtime is done having fun.")
     print("TERMINATING")
-  except:
+  except Exception as e:
     print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
     print("Funtime Runtime Had Too Much Fun")
-    print(traceback.print_exception(*sys.exc_info()))
+    print(e)
```
nikitakit, line 79: I trust that you know what you're doing with regards to changing the exception printing code. (TBH I don't know what the difference between the old and the new code on this line is)

runtime/testy/runtime.py
```diff
@@ -191,38 +201,10 @@ def runtimeTest(testNames):
...
-          restartCount += 1
-        print("Funtime Runtime is done having fun.")
-        print("TERMINATING")
-      except Exception as e:
-        print("Funtime Runtime Had Too Much Fun")
-        print(e)
-        print("".join(traceback.format_tb(sys.exc_info()[2])))
+      runtime(testName)
+
+      terminate_process(PROCESS_NAMES.UDP_SEND_PROCESS)
```
nikitakit, line 210: I think you want a comment about why these two terminate lines are here. (I can see that it's due to having multiple tests, but it took me a bit of cross-referencing other chunks of code to figure that out)

nikitakit (changes requested):
This changeset LGTM, but I'd like to see a comment added about UDP termination.

However a look at the latest version of the code has brought to my attention two other comments that I think should have been brought up in earlier reviews. I would like to see these added to the github issue tracker or addressed in future commits.

runtime/testy/runtime.py

runtime/testy/runtime.py
yizow: Fixed

runtime/testy/runtime.py
yizow: I moved the dictionary lookup into the terminate_process function. This way i can simultaenously remove a terminated process from all processes. 
Before, this code would just overwrite the old value in the dictionary when the new process is started.

runtime/testy/runtime.py
yizow: `except` and `except Exception as e` match the same cases, but this new code also exposes the exception object.
This lets the printed trace be more descriptive.

runtime/testy/runtime.py
yizow: Fixed

yizow (commented):
Addressed changes.

runtime/testy/runtime.py
```diff
@@ -42,33 +45,43 @@ def runtime():
...
+      if not testMode:
+        if restartCount >= 3:
+          print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
+          print("Too many restarts, terminating")
+          break
+        if emergency_stopped:
+          print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
+          print("terminating due to E-Stop")
+          break
+      if not testMode:
```
nikitakit: Collapse out repeated `if not testMode`

runtime/testy/runtime.py
```diff
@@ -103,7 +116,7 @@ def checkTimedOut(func, *args):
 
     # TODO: Replace execCount with a value in stateManager
     execCount = 0
-    while (not terminated) and (maxIter == 0 or execCount < maxIter):
+    while (not terminated) and (maxIter == None or execCount < maxIter):
```
nikitakit: `is None`

nikitakit (changes requested):
Minor style things

nikitakit (approved):

PR #84, opened 2016-12-01
[DAWN] Revert IP and Ports back to Production

kevinmasd:
Values temporarily changed during integration test were never reverted back.


yizow (approved):
Making he ip configurable from within dawn is on the to-do list, right?

kevinmasd:
Gave it to Jiana, I believe.

PR #87, opened 2016-12-21
Hibike/protocol2.0 refactor

yizow:
Updating files to copy Hibike internal spec defined in example_device.*


hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,138 @@
...
-// the return value is the value field of the DeviceRespond packet hibike will respond with
-uint32_t device_update(uint8_t param, uint32_t value) {
-  if (param < NUM_PARAMS) {
-    params[param] = value;
-    return params[param];
+// You must implement this function.
+// It is called when the device receives a Device Write packet.
+// Updates param to new value passed in data.
+//    param   -   Parameter index
+//    data    -   value to write, in bytes TODO: What endian?
```
karthik-shanmugam: Little endian, which is the endianness of the Arduino

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,138 @@
...
-uint32_t device_status(uint8_t param) {
-  if (param < NUM_PARAMS) {
-    return params[param];
+
+// You must implement this function.
+// It is called when the device receives a Device Data Update packet.
+// Modifies data_update_buf to contain the parameter value.
+//    param           -   Parameter index
+//    data_update_buf -   buffer to return data in
+//    buf_len         -   Maximum length of the buffer? TODO: Clarify
```
karthik-shanmugam: buf_len is the number of bytes left in the buffer, so this function only writes to the buffer if their is sufficient space left

yizow:
When reviewing, my be useful to review commit by commit if the aggregate is too confusing.

The following sensors should be updated (they compile): 

- ServoControl
- ExampleDevice
- LimitSwtich
- LineFollower
- Potentiometer
- YogiBear
- TeamFlag

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,116 @@
...
-  append_buf(data_update_buf, &offset, (uint8_t *)&value0, sizeof(value0));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value1, sizeof(value1));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value2, sizeof(value2));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value3, sizeof(value3));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value4, sizeof(value4));
-  return offset;
-}
+//    return          -   sizeof(param) on success; 0 otherwise
 
+uint8_t device_data_update(uint8_t param, uint8_t* data_update_buf, size_t buf_len) {
```
karthik-shanmugam, line 86: `buf_len` is the size of the buffer to prevent buffer overflow. Why is it being passed as the num argument to `write_num_bytes()` ?

hibike/lib/hibike/util.cpp
```diff
@@ -0,0 +1,20 @@
...
+  for (size_t i = 0; i < num; i++) {
+    value += data[i] << (8 * i);
+  }
+  return value;
+}
+
+
+// Assumes little endian
+
+void write_num_bytes(uint64_t data, uint8_t* data_buf, size_t num) {
```
karthik-shanmugam, line 16: If we use this function to write different data types to a buffer, could the cast to uint64_t cause incorrect behavior when dealing with signed integers / floating point numbers? Why not just pass a pointer to data instead of casting and doing bit operations?

yizow: This function should be content-agnostic. 
Regardless of what the type of data we're read/writing is, we *pretend* it is a `uint64_t`. As long as we are consistent in how we package / unpackage the data, and in what we are returning to the calling function, we can have the application deal with interpreting the data. 

Using a pointer to data is probably a better solution. I recommend you make that upgrade in a future commit. It will require a minor refactor of the rest of the code as well.

hibike/devices/ExampleDevice/example_device.cpp
yizow: Oops. Should be fixed.

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,116 @@
...
+//    len     -   number of bytes in data
+//
+//   return  -   size of bytes written on success; otherwise return 0
 
-// you must implement this function. It is called when the devie receives a DeviceStatus packet.
-// the return value is the value field of the DeviceRespond packet hibike will respond with
-uint32_t device_status(uint8_t param) {
-  if (param < NUM_PARAMS) {
-    return params[param];
+uint32_t device_write(uint8_t param, uint8_t* data, size_t len) {
```
karthik-shanmugam, line 44: Same issue as before; len is a buffer size and should not be the argument to `read_num_bytes()`

PR #88, opened 2017-01-01
[DAWN] fixed paragraph paddings for peripherals and gamepads

kev-chien:
The peripheral and gamepads panels had no padding for text, which was bothering me. I modified `static/dawn.css` so all `<p>` elements have 15px padding, which may be hacky considering future use of `<p>`.


kevinmasd:
Seems too hacky for me. p may or not be that prevalent. Adding a class tag may be better.

kevinmasd (changes requested):
Seems too hacky for me. p may or not be that prevalent. Adding a class tag may be better.

kevinmasd (approved):

Issue #38, opened 2016-09-25
Dawn sagas tests


AllanYangZhou:
All sagas are currently located in a single file, `dawn/renderer/utils/sagas.js` (subject to change if refactored).

May want to look at https://github.com/stoeffel/redux-saga-test.

kev-chien:
Closing for now because most saga tests have been done in #89

PR #89, opened 2017-01-04
[DAWN] sagas tests

kev-chien:
Written 6/7 (-ish) tests for sagas in `renderer/utils/sagas.js`. 
Addresses #38 .
PR #91 also changed sagas.js, so merge that first and then I will make changes to this PR.

### Notes
- Modified `sagas.js` a bit to allow tests to run
- Added the action "SAVE_FILE_SUCCEEDED" to `actions/editorActions.js` for consistency
- electron can't be tested (remote and ipcRenderer do not exist), I suppose it is because the test environment differs and those elements don't exist until rendered
- I didn't write a test for ansibleGamepads() because it references navigator.getGamepads() but navigator is not imported, and is only available in the Chrome engine when the app is actually running. To solve this we could try using [electron-mocha](https://github.com/jprichardson/electron-mocha) to actually test in a running instance of electron (and thus fix/complete other tests).


dawn/renderer/utils/sagas.js
```diff
@@ -81,19 +82,16 @@ function* openFile() {
...
   yield cps(fs.writeFile, filepath, code);
-  yield put({
-    type: 'SAVE_FILE_SUCCEEDED',
-    code,
-    filepath,
-  });
+  yield put(saveFileSucceeded(code, filepath));
 }
 
+const editorState = (state) => ({
```
kevinmasd: Probably going to get caught on the Linter. Otherwise looks fine for now; will need to test with the new changes though.

kevinmasd (changes requested):

dawn/renderer/utils/sagas.js
kev-chien: It passed the linter. I've added tests for the new changes.

kevinmasd: Meant post #101 . ESLint has new rules.

kev-chien: *.test.js is in `.eslintignore`. Should we remove this so tests are also linted?

kevinmasd (approved):

PR #90, opened 2017-01-07
Runtime/main runtime v2

yizow:
2 commits:

1st contains fixes for bugs I introduced in previous commits.

The 2nd adds new functionality. 
The main runtime() function now has 3 different states: one for teleop, autonomous, and idle. 
Currently only teleop and idle function, but adding autonomous code should be relatively simple.

### DIFF
--- a/runtime/testy/runtime.py
+++ b/runtime/testy/runtime.py
@@ -34,7 +34,7 @@ def runtime(testName=""):
   testMode = testName != ""
   maxIter = 3 if testMode else None
 
-  def nonTestMode_print(*args):
+  def nonTestModePrint(*args):
     """Prints only if we are NOT in testMode"""
     if not testMode:
       print(args)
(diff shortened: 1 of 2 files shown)


runtime/testy/runtime.py
```diff
@@ -34,43 +34,64 @@ def runtime(testName=""):
   testMode = testName != ""
   maxIter = 3 if testMode else None
 
+  def nonTestMode_print(*args):
```
nikitakit: This mix of camelCase and underscore_style is annoying me

runtime/testy/runtime.py
```diff
@@ -34,43 +34,64 @@ def runtime(testName=""):
...
-        if emergency_stopped:
-          print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
-          print("terminating due to E-Stop")
-          break
-        print(RUNTIME_CONFIG.DEBUG_DELIMITER_STRING.value)
-        print("Starting studentCode attempt: %s" % (restartCount,))
-      spawnProcess(PROCESS_NAMES.STUDENT_CODE, runStudentCode, testName, maxIter)
+      if testMode:
+        # Automatically enter telop mode when running tests
+        badThingsQueue.put(BadThing(sys.exc_info(),
```
nikitakit, line 58: I'm curious if the name `BadThing` is still accurate after it's been overloaded at least for mode-switching.

nikitakit (commented):
Code LGTM other than minor style things.

However, your handling of the different game modes here needs to actually match the student API you're going to expose (i.e. there exist student API decisions where the current implementation is suboptimal). I don't remember what student API you guys ended up deciding on, so I'll take a wait-and-see approach and comment more on this once the student API code is put up for review.

runtime/testy/runtime.py
yizow: I also dont think its entirely accurate. But, changing it now would involve quite a few changes across a number of files, and I wanted to postpone that refactor change for later.

runtime/testy/runtime.py

nikitakit (approved):
lgtm

Issue #32, opened 2016-09-25
Centralize dawn settings state, figure out persistent solution


AllanYangZhou:
Currently different settings are handled in different ways, some persistent (saved even after you close Dawn), others are not. We should centralize this all into one reducer (the `settings` reducer) and save that so no settings are lost after closing Dawn.

kev-chien:
Closed with #91

Issue #31, opened 2016-09-25
Add UI for dawn editor settings


AllanYangZhou:
Currently we have actions and reducers for changing editor settings but no UI for the actual user to change those settings. We should make one.

kev-chien:
Closed with #91

Issue #26, opened 2016-09-24
Duplicates settings actions in Editor Actions


kev-chien:
[In `dawn/renderer/actions/EditorActions`, the functions:

`{ changeTheme, increaseFontsize, decreaseFontsize }`

are duplicates in `dawn/renderer/actions/SettingsActions`.

As a result, my tests in branch `dawn/actiontests` may have to be changed as well to remove duplicates.

kev-chien:
Closed with #91

PR #91, opened 2017-01-07
[DAWN] editor ui, persistent settings, re-rendering on resize

kev-chien:
Addresses several issues (while heavily reusing older dawn's code):

- #31 **New UI buttons for editor**: API, fontsize changing, and changing theme.
    - increase font and decrease font are now throttled through buttons becoming disabled
- #32 and #26 : **settings for theme and fontsize are now persistent**/stored on user's computer.
    - modified the actions increaseFontsize and decreaseFontsize into one action changeFontsize which takes a parameter `newFontsize`
    - font and theme actions are now exclusively in settings
- **Editor size re-renders on resizing of app window**
- **dialogs** on exit/open/new file without saving
- tooltips for buttons


kev-chien:
And since I had been trying to package dawn as an app to test, I had modified `webpack.production.config.js` to "work" in that the commands described in README to package dawn can run, although I'm not sure if it works entirely since my terminal stopped due to wine not being installed.

dawn/renderer/reducers/editor.js
```diff
@@ -30,12 +30,7 @@ const editor = (state = defaultEditorState, action) => {
         latestSaveCode: action.code,
       };
     case 'CREATE_NEW_FILE':
-      return {
-        ...state,
-        filepath: null,
-        editorCode: '',
-        latestSaveCode: '',
-      };
+      return state;
```
kevinmasd: Why does the state not change in this case?

kev-chien: I changed it so that `sagas.js` intercepts the 'CREATE_NEW_FILE' action to call the openFile saga. This was so a dialog would appear if there are unsaved changes when the user tries to create a new file. In the openFile saga, after success, the action "OPEN_FILE_SUCCEEDED" is dispatched. Thus, I changed the flow so that from 'CREATE_NEW_FILE', after success, 'OPEN_FILE_SUCCEEDED' would be dispatched in place with a null filepath, empty editorCode, and empty latestSaveCode.
That being said, I'll remove this code segment as an unchanged state is not the state we necessarily desire, and this was redundant anyways for returning same state.

kevinmasd: Sounds good.

kevinmasd (approved):

PR #92, opened 2017-01-08
Runtime/renaming sensors

yizow:
Contains commits from another pull request: merge after that one is in. 

Found an interesting bug: If there is no internet connection, Ansible errors out.


runtime/testy/Ansible.py
```diff
@@ -142,7 +142,7 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
                         s.sendto(msg, (host, UDPSendClass.SEND_PORT))
                     nextCall += 1.0/self.socketHZ
                     time.sleep(max(nextCall - time.time(), 0))
-                except Exception:
+                except Exception as e:
```
nikitakit, line 145: Wow, this is a bugfix that's just subtly hidden here among the other changes.

runtime/testy/namedSensors.csv
```diff
@@ -0,0 +1,3 @@
+two,2
```
nikitakit: FYI last year we decided that peripheral names must match the regex `^[A-Za-z][A-Za-z0-9]*$` (i.e. alphanumeric, but can't start with a number). Hopefully the code for this checking is still a part of Dawn, and runtime doesn't have to do anything about this. (I'm bringing this up because serialization schemes may break if you allow arbitrary character sets.)

nikitakit: On a separate note:
* Maybe use the more general term "peripherals" instead of "sensors"
* Why is this file in version control? It will eventually be user-writable (though doesn't seem that way yet). This means that pretty soon you'll need to figure out a cleaner scheme for switching between a real file and a dummy file for testing.
* What happens in the case of one hibike board with multiple attached sensors (i.e. two limit switches)? Last year's sensor mapping scheme allowed assigning a unique name to each switch, meaning that this created two entries in the mappings table.

runtime/testy/runtime.py
```diff
@@ -200,16 +221,19 @@ def runtimeTest(testNames):
   for testName in testNames:
     testFileName = "%s_output" % (testName,)
     with open(testFileName, "w", buffering = 1) as testOutput:
+      print("Running test: {}".format(testName), end="")
+      sys.stdout.flush()
       sys.stdout = testOutput
 
       allProcesses.clear()
-
```
nikitakit, line 229: gratuitous whitespace change.

runtime/testy/runtime.py
```diff
@@ -200,16 +221,19 @@ def runtimeTest(testNames):
   for testName in testNames:
     testFileName = "%s_output" % (testName,)
     with open(testFileName, "w", buffering = 1) as testOutput:
+      print("Running test: {}".format(testName), end="")
```
nikitakit: Why `end=""` here?

runtime/testy/runtime.py
```diff
@@ -200,16 +221,19 @@ def runtimeTest(testNames):
...
+      sys.stdout.flush()
```
nikitakit: FYI this may be equivalent to doing `print("...", flush=True)` on the line above.

runtime/testy/Ansible.py
yizow: Whats the proper procedure here? Should i mention it in the commit/ pull request message?

runtime/testy/runtime.py
yizow: python csv module terminates lines with '/r/n' be default.
It looked ugly in git diff

runtime/testy/runtime.py
yizow: Fixed

runtime/testy/runtime.py

runtime/testy/namedSensors.csv
yizow: Note made to have dawn sanitize names.

This file is currently being used for tests, so it is in the version control. When/ if we want to have a real/dummy file, a simple rename will likely be sufficient. 

Note made to address addressing multiple sensors on same board.

nikitakit:
Still think rename namedSensors to use a term like "devices" or "peripherals" instead, since it will include motors and other actuators.

yizow:
Fixed and rebased

nikitakit (approved):
lgtm

PR #93, opened 2017-01-09
Runtime/autonomous

yizow:
Depends on other pull requests.

### DIFF
--- a/runtime/testy/runtime.py
+++ b/runtime/testy/runtime.py
@@ -137,6 +137,7 @@ def runStudentCode(badThingsQueue, stateQueue, pipe, testName = "", maxIter = No
     # TODO: Replace execCount with a value in stateManager
     execCount = 0
     while (not terminated) and (maxIter is None or execCount < maxIter):
+      studentCode.Robot.handleSchedule()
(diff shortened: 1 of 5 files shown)


nikitakit (changes requested):
So what I'm seeing here is the beginnings of a custom scheduler implementation. That raises two questions:

1. Why do we need a scheduler?
2. Why are we writing one ourselves?

With respect to point 1, I would like to see a sketch of the intended student API before implementation begins. Important questions include: what is being enqueued (hibike commands, or arbitrary functions including `lambda` expressions)? What happens when something is enqueued from inside `loop` (i.e. at a rate of 30Hz)? For these and other questions, see our slack channel.

Once we have an idea of what API we want to provide, we can then address point 2 -- i.e. determine what parts we need to implement ourselves, and what parts can reuse existing components. For example, python 3.5 has a built-in `asyncio` module (plus `async`/`await` keywords), which seems like a more promising approach then writing our own event loop/scheduler.

In summary, I find it very likely that the approach in this PR will have to be significantly modified after additional discussion and prototyping take place. Therefore I don't think it makes sense to merge it at this stage.

PR #97, opened 2017-01-16
Add preliminary networking spec.

krystynneisess:
As requested by @yizow

### DIFF
--- a/networking/Field_Control_Networking_Spec.md
+++ b/networking/Field_Control_Networking_Spec.md
@@ -12,35 +12,76 @@
 
 A robot interacts with two networks:  the Student network and the Field Control (FC) network.  The former is the network over which students communicate with their robot via provided mini-routers.  The latter--the focus of this spec--is the network over which students' robots interact with FC over the course of a match.
 
-Clients (field control, driver stations, and robots) are assigned static IPs and communicate over TCP connections.
+Clients (field control, driver stations, and robots) are assigned static IPs and perform field-related communications over TCP connections.
 
-FC sends out messages to initiate robot-to-driver-station pairing and robot mode changes.  FC listens to heartbeat messages from robots.
+Robots announce their presence upon joining the FC network.  FC sends out messages to initiate robot-to-driver-station pairing and robot mode changes.  FC listens to heartbeat messages from driver stations.
 
 ## 2. Setting Up the Clients and Network
 
-### Clients
+### Client Types
 
-The FC network will have up to eight clients, as follows:
+The FC network will have at least six clients, as follows:
 
-- One FC client.  This is the FC computer running the match.  A PiE computer+monitor. 
-- Four driver station clients.  This acts as the student-robot interface (via Dawn and Runtime), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
-- Up to four robot clients.  A wifi-enabled Beagle Bone Black (via wifi dongle).  One per student robot.
+- One FC client.  This is the FC computer running the match.  An internet-enabled PiE computer+monitor. 
+- One Streaming client.  This is the computer recording/processing/displaying a livestream of the field.  An internet-enabled PiE computer+monitor.
+- Four driver station clients.  This acts as the student-robot interface (via Dawn), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
+- Up to four robot clients.  A wifi-enabled Beagle Bone Black (BBB), via wifi dongle.  One per student robot.
+
+Ideally there should be no additional clients; however, in the event that such should be needed, a few IP addresses have been reserved for them. (See "Assigning IP Addresses.")
 
 ### Assigning IP Addresses
 
-Every client will have a statically assigned IP address.  This is feasible due to the small number of total clients (1 per robot ( less than 30 in total), 4 driver stations, and one field control center).
+The Netmask is arbitrarily decided to be 245.245.245.0.  This implies that the subnet prefix, denoted by P, is "245.245.245".
 
-Consider the following example static IP scheme ("P" = subnet prefix):
+Every client will have a statically assigned IP address.  This is feasible due to the small number of total clients.  Assignment/reservations are as follows:
 
-- FCl:  P.255
-- Driver Stations:  P.254-251
-- Robots:  P.TeamNumber
+- P.255:  Broadcast
+- P.254:  FC center computer
+- P.253-250:  Driver station computers
+- P.249:  Streaming computer
+- P.248-246:  Spare FC clients
+- P.245:  Network gateway (i.e. router)
+- P.244-200:  Staff internal robots
+- P.199-51:  Unreserved
+- P.50-1: Student robots
+- P.0:  Netmask
+
+### Software Configurations
+
+#### Network
+
+- **Netmask.**  As mentioned above, this is arbitrarily decided to be P.0.
+- **Gateway IP.** This is arbitrarily decided to be P.245.
+
+#### Robot Clients
+
+Each robot is provided a config file mapping a unique identifier (e.g. BBB's MAC address) to a collection of robot-specific settings:
+
+- **Team Number.**  If less than 50, a student robot; if greater than 50, a staff internal robot.
+- **IP Address.**  This is derived from team number, as P.TeamNumber.
+
+A robot is also provided a script that takes the robot's unique identifier and this config file, and generates a customized network config file containing the above team-specific settings, as well as static network configurations (netmask, gateway IP).
+
+Qualities and Caveats:
+
+- This config management scheme is desirable due to the ease of updating robots en masse without breaking functionality.  Should any robots need to be reconfigured for any reason, one only need update the config file entries for the affected robot(s). 
+- Note, however, that this scheme requires that prior to deployment, _every BBB, student and staff alike, must first be registered_ to some centralized system in order for the config file to be crafted and maintained.  ("Registering" constitutes pairing a MAC addres with a Team Number.)  Known dysfunctional BBBs should also be unregistered, so as to avoid potential config conflicts if accidentally redistributed.
+
+#### FC, Driver Station, and Streaming Clients
+
+**IP Address.** Assigned as described above.
+
+### Internet Access
+
+Both the FC client and Streaming client require internet access; the former to access match scheduling as listed online, and the latter to publish video feed to an online streaming platform (e.g. Twitch).
+
+This is achieved through dual network connections (to both the public internet and FC network), requiring either dual Ethernet ports or an external switch.
 
 ## 3. Connecting to the Field Control Network
 
-### Configuration
+### Prerequisites
 
-A robot can connect to at most one network at a time, which must be either the Student network or the FC network.  The Student network is to be listed as higher priority than the FC Network.  This can be set each robot's network config file.
+A robot can connect to at most one network at a time, which must be either the Student network or the FC network.  The Student network is to be listed as higher priority than the FC Network.  This can be set in each robot's network config file.
 
 In order to participate in a match, teams temporarily relinquish their routers.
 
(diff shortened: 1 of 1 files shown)


networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,79 @@
...
+
+FC sends out messages to initiate robot-to-driver-station pairing and robot mode changes.  FC listens to heartbeat messages from robots.
+
+## 2. Setting Up the Clients and Network
+
+### Clients
+
+The FC network will have up to eight clients, as follows:
+
+- One FC client.  This is the FC computer running the match.  A wifi-enabled PiE computer+monitor.
```
akumaraguru: There's no reason for the FC client or driver stations to be on wireless connections. Probably want to remove "wifi enabled" from the spec so we don't give the wrong idea.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Clients
+
+The FC network will have up to eight clients, as follows:
+
+- One FC client.  This is the FC computer running the match.  A wifi-enabled PiE computer+monitor. 
+- Four driver station clients.  This acts as the student-robot interface (via Dawn and Runtime), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
+- Up to four robot clients.  A wifi-enabled Beagle Bone Black (via wifi dongle).  One per student robot.
+
+### Assigning IP Addresses
```
akumaraguru, line 32: How is the network supposed to handle IP address conflicts? Poor management of Beaglebones may cause this.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,79 @@
...
+2. When in a match, robots will only have the option of connecting to the FC network (due to router confiscation). 
+
+Note that this may result in robots not in a match to still be connected to the FC network.  However, this is totally fine since the FC client will not be listening to robots not currently tied to a driver station (see "Sending Messages").
+
+## 4. Establishing Communication Links
+
+Within each match, the FC network will create/maintain eight TCP connections:
+
+- A connection between the FC client and each Driver Station client.  These can persist across matches, and so need to be opened only once per network initialization.
+- A connection beween a Driver Station client and its corresponding Robot client. These must be opened/closed at the start/end of each match (i.e. once the Robot client joins/leaves the FC network and is paired/unpaired with a Driver Station client).
```
akumaraguru, line 105: Might want to specify who (FC client, Dawn, ???) holds responsibility for when a TCP connection is broken.

akumaraguru (changes requested):
This is good for giving general software folk an overview of how the network is designed. Idk if it's meant for anything else, but if so, it would be good to have more deals on implementation.

krystynneisess:
Comments addressed!

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+Clients (field control, driver stations, and robots) are assigned static IPs and communicate over TCP connections.
+
+FC sends out messages to initiate robot-to-driver-station pairing and robot mode changes.  FC listens to heartbeat messages from robots.
+
+## 2. Setting Up the Clients and Network
+
+### Clients
+
+The FC network will have up to eight clients, as follows:
```
yizow: Nine total

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+- Four driver station clients.  This acts as the student-robot interface (via Dawn and Runtime), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
+- Up to four robot clients.  A wifi-enabled Beagle Bone Black (via wifi dongle).  One per student robot.
+
+### Assigning IP Addresses
+
+Every client will have a statically assigned IP address.  This is feasible due to the small number of total clients (1 per robot ( less than 30 in total), 4 driver stations, and one field control center).
+
+Consider the following example static IP scheme ("P" = subnet prefix):
+
+- FCl:  P.255
```
yizow: Not a valid IP. This is the reserved broadcast IP

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+The FC network will have up to eight clients, as follows:
+
+- One FC client.  This is the FC computer running the match.  A PiE computer+monitor. 
+- Four driver station clients.  This acts as the student-robot interface (via Dawn and Runtime), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
+- Up to four robot clients.  A wifi-enabled Beagle Bone Black (via wifi dongle).  One per student robot.
+
+### Assigning IP Addresses
+
+Every client will have a statically assigned IP address.  This is feasible due to the small number of total clients (1 per robot ( less than 30 in total), 4 driver stations, and one field control center).
```
yizow: Add in specifics of IP. You can do this after taking 168.

Subnet, etc. Everything needed to configure the meraki router.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+- FCl:  P.255
+- Driver Stations:  P.254-251
+- Robots:  P.TeamNumber
+
+## 3. Connecting to the Field Control Network
+
+### Configuration
+
+A robot can connect to at most one network at a time, which must be either the Student network or the FC network.  The Student network is to be listed as higher priority than the FC Network.  This can be set each robot's network config file.
```
yizow: This can be set *in* each

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+In order to participate in a match, teams temporarily relinquish their routers.
+
+### Rationale 
+
+This results in two desirable guarantees:
+
+1. When not in a match, students will always be able to connect with their robots over the Student network even if the robot is within range of the FC network.
+2. When in a match, robots will only have the option of connecting to the FC network (due to router confiscation). 
+
+Note that this may result in robots not in a match to still be connected to the FC network.  However, this is totally fine since the FC client will not be listening to robots not currently tied to a driver station (see "Sending Messages").
```
yizow: Add how students will fix. 
namely, restarting robot.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+- A connection between the FC client and each Driver Station client.  These can persist across matches, and so need to be opened only once per network initialization.
+- A connection beween a Driver Station client and its corresponding Robot client. These must be opened/closed at the start/end of each match (i.e. once the Robot client joins/leaves the FC network and is paired/unpaired with a Driver Station client).
+
+If a FC-to-Driver-Station connection is dropped, FC is responsible for reestablishing the connection.  If a Driver-Station-to-Robot connection is dropped, the Driver Station notifies FC; FC then re-sends the pairing message to said Driver Station (see "Sending Messages").
+
+## 5. Sending Messages
+
+The FC client sends messages to initiate robot-to-driver-station pairing, and to indicate what mode (autonomous, tele-op, etc) the robot is in.   Driver station clients listen accordingly, and pass the message to robots.
+
+Robot clients send only a heartbeat, which contains a status value that the FC client can use to verify the robot is both connected an in the expected mode.
```
yizow: Driver station sends heartbeat. 

Robot heartbeats are a runtime/dawn protocol, and outside the scope of this document.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+In further detail:
+
+### From FC
+
+**Robot-to-Driver-Station Pairing.**  Prior to this message, a robot, while connected to the network, is not yet associated with a driver station.  Containing the static IP of the Robot client with which to pair, the FC client sends this to the Driver Station client, which then opens a connection and begins communication with said robot.
```
yizow: We probably want to send just the robot number. 
The IP can be determined from the number.
In the future, we can use a lookup table if we don't want to use static IP.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+If more than one robot responds to the pairing message (i.e. IP address collision), or if no robot responds, FC is notified by the Driver Station.  The match will not proceed until this is externally resolved (e.g. reprogramming the suspected robot(s)'s Beaglebones).
```
yizow: Respond with what message? Different for each? Same for both error cases? 

Alternative: no special message gets sent, but FC station sees from heartbeat that robot remains disconnected.

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Mode Changing.**  This message contains an integer indicating whether the robot should be in Autonomous (1), Tele-Op (2), Disabled (3), or E-Stopped (4) mode.  Sent from FC to Driver Station, then Driver Station to Robot.
```
yizow: - Recommend zero indexing
- Recommend setting e-stop as 0

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+
+### To FC
+
+**Heartbeat.** Upon pairing, the robot begins sending a "heartbeat"--a message containing a single integer between 0 and 4 (inclusive) indicating its mode. If no mode change message has been received yet, it replies Connected (0) by default. Sent from Robot to Driver Station, then Driver Station to FC.
```
yizow: outside scope.

yizow (changes requested):

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+FC sends out messages to initiate robot-to-driver-station pairing and robot mode changes.  FC listens to heartbeat messages from robots.
+
+## 2. Setting Up the Clients and Network
+
+### Clients
+
+The FC network will have up to eight clients, as follows:
+
+- One FC client.  This is the FC computer running the match.  A PiE computer+monitor. 
+- Four driver station clients.  This acts as the student-robot interface (via Dawn and Runtime), and as the robot-field control interface (see "Establishing Communication Links").  A wifi-enabled laptop or  computer+monitor.
```
nikitakit: I don't think you should mention runtime in this bullet point.

networking/Field_Control_Networking_Spec.md
nikitakit: FYI if there is any need to update the network config file as part of runtime updates, I suggest using the lookup table trick I proposed in an earlier meeting.

* All robots receive a config file mapping from a unique ID (e.g. mac address), to any robot-specific settings like the IP
* All robots receive a script that reads the config file and the unique ID to generate the config file and put it in the right place

This means that if we need to update the network config files for any reason, we can easily deploy an update to all robots without breaking their functionality.

(On that note, if we go with this scheme -- this spec should mention the IPs that are reserved for staff-internal beaglebones)

networking/Field_Control_Networking_Spec.md
nikitakit: So I believe Dawn<->Runtime connection is primarily UDP. (though if you need the TCP endpoint as well on the field, the spec should accommodate both)

And yes, you need to spell out the behavior in case:
* The robot restarts, thus resetting the connection
* The driver station is unplugged from the ethernet cable, thus resetting the connection (yes, this has actually happened on the competition field)
* The connection enters an invalid state, thus necessitating a reset.

networking/Field_Control_Networking_Spec.md
nikitakit: So the static IP can be deterministically generated from the match schedule. But should FC also do some checking first, e.g. pinging the robot?

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+- Up to four robot clients.  A wifi-enabled Beagle Bone Black (via wifi dongle).  One per student robot.
```
nikitakit: If you want to do video streaming, this may add an extra computer on the network. This adds some interesting constraints:
* Video streaming is hardware-demanding (running other things on same computer may not be a good idea)
* Video streaming needs a human to operate occasionally, so it's nice to be able to have one operator manage the FC base station (i.e. input match numbers), and a second person to oversee streaming.
* Video streaming needs access to the internet! (should be obvious why)
* Video streaming needs access to the field control internal network, so that it can get the current match timer in order to display it up on the screen.

The need for both Internet and internal networks complicates the setup somewhat.

nikitakit: In addition to video streaming, the match schedule used to be pulled from an online source (e.g. a Google Spreadsheet), such that both event planners and field control operate off the same source of ground truth. I think one year we passed around schedules on paper slips and USB drives... that was a mess.

This also adds the need for a computer that has access to both field control and the internet.

nikitakit (dismissed):
Read my comments first.

In addition, the following communication modes we previously had in field control are absent from this spec:
* Sending the current match time from FC to Dawn (so students know how much time is left)
* Sending the current match time and teams involved such that they can be displayed in the video on the overhead projector (which may also be streamed)
* Note that e-stop should be triggerable from anywhere: from the robot (e.g. via a determination of unsafe battery buzzer voltage), from Dawn (via the student hitting e-stop), or from the field control operator (i.e. PiE field staff decide the robot is operating in an unsafe manner). Technically the from-robot functionality is not a networking thing because the robot can always decide for itself to permanently stay in e-stop mode.

networking/Field_Control_Networking_Spec.md
nikitakit: An example of one reason we'd need to deploy a custom config: say we need to update the settings for field control connections due to last-minute discovering issues with our field router config.

networking/Field_Control_Networking_Spec.md
Alramech: Since the robot connects to the student network and the field network over the same interface, specify whether they use the same IP for both networks or have independent ones

networking/Field_Control_Networking_Spec.md
Alramech: There also should be a specification of how FC knows which driver station gets paired with which robot. Possibilities include reading from the spreadsheet, manual input, or some other method of communication

networking/Field_Control_Networking_Spec.md
```diff
@@ -0,0 +1,83 @@
...
+Within each match, the FC network will create/maintain eight TCP connections:
+
+- A connection between the FC client and each Driver Station client.  These can persist across matches, and so need to be opened only once per network initialization.
+- A connection beween a Driver Station client and its corresponding Robot client. These must be opened/closed at the start/end of each match (i.e. once the Robot client joins/leaves the FC network and is paired/unpaired with a Driver Station client).
+
+If a FC-to-Driver-Station connection is dropped, FC is responsible for reestablishing the connection.  If a Driver-Station-to-Robot connection is dropped, the Driver Station notifies FC; FC then re-sends the pairing message to said Driver Station (see "Sending Messages").
+
+## 5. Sending Messages
+
+The FC client sends messages to initiate robot-to-driver-station pairing, and to indicate what mode (autonomous, tele-op, etc) the robot is in.   Driver station clients listen accordingly, and pass the message to robots.
```
Alramech, line 117: robot should initiate pairing via message to FC as it is generally unclear when it is ready to communicate unless inspecting every packet

networking/Field_Control_Networking_Spec.md
Alramech: How will detecting more than one robot be detected? What if 2 robots respond, but 1 is much later than the other?

nikitakit:
Umm, what's the state of this? It's been a while since the last comment or spec change.

akumaraguru (approved):

yizow (approved):

PR #98, opened 2017-01-16
Networking/chris/devel

Alramech:
ignore client/server/driver files

### DIFF
--- a/networking/config.py
+++ b/networking/config.py
@@ -1,0 +1,7 @@
+from enum import Enum
+
+class Commands(Enum):
+  ESTOP = "0"
+  START_GAME = "1"
+  TELEOP = "2"
+  END_GAME = "3"
--- a/networking/robot.py
+++ b/networking/robot.py
@@ -1,0 +1,64 @@
+import lcm
+import threading
+import time
+from config import Commands as cmd
+
+class Robot():
+  """
+  Basic robot simulator. Features
+  game states as well as estop and basic i/o
+  """
+  def __init__(self, number):
+    self.number = number
+    self.teleop = False
+    self.motors = [0]*4
+    self.sensors = [True, 4]
+    self.enabled = False
+    self.estop = False
+        
+def heartbeat(lc, number):
+  """
+  Generates heartbeats for the robot
+  sends robot number as well as state flag
+  (0 = disabled, 1 = autonomous, 2 = teleop, 3 = estop)
+  """
+  i = 0
+  while True:
+    i += 1
+    state = robot.enabled + robot.teleop + robot.estop*2
+    status = "{} {}".format(robot.number, state)
+    lc.publish("ROBOT_HEARTBEAT", status)
+    time.sleep(5)
+        
+def cmd_handler(channel, data):
+  """
+  Handles recieved commands
+  """
+  data = cmd(data)
+  if data == cmd.ESTOP:
+    robot.enabled = False
+    robot.estop = True
+
+  if data == cmd.START_GAME:
+    robot.enabled = True
+
+  if data == cmd.TELEOP:
+    robot.teleop = True
+    
+  if data == cmd.END_GAME:
+    robot.enabled = False
+    
+lc = lcm.LCM()
+subscription = lc.subscribe("COMMANDS", cmd_handler)
+
+robot = Robot(1)
+
+t = threading.Thread(target = heartbeat, args = (lc, robot))
+t.daemon=True
+t.start()
+
+try:
+  while True:
+      lc.handle()
+except KeyboardInterrupt:
+    pass
--- a/networking/tentativename.py
+++ b/networking/tentativename.py
@@ -1,0 +1,45 @@
+import lcm
+import threading
+import time
+from config import Commands as cmd
+
+def heartbeat_handler(channel, data):
+  """
+  Prints out heartbeat data
+  """
+  data = data.decode()
+  msg = data.split(' ')
+  print("Robot {}: state {}".format(msg[0], msg[1]))
+  
+lc = lcm.LCM()
+subscription = lc.subscribe("ROBOT_HEARTBEAT", heartbeat_handler)
+
+def teleop():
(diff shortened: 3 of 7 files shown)


yizow:
See comments on slack.

networking/config.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+class Commands(Enum):
+  ESTOP = "0"
+  START_GAME = "1"
+  TELEOP = "2"
+  END_GAME = "3"
```
yizow, line 7: newline at end of file

networking/tentativename.py
```diff
@@ -0,0 +1,45 @@
+import lcm
```
yizow, line 1: Change the file name and choose a casing

networking/tentativename.py
```diff
@@ -0,0 +1,45 @@
...
+def heartbeat_listen(lc, robot_number):
+  """
+  Subthread intended to listen for heartbeats
+  """
+  while True:
+    lc.handle()
+    
+t = threading.Thread(target = heartbeat_listen, args = (lc, 1))
+t.daemon=True
+t.start()
```
yizow, line 45: newline at end of file

networking/test.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+import time
+import filecmp
+import sys
+
+infilename = "test.py"
+
+outfile = open("testout.txt", "w")
```
yizow, line 8: I recommend adding a `.gitignore` file in this directory to apply to various temp files youre going to be generating. such as lcm stuff.

networking/robot.py
```diff
@@ -0,0 +1,64 @@
+import lcm
+import threading
+import time
+from config import Commands as cmd
```
yizow, line 4: cmd is a python default module. 
I recommend not overwriting it.

If you maintain config to only be global constants and enum stuff, etc, I would recommend 
`from config import *`

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Robot():
+  """
+  Basic robot simulator. Features
+  game states as well as estop and basic i/o
+  """
+  def __init__(self, number):
+    self.number = number
+    self.teleop = False
```
yizow, line 13: `self.teleop` and `self.enabled` seem redundant.

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+    self.motors = [0]*4
+    self.sensors = [True, 4]
+    self.enabled = False
+    self.estop = False
+        
+def heartbeat(lc, number):
+  """
+  Generates heartbeats for the robot
+  sends robot number as well as state flag
+  (0 = disabled, 1 = autonomous, 2 = teleop, 3 = estop)
```
yizow, line 23: i recommend having estop as 0, sinces it's more commonly used as an error state

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+  """
+  i = 0
+  while True:
+    i += 1
+    state = robot.enabled + robot.teleop + robot.estop*2
```
yizow, line 28: what the heck is this supposed to mean?

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+  if data == cmd.START_GAME:
+    robot.enabled = True
+
+  if data == cmd.TELEOP:
+    robot.teleop = True
+    
+  if data == cmd.END_GAME:
+    robot.enabled = False
+    
+lc = lcm.LCM()
```
yizow, line 51: probably want to choose a different name than `lc`

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+def cmd_handler(channel, data):
+  """
+  Handles recieved commands
+  """
+  data = cmd(data)
+  if data == cmd.ESTOP:
+    robot.enabled = False
+    robot.estop = True
+
+  if data == cmd.START_GAME:
```
yizow, line 42: Consider using `elif` for these just to catch bug cases where you alias enums together.

yizow: Another alternative is to use a dictionary mapping from enum -> function definitions.

networking/tentativename.py
```diff
@@ -0,0 +1,45 @@
+import lcm
+import threading
+import time
+from config import Commands as cmd
```
yizow, line 4: See above comments

networking/tentativename.py
```diff
@@ -0,0 +1,45 @@
...
+
+def heartbeat_handler(channel, data):
+  """
+  Prints out heartbeat data
+  """
+  data = data.decode()
+  msg = data.split(' ')
+  print("Robot {}: state {}".format(msg[0], msg[1]))
+  
+lc = lcm.LCM()
```
yizow, line 14: Should probably move this to outside of function definitions

yizow (changes requested):
PLS CHANGE FIXES PL0X

yizow (changes requested):
Gitignore.

networking/robot.py
```diff
@@ -0,0 +1,64 @@
...
+  Generates heartbeats for the robot
+  sends robot number as well as state flag
+  (0 = disabled, 1 = autonomous, 2 = teleop, 3 = estop)
+  """
+  i = 0
+  while True:
+    i += 1
+    state = robot.enabled + robot.teleop + robot.estop*2
+    status = "{} {}".format(robot.number, state)
+    lc.publish("ROBOT_HEARTBEAT", status)
```
nikitakit, line 30: I notice you're publishing strings on LCM.

I strongly recommend that you switch to publishing using LCM types instead (or another strict encoding system like protobuf) -- using types helps you keep your sanity when building a distributed system.

nikitakit (changes requested):

networking/robot.py
nikitakit: @yizow So `lcm` is already taken (module name), and you want something short...

PR #99, opened 2017-01-20
[DevOps] Beaglebone setup updates

jonathan-j-lee:
* Added wireless dongle driver installation
    * Fixed driver not loading on boot (added `mt7610u_sta` to `/etc/modules`)
* Updated Python dependencies
    * Added `protobuf` runtime
        * `protobuf` compilation will happen when updates are packaged
    * Added `pyusb` to the list of Python packages with releases
* Fixed resolver configuration file contents to persist between reboots
* Added script for wifi over USB (for development)
* Fixed bad SSH on boot
* Configured date-time to be PST
* Added System-V-style runtime initialization
* Disabled the BBB's default Apache web server
* Removed grizzly stuff and GPG keys
* Restored `memcached_backup.conf` as `memcached.conf`
* Added `mac.py` (MAC address script) installation to `bin`
* Updated many paths


ArjunSridhar1 (approved):
Looks good, nice work!
At this point, runtime should be starting up correctly so we need to begin testing and fix networking.

Issue #96, opened 2017-01-14
Hosted student API documentation


nikitakit:
We previously had student API documentation autogenerated from docstrings using sphinx, and hosted [on readthedocs](http://pie-api.readthedocs.io/). It would be nice to do something similar this year.

I am one of several admins for our readthedocs account/subdomain, and can help if you want to reuse that website.

kevinmasd:
We could also package the documentation into Dawn; Dawn already tries to get documentation (though it didn't work last time due to networking).

jiangdaniel:
If we put the [API on discourse](https://discourse.pierobotics.org/t/api-spec/281), do we still have to host it on readthedocs too?

kevinji:
readthedocs (arguably) has a better interface for searching through the docs, so I'd personally say it's useful.

jiangdaniel:
Our API is pretty short this year though, so it might not be that big of a deal?  And it's a little bit easier for us to maintain a single copy of the API rather than several copies.

nikitakit:
We put an authoritative copy of the API docs up on our website, so I'm now closing this issue.

Issue #35, opened 2016-09-25
Dawn Debug Mode


AllanYangZhou:
This is unnecessary for productions, it is only used in development. See this for ideas: https://github.com/gaearon/redux-devtools/blob/master/docs/Walkthrough.md#exclude-devtools-from-production-builds. We already have a `process.env.NODE_ENV`, so you can check if that variable is equal to `'development`'.

kevinmasd:
Going to expand on this. Various features (Fake Runtime, SFTP logging) seem to only be useful for Devs. Perhaps logging them onto a file or keeping debug flags may be useful.

PR #101, opened 2017-01-30
[Dawn] SFTP Uploads, Code Statuses, De-Zombied Fake Runtime, Periodic Messaging, IP Config, npm Upgrades

kevinmasd:
Student Code and Runtime Updates (minus signatures) have been implemented through a SFTP module. (Can be extended to run programs on robot if needed). Additionally, Ansible.deprecated has been removed.
- Currently uses a timeout mechanism; may try callbacks later.
- UpdateBox does not actually reflect whether upload succeeded on Robot side; it currently only detects errors on Dawn side.

Student Code statuses have been incorporated into the code, indicating when code is running or not and how to react.
- Currently binary: either running or idle. Will expand to errors and other feedback messages.

Bug involving Dawn not releasing ports on quit has been tracked down to Fake Runtime not being killed.
- Bugfix is ugly. Will try to find alternatives.

Dawn will now start sending messages to Runtime periodically, rather than every 50ms after detecting gamepad changes. This helps to test features without gamepads present.
- Currently sends messages in 1 second intervals; may change if needed.

ConfigBox (IP Address Changes) has been created.
- May expand to debug information and other configuration. May also find way to make IP slightly more permanent (Deletes on exit).

Package Updates and Node v6
- Packages like react-sagas and protobuf.js have received major API changes; upgrading them is non-trivial
- See whether React-Joyride has been fixed, or whether it is still an Electron issue.


dawn/fake-runtime/FakeRuntime.js
```diff
@@ -20,6 +20,7 @@ const hostname = 'localhost';
 const client = dgram.createSocket('udp4');// sender
 const server = dgram.createSocket('udp4'); // receiver
 const SENDRATE = 1000;
+let state = 2;
```
nikitakit: This looks like it should be some sort of enum

dawn/main/MenuTemplate/DebugMenu.js
```diff
@@ -2,8 +2,8 @@
  * Defines the debug menu.
  */
 
-import RendererBridge from '../RendererBridge';
 import { fork } from 'child_process';
+import RendererBridge from '../RendererBridge';
```
nikitakit, line 6: Not sure why you're switching the import order here.

I know there are stylistic changes in this PR as well as functional ones, and the fact that they're intermixed makes review harder.

dawn/package.json
```diff
@@ -28,54 +28,55 @@
...
-    "redux": "^3.5.2",
+    "react": "^15.4.2",
+    "react-ace": "^4.1.1",
+    "react-bootstrap": "^0.30.7",
+    "react-dom": "^15.4.2",
+    "react-immutable-proptypes": "^2.1.0",
+    "react-joyride": "^1.10.0",
+    "react-redux": "^5.0.2",
+    "redux": "^3.6.0",
+    "redux-devtools": "^3.3.2",
```
nikitakit: Note that `redux-devtools` is being moved from devDependencies to dependencies. I hope that this is intentional.

dawn/renderer/actions/AlertActions.js
```diff
@@ -12,7 +12,7 @@ export const addAsyncAlert = (heading, message) => ({
   message,
 });
 
-export const removeAsyncAlert = (id) => ({
+export const removeAsyncAlert = id => ({
```
nikitakit, line 15: Unfortunately the mix of linter changes and new functionality makes this PR difficult to review (especially for non-Dawn-developers, as you have assigned to this PR).

I'll do my best to tease out the changes, but I'll have to defer to someone on Dawn for additional review (maybe @AllanYangZhou ?)

dawn/renderer/components/ConfigBox.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+
+  getValidationState() {
+    const current = this.state.ipAddress;
+    let valid = false;
+    let regex = '^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.';
+    regex += '(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.';
+    regex += '(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.';
+    regex += '(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$';
```
nikitakit: There should be a way to construct this regex that doesn't involve concatenating four strings. Does javascript regex support a repeat-3-times operator? (To do something like `^((NUM)\.){3}(NUM)$`)

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,56 @@ class Editor extends React.Component {
...
+      console.log('No file? No upload');
+      return;
+    }
     const correctedText = this.correctText(this.props.editorCode);
     if (correctedText !== this.props.editorCode) {
       this.props.onAlertAdd(
         'Invalid characters detected',
         'Your code has non-ASCII characters, which won\'t work on the robot. ' +
-        'Please remove them and try again.'
+        'Please remove them and try again.',
```
nikitakit, line 171: Why comma at end of line here?

dawn/renderer/components/Editor.js
```diff
@@ -29,10 +27,27 @@ import 'brace/theme/solarized_dark';
...
 
 class Editor extends React.Component {
+  static pathToName(filepath) {
+    if (filepath !== null) {
+      if (process.platform === 'win32') {
+        return filepath.split('\\').pop();
+      }
+      return filepath.split('/').pop();
+    }
+    return '[ New File ]';
```
nikitakit: So `pathToName` appears to be used primarily to determine remote path names for scp uploads. In that case, '[ New File ]' doesn't look like a value you'd ever actually want to use, and throwing an exception seems more useful.

dawn/renderer/components/UpdateBox.js
```diff
@@ -4,9 +4,21 @@ import {
   Button,
 } from 'react-bootstrap';
 import { remote } from 'electron';
+
 const dialog = remote.dialog;
+const Client = require('ssh2').Client;
 
 class UpdateBox extends React.Component {
+  static pathToName(filepath) {
```
nikitakit: This is the second time I'm seeing `pathToName` defined... and the two implementations aren't even the same!

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,56 @@ class Editor extends React.Component {
...
-  }
-
-  upload() {
-    this.sendCode('upload');
+    const conn = new Client();
+    conn.on('ready', () => {
+      conn.sftp((err, sftp) => {
+        if (err) throw err;
+        console.log('SSH Connection');
+        sftp.fastPut(filepath, Editor.pathToName(filepath), (err2) => {
```
nikitakit: So I don't think `pathToName` is right here. As long as the student code remains a single file, the destination name will always be `student_code.py`.

Additionally, it would make sense to refactor everything related to remote paths on the BBB into a separate file, such that it's easier to make global changes if the beaglebone paths change.

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,56 @@ class Editor extends React.Component {
...
+          if (err2) throw err2;
```
nikitakit: Error handling (see above)

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,56 @@ class Editor extends React.Component {
...
-    console.log(`Debug Info: ${command} EOD`);
-    return true;
-  }
-
-  upload() {
-    this.sendCode('upload');
+    const conn = new Client();
+    conn.on('ready', () => {
+      conn.sftp((err, sftp) => {
+        if (err) throw err;
```
nikitakit: It would be good to show a modal if upload fails. I believe that this `throw` only targets the javascript debug console, and doesn't actually signal the students in any way.

(Typical errors for uploading include typing in the wrong IP, or being disconnected from the robot)

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,56 @@ class Editor extends React.Component {
...
-      }
-      return filepath.split('/').pop();
-    }
-    return '[ New File ]';
+  toggleConfigModal() {
+    this.setState({ showConfigModal: !this.state.showConfigModal });
+  }
+
+  openAPI() {
+    window.open('https://pie-api.readthedocs.org/');
```
nikitakit, line 237: Having internet-based docs looks good! (No changes requested)

But if we ever want local docs, it may make sense to have Dawn pull those from the robot instead of shipping them with Dawn. That way the docs always match the current runtime version (and also provide an easy way to check which version of runtime is on the robot).

dawn/renderer/components/UpdateBox.js
```diff
@@ -39,7 +51,26 @@ class UpdateBox extends React.Component {
   }
 
   upgradeSoftware() {
-    console.log('Deprecated');
+    this.setState({ isUploading: true });
+    const conn = new Client();
+    conn.on('ready', () => {
+      conn.sftp((err, sftp) => {
+        if (err) throw err;
```
nikitakit: See error handling comments for this function, as well.

dawn/renderer/utils/sagas.js
```diff
@@ -248,12 +253,13 @@ function* ansibleGamepads() {
     // navigator.getGamepads always returns a reference to the same object. This
     // confuses redux, so we use assignIn to clone to a new object each time.
     const newGamepads = _.assignIn({}, navigator.getGamepads());
-    if (_needToUpdate(newGamepads)) {
+    if (_needToUpdate(newGamepads) || Date.now() - timestamp > 1000) {
```
nikitakit: I hope that this one-second throttling does not affect Dawn-runtime communication initiated due to a change in the robot state.

nikitakit (dismissed):
Glad to see upload functionality being implemented!

It does appear that for now, the SFTP uploads have not been coordinated to use the correct paths for the BBB. It would be good to put the correct paths into the code.

dawn/main/MenuTemplate/DebugMenu.js
kevinmasd: Linter caught this; there's a [rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/first.md) in there to specify absolutes before relatives.

dawn/package.json
kevinmasd: Linter again, though I personally believe it is better to leave dev-tools for easier debugging on a student-by-student basis. May put under #35.

dawn/renderer/actions/AlertActions.js
kevinmasd: Linter again; [rule](http://eslint.org/docs/rules/arrow-parens) specifying no parentheses around single argument if what's coming after it isn't braces.

dawn/renderer/components/Editor.js
kevinmasd: pathToName is used in two contexts; one for file location and other for displaying file name above the Editor. Will change.

dawn/renderer/components/Editor.js
kevinmasd: Linter; not actually sure why it was caught on it.

dawn/renderer/components/Editor.js
kevinmasd: Will talk to DevOps and Runtime in-person about this on Wednesday.

dawn/renderer/components/Editor.js
kevinmasd: #96 If it's internet based, I'm going to need to know more further details. I was going to grab a pdf and add it into Dawn's static folders to load.

dawn/renderer/components/UpdateBox.js
kevinmasd: Would you rather UpdateBox pull in Editor's version of pathToName?

dawn/renderer/utils/sagas.js
kevinmasd: Shouldn't be affecting anything. Current implementation was to already send whenever gamepad information changes (i.e. any movement detected) and set a 50ms cooldown.

dawn/renderer/components/UpdateBox.js
kevinmasd: Done

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -28,52 +33,57 @@ const SENDRATE = 1000;
...
-const randomFloat = (min, max) => ((max - min) * Math.random() + min);
+const randomFloat = (min, max) => (((max - min) * Math.random()) + min);
 
 /**
  * Generate fake data to send to Dawn
  */
 const generateFakeData = () => [
   {
-    robot_state: Math.floor(randomFloat(0, 4)),
+    robot_state: state.valueOf(),
```
nikitakit: Using enum LGTM. Not sure you need `valueOf`, though.

dawn/renderer/components/UpdateBox.js
```diff
@@ -52,25 +53,66 @@ class UpdateBox extends React.Component {
 
   upgradeSoftware() {
     this.setState({ isUploading: true });
+    const update = Editor.pathToName(this.state.updateFilepath);
+    const signat = Editor.pathToName(this.state.signatureFilepath);
```
nikitakit: Signature files are deprecated. Last I checked, we decided not to bother with the extra hassle and use update files only.

nikitakit (commented):
Error handling looks good, though we might want to have more explicit error messages once we figure out exactly what conditions cause each error handler to trigger.

You still need to coordinate with devops about file paths (including removing signature files, which I believe are deprecated), but I'm removing my "request changes" status because functionality otherwise this looks good.

Note that I did not review for javascript-specific style, best practices, or quirks, since those aren't my area of expertise.

dawn/fake-runtime/FakeRuntime.js
kevinmasd: Fixed

dawn/package.json
AllanYangZhou: I don't think we need to add redux-devtools in the student builds, it slows down performance and its not as useful as the chrome devtools.

dawn/renderer/components/Editor.js
```diff
@@ -29,10 +27,27 @@ import 'brace/theme/solarized_dark';
...
+import ConsoleOutput from './ConsoleOutput';
+import EditorButton from './EditorButton';
+
+const Client = require('ssh2').Client;
+
 const dialog = remote.dialog;
 const currentWindow = remote.getCurrentWindow();
 
 class Editor extends React.Component {
+  static pathToName(filepath) {
```
AllanYangZhou: You can move generic functions like this into a `utility.js` file and import wherever you need.

dawn/renderer/components/ConfigBox.js
```diff
@@ -0,0 +1,100 @@
...
+    this.getValidationState = this.getValidationState.bind(this);
+    this.disableUploadUpdate = this.disableUploadUpdate.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+
+  getValidationState() {
+    const current = this.state.ipAddress;
+    let valid = false;
+    const regex = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}' +
```
AllanYangZhou: This test IP step can be moved into a generic utility file and imported (see other comments)

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +155,90 @@ class Editor extends React.Component {
...
+      return;
     }
-    console.log('Deprecated: sendCode');
-    console.log(`Debug Info: ${command} EOD`);
-    return true;
-  }
-
-  upload() {
-    this.sendCode('upload');
+    const conn = new Client();
```
AllanYangZhou, line 175: Should be moved into a saga.

dawn/renderer/components/Editor.js
```diff
@@ -240,21 +320,21 @@ class Editor extends React.Component {
           <ButtonGroup id="code-execution-buttons">
             <EditorButton
               text="Run"
-              onClick={this.startRobot}
+              onClick={() => { this.startRobot.call(this); }}
```
AllanYangZhou: Better to be consistent and bind `this.startRobot` in the constructor (same for all the other cases below)

dawn/renderer/components/UpdateBox.js
```diff
@@ -29,22 +40,54 @@ class UpdateBox extends React.Component {
...
-      filters: [{ name: 'Update signature', extensions: ['asc'] }],
-    }, (filepaths) => {
-      if (filepaths === undefined) return;
-      this.setState({ signatureFilepath: filepaths[0] });
-    });
-  }
-
   upgradeSoftware() {
-    console.log('Deprecated');
+    this.setState({ isUploading: true });
```
AllanYangZhou, line 34: This should also be moved into sagas.

AllanYangZhou (changes requested):

dawn/renderer/components/Editor.js
kevinmasd: Issue with moving into a saga is the amount of variables needed to be transmitted between processes. To be honest, these are also things that should be waited on.

kevinmasd: Also not sure how the Saga's generator can be dealt with.

dawn/renderer/components/Editor.js
```diff
@@ -134,48 +151,90 @@ class Editor extends React.Component {
...
+            });
+            throw err2;
+          }
+        });
+      });
+    }).connect({
+      debug: (inpt) => { console.log(inpt); },
+      host: this.props.ipAddress,
+      port: 22,
+      username: 'ubuntu',
```
nikitakit: So you have the port, username, and password hardcoded both here and in another file later on. Any chance you could move to a single source of truth for host/port/username/password?

Besides avoiding the two places in code falling out of sync, I have some simulator-related ideas that would involve replacing these hardcoded constants with configurable values.

AllanYangZhou (approved):

PR #110, opened 2017-02-02
[Dawn] Notification/TCP Proto

jianahuang10:
Deleted the TCPData submessage and created a Notification.proto to be used for TCP messages. The Notification.proto will be used for console logging on Dawn's side. The network protocol can be seen [here](https://github.com/pioneers/PieCentral/wiki/Dawn-Runtime-Network-Protocol).


kevinmasd (approved):
Will pull in before #101; only small changes needed.

PR #112, opened 2017-02-05
[HIBIKE] update firmware library to new protocol and implement example device

karthik-shanmugam:
- Implement parsing and constructing for new packet types in hibike/lib/hibike_message.cpp
- Implement new protocol behavior in hibike/lib/hibike_device.cpp
- Update ExampleDevice firmware to use updated hibike library api


hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,117 @@
...
-  append_buf(data_update_buf, &offset, (uint8_t *)&value0, sizeof(value0));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value1, sizeof(value1));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value2, sizeof(value2));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value3, sizeof(value3));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value4, sizeof(value4));
-  return offset;
-}
+//    return          -   number of bytes writted to data_update_buf on success; 0 otherwise
 
+uint8_t device_data_update(uint8_t param, uint8_t* data_update_buf, size_t buf_len) {
```
nikitakit: Do note that neither `device_data_update` nor `device_write` do any bounds checking based on the buffer length argument. This need not block merging, but I don't think its right to document that bounds checking must happen and then not do it. At least a TODO somewhere would be nice.

But if correct bounds checking is needed during the course of normal operation (and not just debugging), then it should absolutely be implemented here.

hibike/devices/ExampleDevice/example_device.h
```diff
@@ -2,23 +2,33 @@
...
 //////////////// DEVICE UID ///////////////////
 hibike_uid_t UID = {
   EXAMPLE_DEVICE,                      // Device Type
-  0,                      // Year
+  1,                      // Year
   UID_RANDOM,     // ID
 };
 ///////////////////////////////////////////////
 
+#define NUM_PARAMS 6 //change to however many params are needed
```
nikitakit: Where is `NUM_PARAMS` used? Based on the current diff it looks deprecated, in which case you should consider removing it.

hibike/devices/ExampleDevice/example_device.h
```diff
@@ -2,23 +2,33 @@
...
-
+// this is an optional device specific enum to make implementation easier
+typedef enum {
+  VALUE_0 = 1,
+  VALUE_1 = 11,
+  VALUE_2 = 12,
+  VALUE_3 = 13,
+  VALUE_4 = 14,
+  VALUE_5 = 7
+} param;
```
nikitakit: For style reasons, the name `param` should not be all lowercase. I think better style would be `Param`.

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,117 @@
...
+  switch (param) {
+    case VALUE_0:
```
nikitakit: I wonder if it would be better to namespace the enum when using it, e.g. `Param::VALUE_0`.

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -1,69 +1,117 @@
...
-  append_buf(data_update_buf, &offset, (uint8_t *)&value1, sizeof(value1));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value2, sizeof(value2));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value3, sizeof(value3));
-  append_buf(data_update_buf, &offset, (uint8_t *)&value4, sizeof(value4));
-  return offset;
-}
+//    return          -   number of bytes writted to data_update_buf on success; 0 otherwise
 
+uint8_t device_data_update(uint8_t param, uint8_t* data_update_buf, size_t buf_len) {
+  switch (param) {
```
nikitakit: If you're using an enum type, I think you should cast the variable `param` to the enum. IIRC switching on an enum value will warn if there are any values of the enum that are not handled, but I'm not sure if switching on a uint8_t will do that. (Fix would be something like `switch ((ParamEnum)param)`, or similar but using `static_cast`).

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -1,11 +1,9 @@
 #include "hibike_device.h"
 
-char *DESCRIPTION = DESCRIPTOR;
-
 message_t hibikeBuff;
-
 uint64_t prevTime, currTime, heartbeatTime;
-uint8_t param;
+// uint8_t param;
```
nikitakit: Delete instead of commenting out.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -31,47 +29,63 @@ void hibike_loop() {
...
-          value = *((uint32_t*) &hibikeBuff.payload[DEVICE_PARAM_BYTES]);
-          send_device_response(param + 1, device_update(param, value));
-          break;
+        case DEVICE_READ:
 
-        case DEVICE_STATUS:
-          param = hibikeBuff.payload[0] - 1;
-          send_device_response(param + 1, device_status(param));
+          send_data_update(*((uint16_t*) &hibikeBuff.payload[0]));
+          // param = hibikeBuff.payload[0] - 1;
```
nikitakit: delete instead of commenting out.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -31,47 +29,63 @@ void hibike_loop() {
...
-        case PING_:
-          send_subscription_response(&UID, subDelay);
+        case PING:
+          send_subscription_response(params, subDelay, &UID);
           break;
 
-        case DESCRIPTION_REQUEST:
-          send_description_response(DESCRIPTION);
-          break;
+        // case DESCRIPTION_REQUEST:
```
nikitakit: delete instead of commenting out. Add a TODO if this feature will ever need to be re-implemented.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -120,17 +134,23 @@ void hibike_loop() {
   }
 
   // DataUpdates
+  // if ((subDelay > 0) && (currTime - prevTime >= subDelay)) {
```
nikitakit: Delete instead of commenting out. Add TODO if this functionality will need to be re-enabled later.

hibike/lib/hibike/hibike_message.cpp
```diff
@@ -81,90 +81,69 @@ int read_message(message_t *msg) {
...
 
   if (status != 0) {
     return -1;
   }
   return send_message(&msg); 
 }
 
-
-int send_data_update(uint8_t* data, uint8_t payload_length) {
+//deprecated but not really
```
nikitakit: Uh... what is this comment supposed to mean?

If this is not actually deprecated, the comment should be more descriptive. If it is, why is the code still here?

If this not being deleted yet has to do with the order you're submitting PRs, then let me know (odd intermediate states is a normal occurrence, and usually isn't a problem)

hibike/lib/hibike/hibike_message.cpp
```diff
@@ -81,90 +81,69 @@ int read_message(message_t *msg) {
...
+  msg.messageID = DEVICE_DATA;
+  msg.payload_length = 2;
+  for (uint16_t count = 0; (params >> count) > 0; count++) {
+      if (params & (1<<count)){
+        int bytes_written = device_data_update((uint8_t) count, &msg.payload[msg.payload_length], (size_t) msg.payload_length);
+        if(bytes_written){
+          msg.payload_length += bytes_written;
+        }
+        else{
+          params &= ~(1<<count);
```
nikitakit: This else condition look like a no-op. Should it be deleted?

nikitakit (changes requested):
This PR is very clean and does only one thing. Thanks for splitting things up, it's been a pleasure to review!

I have a few stylistic comments, but so far nothing major stood out to me (though I may notice some more things when I take a second look a little later)

hibike/lib/hibike/hibike_message.cpp
karthik-shanmugam: It's supposed to unset the corresponding bit of `params` to indicate that a value for that param is not contained within the packet. The else clause will occur if `device_data_update()` writes nothing to the buffer for a specific param, possibly due to insufficient buffer space or an invalid param.

hibike/lib/hibike/hibike_device.h
```diff
@@ -35,12 +36,8 @@ void hibike_setup();
...
 void toggleLED();
 
-extern uint32_t device_update(uint8_t param, uint32_t value); 
-
-extern uint32_t device_status(uint8_t param); 
-
-extern uint8_t data_update(uint8_t* data_update_buf, size_t buf_len);
-
+extern uint32_t device_write(uint8_t param, uint8_t* data,size_t len);
+extern uint8_t device_data_update(uint8_t param, uint8_t* data_update_buf, size_t buf_len);
```
karthik-shanmugam: I think `device_data_update()` should be renamed to `device_read()` and have the same argument names as `device_write()` to make things cleaner. This change would be need to be made in multiple places:
 - The hibike firmware library
 - Firmware for each specific device
 - DEVELOPERS.md

hibike/devices/ExampleDevice/example_device.cpp
karthik-shanmugam: resolved by adding bounds checking

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -31,46 +28,56 @@ void hibike_loop() {
...
+            if (params & (1<<count)){
+              int status = device_write(count, &hibikeBuff.payload[offset], hibikeBuff.payload_length-offset);
+              if(status){
+                offset += status;}
+              else{
+                params = params & ~(1<<count);}
+              }
+          }
+
+          // params = hibikeBuff.payload[0] - 1;
```
nikitakit: Commented out code should be removed

hibike/lib/hibike/hibike_message.cpp
nikitakit: You're right. I must have been asleep when reading the code last time.

hibike/lib/hibike/hibike_device.cpp
nikitakit: FYI I looked over your latest changes, and this comment is currently the only non-ideal thing in the code that I've found.

nikitakit (approved):
Good catch on the payload length calculation changes!

This code now LGTM, and I think any remaining bugs can be addressed in future PRs.

(Sorry for late reply. FYI when you pushed the last commit to the PR I don't remember getting notified, probably because github notifies on comments but not on commits)

PR #113, opened 2017-02-05
Runtime/angelina/async await

Angelina-Wang:
[RUNTIME]
-Added to Yizhe's async things, and tried to incorporate pieces in studentAPI
-Unfinished tests are in studentCode


yizow:
Looks mostly good to me, but what do you mean by "unfinished tests"?

Angelina-Wang:
I just wasn't sure it was right, like I'm not exactly sure how to test it out so I was going to look into that a little more to make sure the tests actually test things

runtime/testy/runtime.py
```diff
@@ -267,6 +294,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+2. Calling an `async def` function from inside `setup` or `loop` without using
+`Robot.coroutine`.
+
+Incorrect code:
+    def loop():
+        my_coro()
+
+Consider instead:
+    def loop():
+        Robot.coroutine(my_coro)
```
kevinji: This should be `Robot.run`.

runtime/testy/runtime.py
```diff
@@ -267,6 +294,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+Incorrect code:
+    async def my_coro():
+        delay(1.0)
+
+Consider instead:
+    async def my_coro():
+        await delay(1.0)
+
+2. Calling an `async def` function from inside `setup` or `loop` without using
+`Robot.coroutine`.
```
kevinji: This should be `Robot.run`.

runtime/testy/studentAPI.py
```diff
@@ -7,6 +11,40 @@ def __init__(self, toManager, fromManager):
...
+        raise ValueError("First argument to Robot.coroutine must be a function")
+    elif not inspect.iscoroutinefunction(fn):
+        raise ValueError("First argument to Robot.coroutine must be defined with `async def`, not `def`")
+
+    if fn in self._coroutines_running:
+        return
+
+    self._coroutines_running.add(fn)
+
+    future = fn(*args, **kwargs)
```
kevinji, line 41: You can probably move this into the function below.

kevinji (dismissed):
Seems like some comments just didn't get updated when the corresponding functions were renamed.

runtime/testy/runtime.py
```diff
@@ -126,25 +128,51 @@ def checkTimedOut(func, *args):
...
+        next_call = loop.time() + 1. / RUNTIME_CONFIG.STUDENT_CODE_HZ.value
+        checkTimedOut(mainFunc)
+
+        sleep_time = max(next_call - loop.time(), 0.)
+        stateQueue.put([SM_COMMANDS.STUDENT_MAIN_OK, []])
+        execCount += 1
+        await asyncio.sleep(sleep_time)
+
+      badThingsQueue.put(BadThing(sys.exc_info(), "Process Ended", event=BAD_EVENTS.END_EVENT))
+      if not (maxIter is None or execCount < maxIter):
```
nikitakit: In retrospect, I think a condition of `if exception_cell[0] is not None` would be cleaner and possibly more correct.

runtime/testy/runtime.py
```diff
@@ -126,25 +128,51 @@ def checkTimedOut(func, *args):
...
+      raise RuntimeError("Student code failed to define '{}'".format(test_name + "setup"))
+    try:
+      mainFunc = getattr(studentCode, testName + "main")
+    except AttributeError:
+      raise RuntimeError("Student code failed to define '{}'".format(test_name + "main"))
+
+    ensure_is_function(testName + "setup", setupFunc)
+    ensure_is_function(testName + "main", mainFunc)
+    ensure_not_overridden(studentCode, 'Robot')
+    ensure_not_overridden(studentCode, 'delay')
```
nikitakit: This line is no longer needed. Instead, you should ensure that `Actions` is not overridden (but that's once `Actions` is actually defined)

runtime/testy/runtime.py
```diff
@@ -126,25 +128,51 @@ def checkTimedOut(func, *args):
...
+    async def main_loop():
+      execCount = 0
+      while (exception_cell[0] is None) and (maxIter is None or execCount < maxIter):
+        next_call = loop.time() + 1. / RUNTIME_CONFIG.STUDENT_CODE_HZ.value
+        checkTimedOut(mainFunc)
+
+        sleep_time = max(next_call - loop.time(), 0.)
+        stateQueue.put([SM_COMMANDS.STUDENT_MAIN_OK, []])
+        execCount += 1
+        await asyncio.sleep(sleep_time)
```
nikitakit, line 178: @yizow Could you test whether this sleep logic is correct, and also that the introduction of the event loop does not lead to a drastic performance hit?

runtime/testy/runtime.py
```diff
@@ -267,6 +295,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+def ensure_not_overridden(module, name):
+  if hasattr(module, name):
+    raise RuntimeError("Student code overrides `{}`, which is part of the API".format(name))
+
+def clarify_coroutine_warnings(exception_cell):
+    """
+    Python's default error checking will print warnings of the form:
+        RuntimeWarning: coroutine '???' was never awaited
+
+    This function will inject an additional clarification message about what
```
nikitakit: This should be changed to say:

This function will upgrade such a warning to a fatal error, while also injecting a additional clarification message about possible causes.

runtime/testy/runtime.py
```diff
@@ -267,6 +295,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+            print("""
+The PiE API has upgraded the above RuntimeWarning to a runtime error!
+
+This error typically occurs in one of the following cases:
+
+1. Calling `delay` or anything in `Robot.actions` without using `await`.
+
+Incorrect code:
+    async def my_coro():
+        delay(1.0)
```
nikitakit: `Actions.sleep`, not delay

runtime/testy/runtime.py
```diff
@@ -267,6 +295,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+
+Consider instead:
+    async def my_coro():
+        await delay(1.0)
```
nikitakit: `Actions.sleep`, not delay

runtime/testy/runtime.py
```diff
@@ -267,6 +295,64 @@ def startHibike(badThingsQueue, stateQueue, pipe):
...
+
+        if str(message).endswith('was never awaited'):
+            coro_name = str(message).split("'")[-2]
+
+            print("""
+The PiE API has upgraded the above RuntimeWarning to a runtime error!
+
+This error typically occurs in one of the following cases:
+
+1. Calling `delay` or anything in `Robot.actions` without using `await`.
```
nikitakit: "Calling `Actions.sleep` or anything in `Actions` without using `await`"

runtime/testy/studentAPI.py
nikitakit: I'm not sure this can be moved. Calling the function here means that any errors involved in calling the function (such as invalid number of arguments) get caught immediately and print decent stack traces, as opposed to getting caught later inside `wrapped_future`, where the stack trace would point into the event loop.

runtime/testy/studentAPI.py
```diff
@@ -7,6 +11,40 @@ def __init__(self, toManager, fromManager):
...
+    independently of the main loop of code.
+
+    The first argument must be a function defined with 'async def'.
+
+    The remaining arguments are then passed to that function before it is
+    called.
+
+    Multiple simultaneous coroutines that use the same robot actuators will
+    lead to surprising behavior. To help guard against errors, calling
+    `coroutine` with a `fn` argument that is currently running is a no-op.
```
nikitakit: `run`, not `coroutine`

runtime/testy/studentAPI.py
```diff
@@ -7,6 +11,40 @@ def __init__(self, toManager, fromManager):
     self.fromManager = fromManager
     self.toManager = toManager
     self._createSensorMapping()
+    self._coroutines_running = set()
+
+  def run(self, fn, *args, **kwargs):
```
nikitakit, line 16: Note that the counterpart `is_running` function is not currently part of this PR.

nikitakit (changes requested):
I made a few comments, mostly relating to how the student API spec uses slightly different names than the demo code (meaning that this PR needs names changed in a few places)

runtime/testy/studentAPI.py
kevinji: Oh okay, that makes sense. Ignore this comment then.

nikitakit (changes requested):
This change LGTM, but it needs to be converted to a state where it can clearly merge with master.

Note that this doesn't implement `Actions.sleep` or `Robot.is_running`, which we must make sure to take care of before shipping the API.

jiangdaniel:
@nikitakit Is this PR good to go?

nikitakit (approved):

nikitakit:
@jiangdaniel Yes, looks good. I'm going to open an issue about the remaining API methods that still need to be implemented.

PR #114, opened 2017-02-09
[RUNTIME] Dynamic IP change for Ansible

bodiddles:
IP that Ansible sends to changes according to first message sent to us from Dawn.
    
Tested with fake_dawn but requires integration testing with real Dawn as they need to implement new communications protocol.
    
UDP receive waits to receive initial message and then messages runtime to spawn the UDP send process with known ip address.

### DIFF
--- a/runtime/testy/Ansible.py
+++ b/runtime/testy/Ansible.py
@@ -77,6 +77,7 @@ class UDPSendClass(AnsibleHandler):
         self.sendBuffer = TwoBuffer()
         packagerName = THREAD_NAMES.UDP_PACKAGER
         sockSendName = THREAD_NAMES.UDP_SENDER
+        self.dawn_ip = None
         super().__init__(packagerName, UDPSendClass.packageData, sockSendName,
                          UDPSendClass.udpSender, badThingsQueue, stateQueue, pipe)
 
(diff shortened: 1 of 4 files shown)


runtime/testy/runtime.py
```diff
@@ -47,10 +47,10 @@ def nonTestModePrint(*args):
 
   try:
     spawnProcess(PROCESS_NAMES.STATE_MANAGER, startStateManager)
-    spawnProcess(PROCESS_NAMES.UDP_SEND_PROCESS, startUDPSender)
     spawnProcess(PROCESS_NAMES.UDP_RECEIVE_PROCESS, startUDPReceiver)
     spawnProcess(PROCESS_NAMES.HIBIKE, startHibike)
     controlState = "idle"
+    connectionState = False
```
yizow: consider changing to something like `dawn_connected`

runtime/testy/Ansible.py
```diff
@@ -77,6 +77,8 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
         self.sendBuffer = TwoBuffer()
         packagerName = THREAD_NAMES.UDP_PACKAGER
         sockSendName = THREAD_NAMES.UDP_SENDER
+        stateQueue.put([SM_COMMANDS.SEND_ADDR, [PROCESS_NAMES.UDP_SEND_PROCESS]])
+        self.dawn_ip = pipe.recv()[0]
```
yizow, line 81: i dont think you need to `[0]`

yizow (changes requested):
Minor bug/ style

runtime/testy/Ansible.py
brandonxxlee: Technically what is saved in stateManager is the address of dawn, which is a tuple of (IP, Port). Here we are just pulling out the IP from the address. Perhaps we don't ever need the port, and should filter it out from the beginning?

yizow (approved):
LGTM

PR #118, opened 2017-02-09
[DAWN] package.json: Shorten scripts

kevinji:
`node_modules/.bin/` is automatically added to PATH, so hardcoded paths are unnecessary.


kevinmasd (approved):
Modified due to build process complications.

PR #123, opened 2017-02-11
[HIBIKE] handle errors when opening serial port in hibike_process.py

karthik-shanmugam:
This issue was getting triggered when the shutting down of a virtual device left virtual_devices.txt with invalid filenames. This issue revealed a bigger issue in hibike_process.py; that any errors when opening serial ports crashes the entire program. 

This is a somewhat jank fix that just pushes down the error to the invalid port's specific read/write threads, which won't crash the whole program. @97hongjun is working on refactoring how hibike_process.py manages its worker threads, and his work in the future should solve this issue (and other issues like hot plugging) more elegantly. Meanwhile, this temporary fix should allow other teams using hibike (YogiBear and runtime) to use hibike's python side with fewer issues.


nikitakit:
So this is quite a hack.

Normally I wouldn't merge something like this, but I'll make an exception given that other projects are adversely impacted by this bug.

However, could you first add a comment explaining that the None pushes errors down into the individual threads, and that a better fix is on the TODO list? That way there's less of a danger of this temporary fix becoming permanent.

nikitakit (approved):

dHutchings:
I'd like to re-open this, if I may.

@ArjunSridhar1 @jonathan-j-lee 

Last night, I came up with a udev rule that guarantees that all Arduino Pro Micros (the devices used over hibike) show up as /dev/apm0, /dev/apm1, etc.  Udev rules are used by the "Device Manager" for linux, so it will work w/out us running any code.

This means that hibike no longer needs to try-catch /dev/tty* and /dev/usb*.

The udev rules is as follows:
# Arduino Pro Micro                            

ACTION=="add", SUBSYSTEMS=="usb", KERNEL=="tty*", ATTRS{idVendor}=="2341", ATTRS{idProduct}=="8037", \
    MODE:="0666", \
    SYMLINK:="apm%n"

Please let me (and devops, who i added) know if this is an appropriate improvement.

PR #128, opened 2017-02-16
Runtime/fix udp tests

bodiddles:
Checks if the UDP send process is actually spawned before trying to terminate it in the runtime tests.


runtime/testy/runtime.py
```diff
@@ -91,7 +91,7 @@ def nonTestModePrint(*args):
           continue
         elif newBadThing.event == BAD_EVENTS.ENTER_IDLE and controlState != "idle":
           break
-        print(newBadThing)
+        print(newBadThing.event)
```
nikitakit, line 94: Does this change to printing affect the testing framework at all?

nikitakit (approved):
As long as tests pass with this change (see my comment about the print statement on why they might not), I'm happy with this change.

Please squash the commits when merging it in, though.

yizow:
Make sure to remember to change commit message to use `[RUNTIME]` tag

PR #131, opened 2017-02-16
[DevOps] Set up Travis to build dawn and runtime

nikitakit:
By default, travis will run dawn tests, runtime tests, and package
a runtime update tarball (but not do anything with it).

If a commit is tagged starting with dawn/ or test/dawn/, it will
additionally build Dawn binaries and upload them to GitHub releases.

If a commit is tagged starting with runtime/ or test/runtime, it
will additionally upload the runtime update tarball to GitHub
releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,51 @@
+dist: trusty
+sudo: required
 language: node_js
 node_js:
 - "6"
 os:
-- osx
 - linux
+env:
+  - TEST_DIR=dawn
+  - TEST_DIR=runtime
+before_install:
+- mkdir -p artifacts
+- if [[ $TEST_DIR = dawn ]] && git tag -l --points-at HEAD | grep dawn; then
+    export DEPLOY_ARTIFACTS=1;
+  fi
+- if [[ $TEST_DIR = runtime ]] && git tag -l --points-at HEAD | grep runtime; then
+    export DEPLOY_ARTIFACTS=1;
+  fi
+- mkdir -p build-deps
+- export PATH="$PATH:$PWD/build-deps/bin"
 install:
-- cd dawn
-- make dawn-install
+- pushd $TEST_DIR
+- make install
+- if [[ $DEPLOY_ARTIFACTS = 1 ]]; then
+    make artifacts-install;
+  fi
+- popd
 script:
-- make dawn-test
-- make dawn-lint
+- pushd $TEST_DIR
+- make test
+- make lint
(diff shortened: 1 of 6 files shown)


nikitakit:
Before this can be merged, the following fixes need to be made to the repo:
* (Edit: fixed) `python3 runtime.py --test` needs to be working again (#128)

The following additional changes are desirable (but not required to merge this):
* (Edit: fixed) Changing `python runtime.py --test` to return nonzero exit code if any tests failed
* (Edit: fixed) Fix Dawn builds such that the resulting executables actually run (instead of just compiling)

kevinji:
To make the Travis file cleaner, perhaps you could [use a build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) and having a travis.sh file in each subdirectory that runs each project's own tests? This would also clean up test results by separating per project.

nikitakit:
@mc10 So I'm actually not very familiar with travis. The impression I got was that build matrix lets you set up multiple runs with different VM types, but not so easy to branch on what to build (except via environment vars). Also sadly I'm not sure if you can add sudo:required to the matrix (requiring sudo increases build time by up to a minute, on each matrix element that uses it, so it would be nice to make it conditional.) I may still want to refactoring some of the logic out of travis.yml, though.

kevinji:
Yeah, Travis is still pretty limiting, but it seems like you can use environment vars to customize the build directory:

```
language: python
os:
  - linux
python:
  - "3.6"
env:
  - TEST_DIR=dawn
  - TEST_DIR=runtime
before_install:
  - nvm install node
  - cd $TEST_DIR
install:
  - bash travis-install.sh
script:
  - bash travis.sh
branches:
  only:
    - master
```

This assumes that within each project's subdirectory, we have `travis-install.sh` and `travis.sh` files.

nikitakit:
@mc10 I just pushed a version that splits up the build.

The artifacts add a bit of complication, because building Dawn takes up to 10 minutes and I don't want to do that on every commit. I'm also making it so that dawn and runtime release tags are separate, to avoid confusion.

nikitakit:
Now that the requisite changes in runtime/dawn have been made, I have tested these travis updates and they run cleanly.

Could someone please code review these changes?

(Note that the runtime emergencyStop test is somewhat flaky. Re-running the runtime travis job gets it to pass again -- I've talked to runtime about investigating this)

AllanYangZhou (approved):
Dawn-related stuff LGTM.

PR #136, opened 2017-02-16
[RUNTIME] Use requirements.txt for pip dependencies

kevinji:
If I'm missing any dependencies, feel free to add to the list.


nikitakit:
Don't think runtime has additional dependencies, but it should depend on hibike, with transitively brings in a dependency on at least pyserial. (However I'm not sure if the hibike integration is in a working state or not)

See also https://github.com/pioneers/PieCentral/blob/master/DevOps/frankfurter/scripts/install/master_frank_setup.sh#L34 (which is not up to date, and would ideally not duplicate this functionality)

yizow (approved):
Sigh... I'm pretty sure there's more : [

But I haven't been good at keeping track of them.
Nonetheless, this is a good first step.

PR #138, opened 2017-02-18
[RUNTIME] TCP Integration

brandonxxlee:
Adds TCP support for console logging and notifications with Dawn.

TCP process spawns once first message is received from Dawn and will send notifications and console logs via the notification proto over the tcp socekt.

Adds throttling to UDP receive process as well to hopefully improve performance.

Included compiled notification proto for now.


runtime/testy/Ansible.py
```diff
@@ -5,6 +5,8 @@
 import selectors
 import runtime_pb2
 import ansible_pb2
+import notification_pb2
+
```
nikitakit: No need to add a newline here

runtime/testy/Ansible.py
```diff
@@ -223,15 +230,99 @@ def start(self):
         try:
             while True:
                 nextCall = time.time()
+                nextCall += 1.0/UDPRecvClass.UDP_RECV_HZ
```
nikitakit: You have a call to `select` right under this, which should be blocking. In that case, there should be absolutely no reason to have sleeps anywhere in this loop.

runtime/testy/Ansible.py
```diff
@@ -223,15 +230,99 @@ def start(self):
...
+                return proto_message.SerializeToString()
+            except Exception as e:
+                badThingsQueue.put(BadThing(sys.exc_info(), 
+                    "TCP packager crashed with error: " + str(e),
+                    event = BAD_EVENTS.TCP_ERROR,
+                    printStackTrace = True))
+        while True:
+            try:
+                rawMessage = pipe.recv()
+                nextCall = time.time()
```
nikitakit, line 282: Note that the sleeps in this thread appear to be for throttling purposes, rather than for handling use of non-blocking calls. There should be a comment indicating this.

nikitakit (changes requested):
Just a few minor style comments.

Otherwise code LGTM, but it might be good to get another reviewer because I'm not very familiar with the networking part of the code.

runtime/testy/Ansible.py
```diff
@@ -223,15 +230,99 @@ def start(self):
         try:
             while True:
                 nextCall = time.time()
+                nextCall += 1.0/UDPRecvClass.UDP_RECV_HZ
                 event = sel.select()
                 self.udpReceiver()
                 self.unpackageData()
                 try:
-                    time.sleep(nextCall - time.time())
+                    time.sleep(nextCall +  - time.time())
```
yizow: So this line is wrong i think

But re: nikita's comments we shouldn't need this line at all

runtime/testy/Ansible.py
```diff
@@ -222,16 +228,101 @@ def start(self):
 
         try:
             while True:
-                nextCall = time.time()
                 event = sel.select()
```
nikitakit, line 229: My previous comment regarding `select` remains unresolved.

nikitakit (changes requested):

runtime/testy/Ansible.py
```diff
@@ -151,6 +156,7 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
 
 class UDPRecvClass(AnsibleHandler):
     RECV_PORT = 1236
```
yizow: We should be using the global variables at the top of the file.

yizow (changes requested):

runtime/testy/Ansible.py
bodiddles: Could we use the sleeps for throttling reasons as we will be probably be receiving files faster than we want to be running this thread?

nikitakit: This is a receiver, not a sender. Throttling belongs in sender (in this case, Dawn), not the receiver.

If you want to handle receiving packets faster than this thread can handle, the correct was to implement this is by dropping packets.

nikitakit: Also, we went over this together during the fall semester. The way to implement rate-limited UDP receiver class is:

* Block until there is any data (using `select`)
* Read all available data from the kernel, dropping all but the last packet
* Process the single packet

Do you remember discussing this?

---

The reason sleeping is such a bad idea here is that it forces the kernel packet buffer to fill up, adding system latency equal to the kernel buffer size. If that happens you're going to end up with ridiculously large delays between moving a gamepad and the robot motors reacting. The whole reason we have a stateless UDP system is that any individual packet can be dropped without impacting the system as a whole.

runtime/testy/Ansible.py
```diff
@@ -222,16 +228,101 @@ def start(self):
...
+    def receiver(self, badThingsQueue, stateQueue, pipe):
+        def unpackage(data):
+            received_proto = notification_pb2.Notification()
+            received_proto.ParseFromString(data)
+            return received_proto
+        try:
+            while True:
+                recv_data, addr = self.sock.recvfrom(2048)
+                unpackagedData = unpackage(recv_data)
+                if not unpackagedData:
```
nikitakit: EDIT: I misread this the first time.

Please check that `recvfrom` is actually a blocking call.

This condition of `if not unpackagedData` threw me off, and part of it is that unpackage currently doesn't ever return None. I think this condition should be removed, or an exception handler should be added to `unpackage`

runtime/testy/Ansible.py
nikitakit: Upon further consideration, I don't think sleeps are the correct throttling mechanism here, either. But I'd be fine with merging this as-is, too, because it's not *too* bad.

Remember how you were debugging an issue where half of the message types were silently being dropped by the kernel? Perhaps that's actually the behavior we want -- your test code might have been spamming things at max speed, but the student code certainly won't be.

runtime/testy/Ansible.py
bodiddles: I changed the code to better reflect this.

runtime/testy/Ansible.py
```diff
@@ -150,15 +157,15 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
                     printStackTrace = True))
 
 class UDPRecvClass(AnsibleHandler):
-    RECV_PORT = 1236
-
+    UDP_RECV_HZ = 20
```
nikitakit: Delete this, you don't need `UDP_RECV_HZ` since you're not sleeping

runtime/testy/Ansible.py
```diff
@@ -150,15 +157,15 @@ def udpSender(self, badThingsQueue, stateQueue, pipe):
...
+
```
nikitakit: no need to make this whitespace change

runtime/testy/Ansible.py
```diff
@@ -5,15 +5,23 @@
 import selectors
 import runtime_pb2
 import ansible_pb2
-
+import notification_pb2
 from runtimeUtil import *
 
+SEND_PORT = 1235
```
nikitakit: Maybe name this `UDP_SEND_PORT`/`UDP_RECV_PORT` to distinguish from TCP now that TCP exists?

nikitakit (changes requested):
Just minor style comments this time.

Maybe @yizow should also review though, since I've had a bad track record of catching networking bugs.

runtime/testy/Ansible.py
```diff
@@ -222,16 +227,90 @@ def start(self):
...
+        try:
+            while True:
+                recv_data, addr = self.sock.recvfrom(2048)
+                unpackagedData = unpackage(recv_data)
+                stateQueue.put([SM_COMMANDS.STUDENT_UPLOAD, []])
+        except Exception as e:
+                badThingsQueue.put(BadThing(sys.exc_info(), 
+                    "TCP receiver crashed with error: " + str(e),
+                    event = BAD_EVENTS.TCP_ERROR,
+                    printStackTrace = True))
```
nikitakit: You should keep the newline at eof

runtime/testy/Ansible.py
```diff
@@ -69,10 +77,9 @@ def start(self):
         socketThread.join()
 
 class UDPSendClass(AnsibleHandler):
-    SEND_PORT = 1235
     packagerHZ = 20.0
     socketHZ = 20.0
-
+
```
nikitakit: gratuitous whitespace change

nikitakit (approved):
lgtm now, ideally @yizow would also give this a look

runtime/testy/Ansible.py
yizow: The problem with throttling the sending liek this is that if whatever is on the other end of the `pipe` is generating data faster than this thread processes, its gonna fill up and cause a constant latency, as everything on the other end starts blocking on this function. 

Or, if we use non-blocking, the stuff on the other end will error out with `pipe full` errors. 
Especially since this is TCP, it doesn't seem like we want to drop any of these packets. 

It looks like this is purely used for 
1) confirming student code uploaded
2) sending over console output

The former is low frequency/ bandwidth and likely won't cause an issue.
The latter is potentially high frequency, high bandwidth, and most likely WILL cause an issue. Especially if you're throttling.

Maybe batching messages in the pipe together and sending them that way? 


In the interest of getting this merged though, I'm also okay with merging this.
But the console logging commit better fix it.

yizow (approved):

PR #139, opened 2017-02-18
[RUNTIME] Adds Student Code Console Logging

brandonxxlee:
Overrides print for studentCode process and redirects to TCP socket to be sent to dawn.

Redirects BadThing stack trace to TCP socket to be sent to dawn.

Still maintains regular stdout print for testing purposes.

Note: This is branched from runtime/TCPIntegration so includes a lot of other commits, don't merge until other PR is merged.


nikitakit (changes requested):
Could you please figure out why 11/11 tests are failing on travis?

nikitakit:
This mostly looks good, but I'll wait for TCP to be merged before reviewing it.

yizow:
It looks like this shares a lot of commits with #138 TCP implementation
If that gets squashed and merged, will that cause merge conflicts here because of the squash?

runtime/testy/runtime.py
```diff
@@ -139,9 +141,9 @@ def checkTimedOut(func, *args):
       testName += "_"
     setupFunc = getattr(studentCode, testName + "setup")
     mainFunc = getattr(studentCode, testName + "main")
-
```
nikitakit: This whitespace noise really clutters diffs. I think it's a good habit to run `git diff` before every commit, and undoing accidental whitespace changes before sending anything to github.

runtime/testy/studentAPI.py
```diff
@@ -1,6 +1,7 @@
 import csv
 
 from runtimeUtil import *
+import builtins as __builtin__
```
nikitakit: No need to import builtins

nikitakit (changes requested):
Just a few minor style comments.

nikitakit (approved):
lgtm

yizow (changes requested):
Test student code sending multiple items in one loop

nikitakit:
Changes still lgtm, though the throttling business is still slightly unsatisfying. I think we should consider adopting a more principled approach after first shipping v1 of a working integrated system.

That said, @yizow reminds me that there are no associated tests. Ideally, there would be some way of having travis CI be able to detect changes that break TCP functionality, so we don't discover this by way of shipping broken releases.

runtime/testy/Ansible.py
```diff
@@ -236,7 +236,7 @@ def start(self):
             printStackTrace = True))
 
 class TCPClass(AnsibleHandler):
-    TCP_HZ = 25
+    TCP_HZ = 180
```
yizow: Move to global constant

runtime/testy/runtime.py
```diff
@@ -95,6 +95,8 @@ def nonTestModePrint(*args):
         print(newBadThing.event)
         nonTestModePrint(newBadThing.data)
         if newBadThing.event in restartEvents:
+          if newBadThing.event in studentErrorEvents:
+            stateQueue.put([SM_COMMANDS.SEND_CONSOLE, [newBadThing.stackTrace]])
```
yizow: I dont think this contains the actual error. (newBadThing.event)

you can just pass in newBadThing as an argument though, instead of newBadThing.stacktrace.
newBadThing objects have a built in print method that does this already i believe.

yizow (changes requested):
Few minor changes

runtime/testy/runtime.py
bodiddles: I think the newBadThing.event contains the enum for the actual bad event that we are checking in this case.

Also, the send console command expects a string rather than an object and we are directly setting a field in the proto to the argument used. We aren't printing it so I don't think the str method of the object would be called.

yizow (approved):

PR #142, opened 2017-02-19
[HIBIKE] runtime integration fixes

karthik-shanmugam:
- update `hibike_process.py` to send uid's along with device values, as was written in the integration spec (fix implementation to match documentation)
- update and rename integration spec
  - add `[]`'s around arguments to accurately reflect how contents of messages are nested (fix documentation to match implementation)
  - rename ATALANTA => RUNTIME


yizow (dismissed):
LGTM. 

Does the implementation change break anything? Cause i don't see the 'receiving end' changes

yizow (commented):
Mostly LGTM, but does the implementation change break anything? Cause i don't see the 'receiving end' changes

karthik-shanmugam:
@yizow  The only "receiving end" within hibike was just printing out the whole message to console anyways. If runtime did any integration around this it would almost definitely break that though.

yizow (approved):
`device_values` command is not actually implemented on runtime side. Adding to my todo

PR #143, opened 2017-02-19
[DAWN] webpack: Use ES6 and simplify config

kevinji:
This also removes the redundant production file by combining them into one file.


dawn/webpack.production.config.js
```diff
@@ -1,62 +0,0 @@
...
-const plugins = [
-  new webpack.DefinePlugin({
-    VERSION: JSON.stringify(require('./package.json').version)
-  }),
-  new webpack.DefinePlugin({
-    "global.GENTLY": false
-  }),
-  new webpack.DefinePlugin({
-    'process.env': {
-      'NODE_ENV': '"production"'
```
kevinmasd, line 34: We keep a production variable to differentiate what is being developed vs what should be shipped to students

kevinmasd (changes requested):
I would leave off handling Webpack. This can be done later when we can upgrade from Webpack 1 to 2.

dawn/webpack.production.config.js
kevinmasd: Actually, after seeing package.json, I'll get back to you on this.

kevinji:
I don't think I've changed any of the code of the webpack file, just ES6-ifyed it. Updating to Webpack 2.0 shouldn't really be affected imo.

kevinji:
@kevinmasd I think this is mergeable without worrying about Webpack 2.0; you should be able to make that upgrade right on top of these changes.

dawn/webpack.config.babel.js
```diff
@@ -0,0 +1,58 @@
+import path from 'path';
+import webpack from 'webpack';
+import { version } from './package.json';
+
+const target = process.env.NODE_ENV === 'production' ? 'electron' : 'atom';
```
kevinmasd: This was a mistake in the originals; we should probably default to electron.

kevinmasd (changes requested):

dawn/webpack.config.babel.js
kevinji: So should this always be `'electron'`?

kevinmasd: Yes

kevinji: Fixed in latest commit.

kevinmasd (approved):

PR #144, opened 2017-02-20
[RUNTIME] runtime.py: studentCode timing bugfix

yizow:
Calculates the time for execution of the next studentCode main loop
*before* we execute the main loop.


nikitakit (approved):
lgtm

Issue #141, opened 2017-02-19
[Dawn] Release build of Dawn crashes on exit


nikitakit:
```
Uncaught Exception:
TypeError: Cannot read property 'kill' of undefined
    at BrowserWindow.<anonymous> (/private/var/folders/kx/1dxccvcx5bvcbs426mjjjyymbm5qnj/T/AppTranslocation/1F8B67D6-CEE7-4818-8AA3-F453F0315027/d/dawn.app/Contents/Resources/app.asar/build/main.js:1:680)
    at emitOne (events.js:96:13)
    at BrowserWindow.emit (events.js:188:7)
```

@kevinmasd said he thinks he knows why this is happening.

PR #146, opened 2017-02-20
[Dawn] Production Checks on Fake Runtime

kevinmasd:
Addresses #141


nikitakit (approved):
Diff lgtm

Note that I haven't actually tested this, because that requires kicking off a build. I trust that you tested this before sending the PR though.

PR #148, opened 2017-02-23
[HIBIKE] TeamFlag firmware

karthik-shanmugam:
Hopefully this is a pretty straightforward pull request.

Team flag firmware is now basically just 6 individually controllable LEDs. It doesn't support any of the modes that last years' firmware had; it now relies on runtime to implement all of its behavior. This can be changed if this is an issue.


nikitakit (changes requested):
The changes here actually LGTM.

Sad to see the special on-boot blinking mode go, but I understand that it's non-critical.

You need to also update `hibikedevices.json`, though.

karthik-shanmugam:
The only change to make in the json file would be removing the mode parameter right? It might be better to leave that in there even though it's not implemented for now; it makes it easier to implement it later and makes it more understandable as to why the TeamFlag's parameters start at 1 instead of 0

hibike/devices/TeamFlag/team_flag.h
```diff
@@ -7,49 +7,43 @@
...
-#define STAT2 A2
-#define STAT3 A3
+#define LED1 A3
+#define LED2 A2
+#define LED3 A1
+#define LED4 A0
 
 #define LED_PIN 13
+
+#define PARAM_BLUE 1
```
nikitakit: Add comment here about param 0

nikitakit (changes requested):
Oh I see, I actually missed that the params don't start at zero.

Could you make a TODO comment in the code, and also make sure that accidental reads/writes to the value don't cause unrecoverable errors for either the firmware or the python side?

hibike/devices/TeamFlag/team_flag.cpp
```diff
@@ -16,168 +11,110 @@ void setup() {
...
+      break;
+    case PARAM_LED4:
+      data[0] = getLed(5);
+      return sizeof(uint8_t);
+      break;
+    default:
+      break;
   }
+  return 0;
 }
```
yizow: newline at end of file

yizow (changes requested):
Minor style.

Though, i'm curious why the .json file was removed. I wasn't aware hibike had any architecture changes; doesn't it still need that file? 

Double check pinout

hibike/devices/TeamFlag/team_flag.cpp
karthik-shanmugam: There's just one json file in the hibike folder for all the devices now

nikitakit (approved):

yizow (approved):

PR #149, opened 2017-02-23
[Networking] Added setup script for routers

Alramech:
- Requires selenium and gecko driver

- Renames router to Team{team#}


nikitakit (changes requested):
I would recommend adding a docstring at the top of the file specifying what manufacturer and model router this script targets, and what the intended target configuration for the routers is.

This would be useful in the future, in case people wonder whether the script still applies to whatever a future kit configuration might be.

networking/eltnam/setup_routers.py
```diff
@@ -0,0 +1,33 @@
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+import sys
+import time
+
+teamNum = sys.argv[1]
```
nikitakit: I think the script should print some usage information if you don't pass the right number of args, or if arg 1 isn't a valid integer (this case would also make `-h` or `--help` work correctly)

nikitakit (changes requested):

networking/eltnam/setup_routers.py
yizow: seconded

networking/eltnam/setup_routers.py
```diff
@@ -0,0 +1,33 @@
...
+
+wd = webdriver.Firefox()
```
yizow: what happens if no firefox? or no internet access? whole script errors? 

If this is part of a larger devops script, how are they handling that? Are they aware of possible outputs, failure states this might generate? 

It's gonna be annoying if they spend 3 hours flashing and running scripts on everything, then realize they have to do it again cause a subscript somewhere broke the whole thing.

yizow (commented):
More philosophocal comments than anything. 

mostly LGTM

networking/eltnam/setup_routers.py
Alramech: This script is supposed to be run on a dev computer, not the Beaglebones. Firefox is required, but assuming that most of them are running some Ubuntu derivative, it should be there by default. 

No internet is expected since they need to be connected to the router already.

networking/eltnam/setup_routers.py
```diff
@@ -0,0 +1,64 @@
...
+connect to router over wifi
+run script with team number
+"""
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+import time
+import argparse
+
+parser = argparse.ArgumentParser(description=
+                                 "Sets up team routers by changing ssid to be Team{teamNum}. \
```
nikitakit: Please use triple-quoted string instead of `\`

nikitakit (changes requested):

nikitakit (approved):
LGTM, please make sure to squash the commits together when merging (and to write a good commit message as you do that)

Issue #95, opened 2017-01-12
[RUNTIME] changing modes doesnt kill old process


yizow:
https://github.com/pioneers/PieCentral/blob/master/runtime/testy/runtime.py#L75

nikitakit:
My bad for not catching this earlier, during code review.

nikitakit:
This should be fixed now.

PR #150, opened 2017-02-23
[RUNTIME] Runtime/student code upload setup

brandonxxlee:
When Dawn is about to upload student code, we make sure the studentCode process is not running.


runtime/testy/fake_dawn.py
```diff
@@ -50,9 +50,9 @@ def tcp_relay(port):
     s.bind((host, port))
     s.listen(1)
     conn, addr = s.accept()
+    conn.send(msg)
```
nikitakit, line 53: What's up with this change?

It looks like it would make quite a big difference on the functionality of fake_dawn, but I fail to see how it relates to student code upload.

runtime/testy/runtime.py
```diff
@@ -91,17 +91,18 @@ def nonTestModePrint(*args):
           controlState = "auto"
           continue
         elif newBadThing.event == BAD_EVENTS.ENTER_IDLE and controlState != "idle":
+          controlState = "idle"
```
nikitakit, line 94: Has issue #95 been addressed yet?

runtime/testy/runtime.py
```diff
@@ -91,17 +91,18 @@ def nonTestModePrint(*args):
...
           break
         print(newBadThing.event)
         nonTestModePrint(newBadThing.data)
         if newBadThing.event in restartEvents:
+          restartCount += 1
           if (not emergency_stopped and newBadThing.event is BAD_EVENTS.EMERGENCY_STOP):
             emergency_stopped = True #somehow kill student code using other method? right now just restarting on e-stop
           break
```
nikitakit: Given that you added line 94 `controlState = "idle"`, I think you should also add it here and remove it from line 105.

runtime/testy/runtime.py
```diff
@@ -91,17 +91,18 @@ def nonTestModePrint(*args):
...
       stateQueue.put([SM_COMMANDS.RESET, []])
       terminate_process(PROCESS_NAMES.STUDENT_CODE)
       controlState = "idle"
```
nikitakit: It looks like you're moving all control state changes into the while loop, so I think you should move this one too (and maybe add an assert if the controlstate must always be idle at this point).

runtime/testy/runtime.py
```diff
@@ -155,7 +156,8 @@ def checkTimedOut(func, *args):
       time.sleep(max(nextCall - time.time(), 0))
       execCount += 1
 
-    badThingsQueue.put(BadThing(sys.exc_info(), "Process Ended", event=BAD_EVENTS.END_EVENT))
+    if not terminated:
```
nikitakit, line 159: Note that this section of code is also touched by #113, which has basically passed code review. I think you guys should make an effort to get all changes merged into master asap, instead of dragging out the merge conflicts for a long time.

runtime/testy/stateManager.py
```diff
@@ -143,7 +143,7 @@ def send_addr(self, process_name):
     self.processMapping[process_name].send(self.state["dawn_addr"][0])
 
   def student_upload(self):
-    #TODO Implement Student Upload Confirmation
+    self.badThingsQueue.put(BadThing(sys.exc_info(), None, BAD_EVENTS.ENTER_IDLE, False))
     self.processMapping[PROCESS_NAMES.TCP_PROCESS].send([ANSIBLE_COMMANDS.STUDENT_UPLOAD, True])
```
nikitakit, line 147: Would there ever be a situation where you write to ansible with `ANSIBLE_COMMANDS.STUDENT_UPLOAD`, but set the data bit to `False` rather than `True`?

nikitakit (changes requested):
A few minor suggestions in comments, but mostly looks good.

runtime/testy/fake_dawn.py
bodiddles: The student upload message tells runtime to kill any running student code process as a safety measure so Dawn can scp over a new version of student code for us to run. We moved it out of the while loop so it only runs once as a command to kill any running student code which also mirrors actual dawn usage of this command.

runtime/testy/runtime.py
brandonxxlee: We will address this in #153 to properly kill studentCode Process when swapping modes other than IDLE. (IDLE here should kill the student code.)

runtime/testy/stateManager.py
brandonxxlee: Theoretically we would send False if we are not ready for an upload, but as far as I can tell as long as we kill student code we should always be ready, so I guess this will just always be `True`.

runtime/testy/runtime.py
brandonxxlee: Could you specify what you mean here? I'm not entirely sure what I'd be moving, or what the assert would be for.

nikitakit: The line 94 you added looks half-redundant with this. Having both is confusing, but you do need to handle other cases.

brandonxxlee: I have moved the `controlstate = "idle"` into both parts inside the while loop. Are there more things I should add/move?

runtime/testy/fake_dawn.py
nikitakit: sounds good

nikitakit (approved):
lgtm

yizow (approved):
LGTM

Make sure before you click the merge button to run all tests. Hate to break master, and it's happened a number of times already.

PR #151, opened 2017-02-23
[RUNTIME] Update TCP notification proto to include sensor mappings

brandonxxlee:
These mappings will be read from a csv file on the robot and sent to Dawn on initial connection of TCP.


yizow (approved):
LGTM, assuming you change the ansible tag to runtime for consistency.

PR #152, opened 2017-02-24
[DEVOPS] Makefile: Use .PHONY targets, simplify commands

kevinji:
Just in case we have files called `install` or `artifacts`.


nikitakit (approved):
Oh right, I completely forgot to mark targets as phony.

Travis passed, so lgtm

PR #154, opened 2017-02-24
[RUNTIME] Send sensor mappings to Dawn

bodiddles:
-Parses CSV file currently in the runtime repo and sends it via the notification proto to Dawn once Runtime connects to Dawn. 
-Sends each individual mapping as a repeated message in notification.proto with uid/old name as device_name and the name the student set as new_device_name.
-Compiled new notification proto


runtime/testy/Ansible.py
```diff
@@ -251,6 +252,16 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
         self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.sock.connect((self.dawn_ip, TCP_PORT))
 
+        proto_message = notification_pb2.Notification()
+        proto_message.header = notification_pb2.Notification.SENSOR_MAPPING
+        with open('namedPeripherals.csv', 'r') as f:
+            sensorMappings = csv.reader(f)
+            for row in sensorMappings:
+                pair = proto_message.sensor_mapping.add()
+                pair.device_new_name = row[0]
```
nikitakit: Uhh so the sensor mapping should relate some hibike-related uids or device ids to human-readable names.

The use of `device_new_name` and `device_name` as the names for the proto fields is highly misleading if this is the case.

nikitakit (changes requested):
See comment.

Also, what is the plan for receiving updated sensor names from Dawn?

nikitakit:
Don't forget to change and commit the `.proto` file during the rename!

brandonxxlee:
Compiling the proto right now!

Our current plan for updating sensor mapping is Dawn will scp a new csv file everytime they upload student code.

runtime/testy/fake_dawn.py
```diff
@@ -50,19 +50,18 @@ def tcp_relay(port):
...
         receive_msg, addr = conn.recvfrom(2048)
         if receive_msg is None:
             continue
         else:
             parser = notification_pb2.Notification()
             parser.ParseFromString(receive_msg)
-            print(parser.header)
-            if parser.header == notification_pb2.Notification.CONSOLE_LOGGING:
-                print(parser.console_output)
-        next_call += 1.0/dawn_hz
```
nikitakit, line 65: I think you don't want to delete the `next_call +=` line

nikitakit (changes requested):

runtime/testy/fake_dawn.py
```diff
@@ -50,19 +50,19 @@ def tcp_relay(port):
     s.bind((host, port))
     s.listen(1)
     conn, addr = s.accept()
+    conn.send(msg)
     while True:
-        next_call = time.time()
-        conn.send(msg)
         receive_msg, addr = conn.recvfrom(2048)
+        next_call = time.time()
```
yizow: i agree with moving these two `next_call` lines together, but should probably swap back up to be immediately after while True. 
Unneccessary diff.

nikitakit (approved):
lgtm

PR #157, opened 2017-02-25
[Hibike] Testing: New Hibike_Communicator Library

ivanache:
I have coded and successfully tested the Hibike_Communicator, a library which allows the user to act as the BBB and send and receive messages to and from devices by creating a Hibike_Communicator class and calling its functions.


hibike/hibike_communicator.py
```diff
@@ -0,0 +1,218 @@
...
+
+   def process_output(self):
+      """
+         Prints out messages from the devices that are uploaded by newProcess to stateQueue
+         If it's a subscription response from a device whose uid is not in self.uids, the uid will be added to self.uids
+         If it's a device disconnection from a device whose uid in self.uids, the uid will be removed from self.uids
+      """
+      while True:
+         output = self.stateQueue.get()
+         print(output)
```
karthik-shanmugam: I don't think we want this library to automatically print things

karthik-shanmugam (commented):
Is it possible to remove the changes to `hibike_process.py` from this? They're irrelevant now and it's causing merge conflicts. I'd like to get this merged in.

hibike/hibike_communicator.py
ivanache: In that case, should I create another function that prints it, or can I just leave this printing functionality out of the library?

tingtingchiang (changes requested):
Confused about old_hibike_communicator_test.py, is it deprecated or is it examples of using HibikeCommunicator to test? Can we change either the name of the script or the comment? Otherwise it looks good to me.

ivanache:
It's a deprecated testing program for hibike_communicator, made before I
gave hibike_comunicator its own built-in tester.

tingtingchiang:
Do you think we'll need it or can we get by with just hibike_communicator.py? It could get confusing if we merge an "old" file if we don't need it later on.

ivanache:
I don't think we'll need it.

hibike/hibike_communicator.py
```diff
@@ -0,0 +1,218 @@
...
+import hibike_process as hp
+import multiprocessing
+import threading
+import time
+
+class HibikeCommunicator:
+  
+   def __init__(self):
+      """
+          Set up the pipes for communication between the device and the Beegle Bum Black, creates a thread to recieve communications from the device, starts up the process that runs the communication, and sets up a dictionary that caches all device values sent to it.
```
tingtingchiang: BeagleBone Black, in general though, please read through the comments and make sure that everything is spelled correctly.

hibike/hibike_communicator.py
```diff
@@ -0,0 +1,218 @@
...
+           uid - the device's uid
+           delay - the delay between device datas to be sent
+           params - an iterable of the names of the params to be subscribed to
+        """
+        self.pipeToChild.send(["subscribe_device", [uid, delay, params]])
+
+##############
+#  TESTING   #
+##############
+def device_comms(comms, uid_type_tuple):
```
tingtingchiang: Let's separate the two actions in each if block: Sending commands and printing the most current data of all live sensors each at regular (different) intervals. I recommend just copying from hibike_process's main function because it's nice and clean which == easier to modify.

tingtingchiang: I will push my most recent edits of hibike/testing soon, might be useful as reference when switching to hibike_process's methods.

tingtingchiang (changes requested):
According to Sam, ideal end product would be: easy to edit and send commands + print to console at regular intervals.

Also, we should definitely check how hotplugging interacts with this script.

tingtingchiang (commented):
LGTM, let's test it and then we can merge

hibike/hibike_communicator.py
```diff
@@ -88,10 +89,14 @@ def process_output(self):
...
          Returns a tuple of the value and the timestamp of the last device_values package recieved from a uid and a parameter
-         Precondition: a device_data must have been recieved from the param before calling this function
+         Precondition: a device_data with a UID, params, and values must have been recieved from the param before calling this function
       """
-      print("Where the error is coming from: ", self.device_values_cache)
-      return self.device_values_cache[uid][param]
+      #print("Where the error is coming from: ", self.device_values_cache)
+      try:
+         return self.device_values_cache[uid][param]
+      except KeyError:
```
kevinji: It'd probably be nicer to use `dict.get(key, default)` here.

hibike/hibike_communicator.py
```diff
@@ -61,24 +61,25 @@ def process_output(self):
...
-                print("No UIDs detected for a device_values request")
+            except KeyError:
+                #print("No UIDs detected for a device_values request")
                 params_values = [{}]
              
-             for param_val_tuple in params_values:
+            for param_val_tuple in params_values:
                 try:
                     parameter = param_val_tuple[0]
                     value = param_val_tuple[1]
```
kevinji: This can just be `parameter, value = param_val_tuple`

hibike/hibike_communicator.py
```diff
@@ -61,24 +61,25 @@ def process_output(self):
...
+                #break
+                
+            #print("UID: ", uid, "\n")
+            try:
                 params_values = output[1][0][uid]
-             except IndexError:
-                print("No UIDs detected for a device_values request")
+            except KeyError:
+                #print("No UIDs detected for a device_values request")
                 params_values = [{}]
```
kevinji: You can probably condense this entire section to `params_values = next(output[1][0].values(), [{}])`.

kevinji (changes requested):
Left some various style comments.

baby-bell:
I'm afraid that this work has become stale. We are working on it in another branch.

PR #158, opened 2017-02-26
[HIBIKE] Mac-compatible version of hibike_process

ivanache:
Previously, when I tried to use hibike_process on a MacBook Air that uses OS X Sierra, the program failed to find any USB ports that were plugged in. I made a small fix to hibike_process so that this would not be the case.


ivanache:
Made my code cleaner by using an if-else statement instead of hibike_process_osx_sierra

vdonato (approved):

hibike/hibike_process.py
```diff
@@ -6,15 +6,22 @@
...
 uid_to_index = {}
 
 def hibike_process(badThingsQueue, stateQueue, pipeFromChild):
-
-    ports = glob.glob("/dev/ttyACM*") + glob.glob("/dev/ttyUSB*")
+    
+    # So that it's compatible with OS X Sierra
+    # Note: If you are running OS X Sierra, do not access the directory through vagrant ssh
+    # Instead access it through Volumes/vagrant/PieCentral
+    if system() == "Darwin":
```
nikitakit: I don't remember what pre-Sierra macOS uses for USB device paths. Is it possible that it has either `ttyACM` or `ttyUSB` paths? Because in that case, it may make sense to just combine all the globs on one line instead of doing the `if` statement.

vdonato: I think that all reasonably modern versions of OSX/macOS all use tty.usbmodem*, but yeah I don't really see a reason not to just combine all of the globs onto one line.

PR #159, opened 2017-02-26
[RUNTIME] Send Sensors to Dawn

brandonxxlee:
Reformatted sending to accurately read SM to send sensors to dawn.


ansible-protos/runtime.proto
```diff
@@ -1,20 +1,26 @@
...
+        float value = 2;
+    }
+    message SensorData {
+        reserved "value";
+        reserved 3;
+        string device_type = 1;
+        string device_name = 2;
+        uint64 uid = 4;
+        uint32 int_device_type = 5;
+        repeated ParamValue param_value = 6;
```
nikitakit, line 25: How will we handle non-float-valued params?

One possibility is to later modify ParamValue to instead be:
```
string param = 1;
oneof kind {
    float float_value = 2;
    int int_value = 3
}
```

If that is the eventual plan, then it may make sense to name the field `float_value` to facilitate this eventual transition.

runtime/testy/runtimeUtil.py
```diff
@@ -54,6 +54,13 @@ class ANSIBLE_COMMANDS(Enum):
   CONSOLE        = "console"
 
 @unique
+class STUDENT_CODE_STATE(Enum):
+  IDLE = 2
```
nikitakit: So these are out of order and don't start at zero. Could you explain why that is?

(If you're just matching the protobuf, then you should import these values from `runtime_pb2`)

nikitakit (dismissed):
see comments

ansible-protos/runtime.proto
brandonxxlee: Changed proto to include oneof so that can in future support different data types.

nikitakit (changes requested):
I suspect the compiled protos in this PR have not been updated.

yizow (approved):
lgtm

nikitakit (approved):
lgtm

PR #162, opened 2017-02-26
[Dawn] Initial Implementation of Notifications; Faster Rendering Rates

kevinmasd:
Initial Implementation of Notifications; untested with real robot so far.

Info-Actions and Ansible modified to reduce state dispatch calls to reduce rendering rates. Editor still slow, and future speed changes will come in future improvements.


dawn/.eslintrc
```diff
@@ -33,6 +33,8 @@
     "react/no-unused-prop-types": 0,
     "no-useless-escape": 0,
     "import/prefer-default-export": 0,
-    "import/no-extraneous-dependencies": 0
+    "import/no-extraneous-dependencies": 0,
+    "no-empty": 0,
```
nikitakit: Why do you need to disable these linter rules?

dawn/main/ansible/Ansible.js
```diff
@@ -1,35 +1,103 @@
...
+const tcp = net.createServer((sock) => {
+  console.log('Runtime Connected');
+  sock.on('end', () => {
+    console.log('Runtime Disconnected');
+  });
+  sock.on('data', (data) => {
+    received = true;
+    const decoded = Notification.decode(data);
+    if (decoded.header === Notification.Type.STUDENT_RECEIVED) {
+      RendererBridge.reduxDispatch(notifyChange(0));
```
nikitakit: What is the `0` here?

Would this be better as an enum?

dawn/main/ansible/Ansible.js
```diff
@@ -1,35 +1,103 @@
...
+  console.log(event);
+  const message = new Notification({
+    header: Notification.Type.STUDENT_SENT,
+    console_output: '',
+  });
+  runtimeSocket.write(message.encode().toBuffer(), () => {
+    console.log('Runtime Notified');
+  });
+  received = false;
+  RendererBridge.reduxDispatch(notifyChange(1));
```
nikitakit: What does `1` here signify?

dawn/main/ansible/Ansible.js
```diff
@@ -1,35 +1,103 @@
...
+    });
+    count += 1;
+    const timestamp = Date.now();
+    while (Date.now() - timestamp < 1000) {
+
+    }
+  }
+  if (!received) {
+    console.error('No Confirmation');
+    RendererBridge.reduxDispatch(notifyChange(2));
```
nikitakit: What does `2` here signify?

dawn/renderer/components/PeripheralList.js
```diff
@@ -1,29 +1,86 @@
 /**
- * A generic component for displaying a list of Peripherals.
+ * Haxors
```
nikitakit: ???

nikitakit (commented):
Thanks for fixing the lag!

I tested this with fake runtime, and it can now handle 10Hz (but not 20Hz). My only thought about this change was that it might be good to lower `SENDRATE` in fake runtime to 100 or lower, to make sure that any performance regressions are caught as part of Dawn development.

This PR also contains protocol changes, so I would like a second reviewer to look at that part of the code.

dawn/.eslintrc
kevinmasd: Allow empty loops to act as blocking mechanisms. (See Editor's SCP handling)

dawn/renderer/components/PeripheralList.js
kevinmasd: Renamed FCPL to PeripheralList, so this was carried over.

dawn/main/ansible/Ansible.js
kevinmasd: I'll enum it. It's supposed to match the protobuf numbering.

dawn/renderer/components/Editor.js
```diff
@@ -119,6 +120,10 @@ class Editor extends React.Component {
     }, { passive: true });
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextProps !== this.props || nextState !== this.state;
```
nikitakit: Oh I see, this is how you tell react to not re-render a component despite a changes taking place in the state/props.

Would implementing this for the peripheral components as well provide any speed boost?

Some quick searching also revealed the existence of [React.PureComponent](https://facebook.github.io/react/docs/react-api.html#react.purecomponent). Would it be possible to use that as the base class for all (or nearly all) components instead, to get a speed boost? This will make it so you don't have to manually write a `shouldComponentUpdate` for each component, but it does require purity.

dawn/main/ansible/Ansible.js
nikitakit: Oh, in that case is it possible to pull the enum directly out of the `.proto` file, instead of replicating the numbers?

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -33,11 +34,13 @@ let state = stateEnum.IDLE;
 server.on('message', (msg) => {
   // Decode and get the raw object.
   const data = DawnData.decode(msg).toRaw();
-  console.log(data);
-  if (data.student_code_status !== 'IDLE') {
-    state = stateEnum.RUNNING;
-  } else {
+  // console.log(data);
```
awli: Probably want to drop the comment if you're not going to log this at all.

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -33,11 +34,13 @@ let state = stateEnum.IDLE;
...
+  if (data.student_code_status === 'IDLE') {
     state = stateEnum.IDLE;
+  } else if (data.student_code_status === 'ESTOP') {
+    state = stateEnum.ESTOP;
+  } else {
+    state = stateEnum.RUNNING;
```
awli: consider a switch statement? Your preference though.

dawn/main/ansible/Ansible.js
```diff
@@ -1,35 +1,103 @@
...
 
 const runtimeBuilder = ProtoBuf.loadProtoFile(`${__dirname}/runtime.proto`);
 const RuntimeData = runtimeBuilder.build('RuntimeData');
+const notificationBuilder = ProtoBuf.loadProtoFile(`${__dirname}/notification.proto`);
+const Notification = notificationBuilder.build('Notification');
+
+let runtimeSocket = null;
+let received = false;
+
+if (!Date.now) {
```
awli: Are you ever running in an environment where you don't have Date.now?

dawn/main/ansible/Ansible.js
```diff
@@ -1,35 +1,103 @@
...
+
 
 /**
  * Serialize the data using protocol buffers.
  */
 function buildProto(data) {
-  const status = data.studentCodeStatus ?
-    StudentCodeStatus.TELEOP : StudentCodeStatus.IDLE;
+  let status = null;
+  if (data.studentCodeStatus === 1) {
```
awli: Magic numbers...

dawn/main/ansible/Ansible.js
kevinmasd: I'd like to keep a barrier between main/renderer components.

nikitakit: That's fine. Though if it's possible to also import a second copy of the proto file here, then maybe you can do that.

kevinmasd: Actually, thought it was something different. This enum deals with the protocol for notifications (i.e. Dawn -> Runtime Call, Runtime -> Dawn Replay or timeout)

dawn/renderer/components/Editor.js
```diff
@@ -182,6 +187,19 @@ class Editor extends React.Component {
       });
       throw err;
     });
+    ipcRenderer.send('NOTIFY_UPLOAD');
+    while (this.notificationHold === 1) {
```
awli: Is this intentionally blocking the thread until notificationHold is not 1? Consider using a Promise or something else instead.

dawn/renderer/components/PeripheralList.js
awli: I can't believe FCPL lasted that long...

dawn/main/ansible/Ansible.js
nikitakit: Oh I see

dawn/renderer/components/Editor.js
kevinmasd: I'll look into Promises later; blocking is intentional.

dawn/renderer/components/Editor.js
kevinmasd: I can try to check if this switch causes better improvements later, during state revision.

nikitakit:
Please make sure to open a GitHub issue about the busy-waiting loop before merging this. (Or even better, fix the busy-waiting. But I understand that getting a driving robot takes priority)

dawn/fake-runtime/FakeRuntime.js
```diff
@@ -33,11 +34,15 @@ let state = stateEnum.IDLE;
...
-  } else {
-    state = stateEnum.IDLE;
+  switch (data.student_code_status) {
+    case 'IDLE':
+      state = stateEnum.IDLE;
+      break;
+    case 'ESTOP':
+      state = stateEnum.ESTOP;
+      break;
+    default:
```
yizow: I think another case is 'auto' for autonomous code. look at the dawn/runtime spec for details.

dawn/.eslintrc
kevinji: Wouldn't it make more sense to just disable the rule on a case-by-case basis instead of for the entire project? These rules are kinda useful to keep enabled.

dawn/renderer/components/Editor.js
```diff
@@ -136,93 +139,105 @@ class Editor extends React.Component {
...
-            type: 'warning',
-            buttons: ['Close'],
-            title: 'Connection Issue',
-            message: 'Could Not Connect to Robot',
-          });
-          throw err;
+    ipcRenderer.send('NOTIFY_UPLOAD');
+    const waiting = () => {
+      let count = 0;
+      const waitPromise = (resolve, reject) => {
```
yizow, line 183: Move this to a saga

dawn/renderer/components/GamepadItem.js
```diff
@@ -127,7 +127,7 @@ class GamepadItem extends React.Component {
 
 GamepadItem.propTypes = {
   gamepad: React.PropTypes.object,
-  index: React.PropTypes.number,
+  index: React.PropTypes.string,
```
yizow: revert and fix elsewhere.

dawn/renderer/components/UpdateBox.js
```diff
@@ -51,28 +51,47 @@ class UpdateBox extends React.Component {
...
+              this.setState({ isUploading: false });
+              this.props.hide();
               dialog.showMessageBox({
                 type: 'warning',
                 buttons: ['Close'],
                 title: 'Upload Issue',
                 message: 'Update File Upload Failed.',
               });
               throw err2;
+            } else {
```
yizow, line 64: create dispatch and move this to saga.

dawn/renderer/reducers/info.js
```diff
@@ -1,27 +1,36 @@
...
 
 const info = (state = initialInfoState, action) => {
   switch (action.type) {
-    case 'ANSIBLE_CONNECT':
+    case 'PER_MESSAGE':
       return {
         ...state,
         connectionStatus: true,
+        robotState: action.robotState,
+        isRunningCode: (action.robotState === 1 ||
```
yizow: make enum

dawn/renderer/reducers/peripherals.js
```diff
@@ -1,16 +1,35 @@
...
       return nextState;
-    case 'PERIPHERAL_DISCONNECT':
+    }
+    case 'PERIPHERAL_DISCONNECT': {
       delete nextState[action.id];
       return nextState;
-    default:
+    }
+    case 'PERIPHERAL_RENAME': {
+      nextState[action.id] = state[action.id];
```
yizow: nop

yizow (changes requested):

yizow (approved):

PR #163, opened 2017-02-28
[RUNTIME] stateManager: add hibike integration functions

yizow:
updated hibikeResponseDeviceSubbed
add hibikeResponseDeviceValues
adds hibikeEnumerateDevices and has Robot call it automatically
Add hibikeWriteValue

When receiving data from a device that hasn't enumerated yet, hibike
automatically responds with a sub delay 0 message, effectively telling
it to shut up.

Added new test. Tests that it is possible to read hibike devices.
Requires the use of hibike/spawn_virtual_devices.py

Removed some tests from the default runtime test suite that interact
with hibike. Due to how virtual devices work, these tests need to be
tested in isolation, manually. The virtual devices continue broadcasting
even after runtime has restarted.

yogibear code lives in hibike/yogibear
hibike/hibikeDevices.json file updated from that branch

This commit drives motors on the robot.
Done in a very hacky way by running a runtime.py studentCode "test"
Namely, hibikeRead. The setup() and main() functions also do this, but
require a start signal from Dawn unless the code is modified.

new main and setup method for testing with dawn


hibike/hibikeDevices.json
```diff
@@ -67,6 +61,26 @@
                                                      ]
     },
 
+
+    {"id": 10,     "name": "YogiBear",      "params": [
```
nikitakit: If @karthik-shanmugam thinks the YogiBear PR is ready for merging, I'd rather merge that first than include this diff here.

runtime/testy/runtime.py
```diff
@@ -322,9 +322,18 @@ def startHibike(badThingsQueue, stateQueue, pipe):
   # badThingsQueue - queue to runtime
   # stateQueue - queue to stateManager
   # pipe - pipe from statemanager
+  def addPaths():
+    """Modify sys.path so we can find hibike.
+    """
+    path = os.path.dirname(os.path.abspath(__file__))
+    parent_path = path.rstrip("runtime/testy")
+    hibike = os.path.join(parent_path, "hibike")
+    sys.path.insert(1, hibike)
```
nikitakit, line 331: Why 1?

I'd prefer `append`, or `insert(0, ...` if need be.

runtime/testy/runtime.py
```diff
@@ -322,9 +322,18 @@ def startHibike(badThingsQueue, stateQueue, pipe):
   # badThingsQueue - queue to runtime
   # stateQueue - queue to stateManager
   # pipe - pipe from statemanager
+  def addPaths():
+    """Modify sys.path so we can find hibike.
+    """
+    path = os.path.dirname(os.path.abspath(__file__))
+    parent_path = path.rstrip("runtime/testy")
```
nikitakit, line 329: Your call, but maybe `os.path.join(path, '../../hibike')` instead of this line and the next?

runtime/testy/studentCode.py
```diff
@@ -102,13 +113,11 @@ def createKey_main():
   restarts = Robot.getValue("Restarts")
   Robot.setValue(restarts+1, "Restarts")
 
-def hibikeSubscribeDevice_setup():
-  pass
+# def hibikeSubscribeDevice_setup():
```
nikitakit, line 116: I would prefer you rename these to end in `_setup_DISABLED`/`_main_DISABLED` rather than commenting out (so at least they get syntax checked). Or hide them behind an `if/else`. Or delete them entirely.

runtime/testy/studentCode.py
```diff
@@ -166,11 +175,36 @@ def emergencyStop_main():
...
 
-def hibikeSensorMappings_setup():
-  pass
-
-def hibikeSensorMappings_main():
-  print(Robot._hibikeGetUID('zero'))
-  print(Robot._hibikeGetUID('one'))
-  print(Robot._hibikeGetUID('two'))
-  print(Robot._hibikeGetUID('three'))
+# Due to interaction with hibike virtual devices, currently
```
nikitakit, line 178: Would it be possible for you to modify the `test` target in the runtime Makefile to programmatically execute these tests? (You can add calls to `spawn_virtual_devices.py` in the makefile)

runtime/testy/studentCode.py
nikitakit: Another option is to define some mechanism (like a `@manual` decorator) to annotate tests such that they run with `-t <testname>` but not with plain `-t`.

nikitakit (changes requested):
See comments

nikitakit:
@yizow You should be able to use something like the optional tests framework I proposed in #165 to run the virtual-device tests as part of travis CI.

One way to spawn virtual devices in the makefile is to add a line to the `test` target:
```bash
cd ../hibike; python3 spawn_virtual_devices.py & cd ../runtime/testy; python3 runtime.py --test optional_myHibikeTest; kill -9 %1
```

This needs to be all on one line, substituting test name as appropriate. There's probably a cleaner way to do this, but this is what I can come up with at the moment.

You may need to modify the signal kill sends (maybe don't do `-9`) and possibly add a `sudo apt-get install socat` to the `install` target.

hibike/hibikeDevices.json
nikitakit: @yizow YogiBear PR is now merged. You should be able to rebase this branch to get the json changes, instead of making them yourself.

yizow:
Closing because these changes have been merged into #179 

@nikitakit re: comment about `sys.path.insert(1`
http://stackoverflow.com/questions/10095037/why-use-sys-path-appendpath-instead-of-sys-path-insert1-path

PR #168, opened 2017-02-28
[RUNTIME] Allows Dawn to reconnect to the robot

### LINKED ISSUE #140: [RUNTIME] Disconnecting Dawn crashes TCP sender
Disconnecting Dawn from Runtime crashes the TCP send thread in the TCP process as it throws an error due to the now broken pipe.

bodiddles:
Fixes #140 by killing all Ansible processes and studentCode if Dawn disconnects and resets UDP Recv to allow for a new Dawn connection. 

Also fixes bug where killing studentCode process would leave it alive and cause runtime to kill with os.kill and end the runtime process. Made the while loop calling the function for running studentCode check if the terminated flag is true.


runtime/testy/Ansible.py
```diff
@@ -258,13 +258,12 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
         sendName = THREAD_NAMES.TCP_SENDER
         recvName = THREAD_NAMES.TCP_RECEIVER
         super().__init__(sendName, TCPClass.sender, recvName, TCPClass.receiver, badThingsQueue, stateQueue, pipe)
-
```
yizow: why delete this line?

runtime/testy/Ansible.py
```diff
@@ -258,13 +258,12 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
...
         stateQueue.put([SM_COMMANDS.SEND_ADDR, [PROCESS_NAMES.TCP_PROCESS]])
-        self.dawn_ip = pipe.recv()[0]
-
+        from_pipe = pipe.recv()
```
yizow: why make this change?

runtime/testy/Ansible.py
```diff
@@ -258,13 +258,12 @@ def __init__(self, badThingsQueue, stateQueue, pipe):
...
+        self.dawn_ip = from_pipe[0]
         self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         self.sock.connect((self.dawn_ip, TCP_PORT))
-
```
yizow: why delete this line

runtime/testy/Ansible.py
```diff
@@ -330,6 +329,12 @@ def unpackage(data):
         try:
             while True:
                 recv_data, addr = self.sock.recvfrom(2048)
+                if recv_data == b'':
```
yizow, line 334: no timeout? what if we drop a connection for a few milliseconds?

yizow: also, this is tcp. when would we ever receive an empty tcp packet?

runtime/testy/stateManager.py
```diff
@@ -40,7 +40,7 @@ def makeCommandMap(self):
       SM_COMMANDS.SEND_ADDR: self.send_addr,
       SM_COMMANDS.ENTER_IDLE: self.enter_idle,
       SM_COMMANDS.ENTER_TELEOP: self.enter_teleop,
-      SM_COMMANDS.ENTER_AUTO: self.enter_auto
+      SM_COMMANDS.ENTER_AUTO: self.enter_auto,
```
yizow: I thought commas at the end like this broke things?

yizow (changes requested):

runtime/testy/Ansible.py
bodiddles: timeout should be included in the socket itself. When it returns an empty byte string, it indicates that the socket on the other end has closed so in this case, we would want to actually restart all ansible processes

yizow (approved):
lgtm

nikitakit:
I see a failing `hibikeSubscribeDevice` on one of the two travis runs. Is it possible that this flaky tests points to some underlying problem?

nikitakit:
Actual code looks good here.

However, we should make a note to revisit the assumption that the correct way to handle TCP disconnects is to restart the UDP server and also to enter idle mode. This may not be the right thing to do on the field (especially if we have flaky wifi that sometimes drops packets). It may be the case that on the field, we don't actually want to enter idle mode.

I'm also skeptical that UDP receiver actually needs to be restarted. It may just be sufficient to have it clear its dawn_ip such that is ready for reconnects from anywhere.

nikitakit (approved):
lgtm

PR #172, opened 2017-03-02
[HIBIKE] PDB Firmware

karthik-shanmugam:
This is the PDB firmware


nikitakit:
I'm assuming all the libraries other than SevenSeg were already merged (as part of yogibear PR). Is that correct?

dHutchings (changes requested):
I think that the SevenSeg.pdf file should be removed, and pulled from all history.

nikitakit:
Note: you probably want to delete `battery_buzzer.json`

hibike/devices/BatteryBuzzer/battery_buzzer.cpp
```diff
@@ -1,316 +1,146 @@
...
+
+//got to keep these globals here to keep the compiler happy.
+float v_cell1;  // param 1
+float v_cell2;  // param 2
+float v_cell3;  // param 3
+float v_batt;   // param 4
+float dv_cell2; // param 5
+float dv_cell3; // param 6
+
+bool prints = false; // keep until all Serial methods get deleted
```
nikitakit, line 37: I think both `prints` and any uses of Serial should be deleted, because any stray serial messages will mess up the hibike communications.

If you do want to keep the print statements around for potential future debugging, I suggest you change this to say `#define DEBUG_PRINTS 0` and use `if (DEBUG_PRINTS)` everywhere. The all-caps name will make it clear that this is a compile-time constant, not a variable that might change.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+}
+
+
+
+
+
+
+void print_eeprom() //prints first few eeprom values as text.
```
nikitakit, line 111: Note that this function is never used anywhere.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+    
+    next_addr = next_addr + sizeof(float);
+
+    
+  }
+
+}
+
+
+
```
nikitakit, line 107: Could you normalize the whitespace in this file? There are empty lines in odd places.

hibike/devices/BatteryBuzzer/battery_buzzer.h
```diff
@@ -1,56 +1,32 @@
...
+extern float dv_cell2;
+extern float dv_cell3;
+extern float vref_guess;
+extern bool prints;
 
-// Calibration reference (from external voltage source)
-#define MIN_REF_VOLT 4.5
-#define MAX_REF_VOLT 5.5
-#define REF_VOLT 5
+extern bool calibrated;
```
nikitakit, line 20: I see no uses of this outside `battery_buzzer.cpp`. Does this need to be `extern`?

hibike/devices/BatteryBuzzer/battery_buzzer.h
```diff
@@ -1,56 +1,32 @@
...
+extern bool triple_calibration;
+extern int buzzer;
 
-#define LOW_VOLTAGE_ENTER_THRESHOLD 3.4
-#define LOW_VOLTAGE_EXIT_THRESHOLD 3.9
+extern unsigned long last_print_time;
 
-#define UNBALANCED_ENTER_THRESHOLD 0.7
-#define UNBALANCED_EXIT_THRESHOLD 0.1
+// extern bool unsafe_status
```
nikitakit, line 27: Delete commented out code.

hibike/devices/BatteryBuzzer/safety.h
```diff
@@ -0,0 +1,14 @@
+#ifndef SAFETY_H
+#define SAFETY_H
+extern bool unsafe_status;
+
+#include "eeprom.h"
+#include "battery_buzzer.h"
+#include "pitches.h"
+
+void handle_safety();
+bool is_unsafe();
```
nikitakit, line 10: So it turns out that `is_unsafe` actually does a bunch of computation, instead of being a simple getter.

Could you move the contents of `is_unsafe` into `handle_safety`, and make `is_unsafe` a simple getter instead? This would also eliminate the need for making `unsafe_status` extern.

hibike/devices/BatteryBuzzer/safety.cpp
```diff
@@ -0,0 +1,79 @@
+#include "safety.h"
```
nikitakit, line 1: I think the `#include "pitches.h"` should be moved here (from inside `safety.h`). Possibly do the same for `#include "battery_buzzer.h"`, too.

hibike/devices/BatteryBuzzer/disp_8.cpp
```diff
@@ -0,0 +1,201 @@
...
+SevenSeg disp(8,9,7,2,0,15,14);
+
+const int numOfDigits=4;
+int digitPins[numOfDigits] = {6, 5, 4, 3};
+
+
+unsigned long last_LED_time = 0;  //Time the last LED switched
+int sequence = 0; //used to switch states for the display.  Remember that the hangle_8_segment cannot be blocking.
+
+int segment_8_run = 0;  //0 for the normal voltage readout.  1 for "Clear Calibration".  2 for "New Calibration"
```
nikitakit, line 16: This should be an enum.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+void measure_cells() //measures the battery cells. Should call at least twice a second, but preferrably 5x a second.  No use faster.
+{
+    int r_cell1 = analogRead(cell1);
+    int r_cell2 = analogRead(cell2);
+    int r_cell3 = analogRead(cell3);
+
+    //Serial.println("Raw Voltages");
```
nikitakit, line 52: I don't think you should have huge chunks of commented-out code in this file.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    //Serial.println(5*float(r_cell1)/1024);
+    //Serial.println(5*float(r_cell2)/1024);
+    //Serial.println(5*float(r_cell3)/1024);
+
+
+    v_cell1 = float(r_cell1) * (10 + 10.0) / 10.0 * calib[0] / 1024;  // (10k + 10k)/(10k) to extract cell voltage in counts, then convert 1024 counts per vref volts.
```
nikitakit, line 58: Lots of magic numbers here. I suggest you make constants for the resister values (e.g. `RESISTOR1_KOHM`), and use those here.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    // }
+    last_print_time = millis();
+
+}
+
+void handle_calibration() //called very frequently by loop.
+{
+  if (digitalRead(calib_button)) //I pressed the button, start calibrating.
+    {
+    if(triple_calibration && calib[0] == 2.56 && calib[1] == 2.56 & calib[2] == 2.56)
```
nikitakit, line 103: `2.56` is a magic number used throughout the code, in multiple files... and you're even relying for equality tests to succeed between uses of this constant from different places. Please switch to using a single source of truth via `#define` or similar.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+      delay(1);
+    }
+    delay(100);
+
+
+    //ok.  let's assume (for now) that every line has 5V put at it.
+    //I want to calculate the best-fit for the true vref.
+
+    //expected voltages, after the resistive dividers.
+    float expected_vc1 = 5 * 10.0 / (10.00 + 10.00);
```
nikitakit, line 167: More uses of magic numbers. Resistor values should have a single source of truth.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    float expected_vc3 = 5 * 10.0 / (51.00 + 10.0);
+
+
+    int r_cell1 = analogRead(cell1);
+    int r_cell2 = analogRead(cell2);
+    int r_cell3 = analogRead(cell3);
+
+
+    //as per whiteboard math.  Is it 1023 or 1024?  I think it's 1023.
+    float vref1 = 1024 / (float(r_cell1)) * 10.0 / (10.0 + 10.00) * 5;
```
nikitakit, line 178: More uses of magic numbers. Resistor values should have a single source of truth.

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    float vref3 = 1024 / (float(r_cell3)) * 10.0 / (51.00 + 10.0) * 5;
+
+    //calibration array values are updated.  These are only used if triple calibration is in use.
+    calib[0] = vref1;
+    calib[1] = vref2;
+    calib[2] = vref3;
+
+
+    //avg all vrefs together to get best guess value    
+    vref_guess = (vref1 + vref2 + vref3) / float(3);
```
nikitakit, line 189: My personal style preference is `3.0f` rather than `float(3)`

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    float expected_vc2 = 5 * 10.0 / (30.00 + 10.00);
+    float expected_vc3 = 5 * 10.0 / (51.00 + 10.0);
+
+
+    int r_cell1 = analogRead(cell1);
+    int r_cell2 = analogRead(cell2);
+    int r_cell3 = analogRead(cell3);
+
+
+    //as per whiteboard math.  Is it 1023 or 1024?  I think it's 1023.
```
nikitakit, line 177: This comment disagrees with the code

hibike/devices/BatteryBuzzer/safety.h
```diff
@@ -0,0 +1,14 @@
...
+void buzz(boolean should_buzz);
+
+
+#endif
```
nikitakit, line 14: Please add newlines at eof (in other files, too)

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+void handle_calibration() //called very frequently by loop.
+{
+  if (digitalRead(calib_button)) //I pressed the button, start calibrating.
+    {
+    if(triple_calibration && calib[0] == 2.56 && calib[1] == 2.56 & calib[2] == 2.56)
+    {
+      //i'm in triple calibration mode, but the calibration array is exactly the same as the datasheet values... which means i haven't been calibrated.
+      float vref_new = calibrate();
+      write_triple_eeprom(calib[0],calib[1],calib[2]);
+      start_8_seg_sequence(2);
```
nikitakit, line 108: Argument looks like a magic number

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+      //i'm in triple calibration mode, but the calibration array is exactly the same as the datasheet values... which means i haven't been calibrated.
+      float vref_new = calibrate();
+      write_triple_eeprom(calib[0],calib[1],calib[2]);
+      start_8_seg_sequence(2); 
+    }
+    else if(!triple_calibration && get_calibration() == -1.0) //i'm not in triple calibration mode, and i haven't been calibrated.
+    {   
+      float vref_new = calibrate();
+      write_single_eeprom(vref_new);
+      start_8_seg_sequence(2);
```
nikitakit, line 114: Argument looks like a magic number

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+        delay(1);
+      }
+      
+      if(prints)
+      {
+        Serial.println("Clear EEPROM");
+      }
+      //reset all my calibration values as well, so my calibration values that i'm using are in lockstep with the EEPROM.
+      
+      vref_guess = 2.56;
```
nikitakit, line 130: Magic numbers (see above)

hibike/hibikeDevices.json
```diff
@@ -28,13 +28,14 @@
...
-                                                          {"number": 1  , "name": "safe"       , "type": "bool"    , "read": true , "write": false },
-                                                          {"number": 2  , "name": "cell1"      , "type": "float"   , "read": true , "write": false },
-                                                          {"number": 3  , "name": "cell2"      , "type": "float"   , "read": true , "write": false },
-                                                          {"number": 4  , "name": "cell3"      , "type": "float"   , "read": true , "write": false },
-                                                          {"number": 5  , "name": "total"      , "type": "float"   , "read": true , "write": false },
-                                                          {"number": 6  , "name": "callibrate" , "type": "bool"    , "read": false, "write": true  }
+                                                          {"number": 0  , "name": "unsafe_status"  , "type": "bool"    , "read": true , "write": false },
```
nikitakit, line 31: To be honest, the name `unsafe_status` is confusing to me, because I'm not sure whether true means safe, or unsafe. The old name "safe" was clearer (and something like `unsafe`, `is_unsafe`, etc. would also be clearer)

hibike/devices/BatteryBuzzer/voltage_sense.cpp
```diff
@@ -0,0 +1,212 @@
...
+    //   }
+  
+    //   Serial.println("ABS: Cell1 : " + str_cell1 + " Cell2 : " + str_cell2 + " Cell3 : " + str_cell3);
+    //   Serial.println("REL: Cell1: " + str_cell1 + " Cell2 : " + sdv_cell2 + " Cell3 : " + sdv_cell3);
+    //   Serial.println("");
+    //   Serial.println("");
+
+
+    // }
+    last_print_time = millis();
```
nikitakit, line 95: Moving this to `battery_buzzer.cpp::loop` would allow making this variable non-extern. Moreover, the entire name `last_print_time` seems not to accurately reflect the usage of this variable.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "eeprom.h"
+
+//low-level functions that deal with storing the calibration values using the agreed upon format.
+
+float get_calibration()  //returns calibration voltage.  Returns -1 (float) if no calibration.
+{
+   if(EEPROM.read(0)=='C' && EEPROM.read(1)=='A' && EEPROM.read(2)=='L' && EEPROM.read(3)==':')
+   {
+      //instantiate a float.
+      float f = 0.00000f;
```
nikitakit, line 11: You don't need so many zeros.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  
+  int next_addr = 0;
+  for(int i = 1; i<=3; i++)
+  {
+     if(EEPROM.read(next_addr)=='C' && EEPROM.read(next_addr+1)=='A' && EEPROM.read(next_addr+2)=='L' && EEPROM.read(next_addr+3) == char(i+48) &&EEPROM.read(next_addr+4)==':')
+     //so i only take a valid calibration, and there's no chance of accidentally getting messed up with partial triple calibration left over in single mode.
+     {
+        //instantiate a float.
+        float f = 0.00000f;
```
nikitakit, line 36: You don't need so many zeros.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+  int next_addr = 0;
+  for(int i = 1; i<=3; i++)
+  {
+    EEPROM.write(next_addr,'C');
+    next_addr++;
+    EEPROM.write(next_addr, 'A');
+    next_addr++;
+    EEPROM.write(next_addr, 'L');
+    next_addr++;
+    EEPROM.write(next_addr, char(i+48));  //plus 48 because ASCII table
```
nikitakit, line 82: I think something like `char('0' + i)` would be cleaner. You can use characters as arguments to addition, instead of hardcoding 48 here -- you just have to make sure the result is correctly cast to char.

hibike/devices/BatteryBuzzer/eeprom.cpp
```diff
@@ -0,0 +1,145 @@
...
+void update_triple_calibration() //updates the global "calib" array based on EEPROM.
+//Calib will contain datasheet values (2.56) if there's no calibration.
+//If there is, the appopriate elements will be updated.
+{
+
+  
+  int next_addr = 0;
+  for(int i = 1; i<=3; i++)
+  {
+     if(EEPROM.read(next_addr)=='C' && EEPROM.read(next_addr+1)=='A' && EEPROM.read(next_addr+2)=='L' && EEPROM.read(next_addr+3) == char(i+48) &&EEPROM.read(next_addr+4)==':')
```
nikitakit, line 32: Writing `48` as an integer is not strictly necessary (see my other comment on this issue)

nikitakit (changes requested):
Here's my first pass review of the code

dHutchings:
This PR is being closed, because the newer (and rebased) PDB PR has been merged.

PR #173, opened 2017-03-02
Remove redux devtools

AllanYangZhou:


kevinmasd (approved):
Never actually knew we used this.

PR #174, opened 2017-03-02
[RUNTIME] Clean up/refactor code

kevinji:
We don't really need those `r`, `g`, and `a` vars.


runtime/testy/runtime.py
```diff
@@ -269,14 +264,14 @@ def runtimeTest(testNames):
...
-  if len(testNames) == 0:
-    print("Running all tests")
-    testNames = allTestNames
-  else:
+  if testNames:
     for testName in testNames:
       if testName not in allTestNames:
         print("Error: {} not found.".format(testName))
         return
+  else:
```
yizow: Style thing i picked up over the summer: 

With if/ else chains, you want to structure the code such that the simpler case comes first, and the more complicated case comes second. It's a readibility thing: that way, people can read the simple case and move on to the complicated case while the conditional clause is still fresh in their mind. 

tl;dr change to `if not testNames` and swap the if/else blocks

runtime/testy/runtime.py
```diff
@@ -300,13 +295,14 @@ def runtimeTest(testNames):
       if PROCESS_NAMES.TCP_PROCESS in allProcesses:
         terminate_process(PROCESS_NAMES.TCP_PROCESS)
       sys.stdout = sys.__stdout__
-      print("{}DONE!".format(" "*(50-len(testName))))
-    if not testSuccess(testFileName):
+      print("{}DONE!".format(" " * (50-len(testName))))
+
+    if testSuccess(testFileName):
```
yizow, line 300: Oh, this is the perfect example of what i mentioned above. good fix.

runtime/testy/studentAPI.py
```diff
@@ -61,17 +62,16 @@ class Gamepad(StudentAPI):
...
+    super().__init__(toManager, fromManager)
 
   def get_value(self, name, gamepad_number=0):
-    gamepad_dict = self._getSMValue('gamepads')[gamepad_number]
-    if name in self.joysticks.keys():
+    gamepad_dict = self._getSMValue("gamepads")[gamepad_number]
+    if name in self.joysticks:
       return gamepad_dict["axes"][self.joysticks[name]]
-    elif name in self.buttons.keys():
+    if name in self.buttons:
```
yizow: why remove the `elif`?

runtime/testy/studentAPI.py
```diff
@@ -109,13 +109,13 @@ def run(self, fn, *args, **kwargs):
   def is_running(self, fn):
     if not inspect.isfunction(fn):
       raise StudentAPIValueError("First argument to Robot.run must be a function")
-    elif not inspect.iscoroutinefunction(fn):
+    if not inspect.iscoroutinefunction(fn):
```
yizow: same as above. seems like an unneccessary change

runtime/testy/studentAPI.py
```diff
@@ -143,15 +143,12 @@ def _hibikeSubscribeDevice(self, uid, delay, params):
     self.toManager.put([HIBIKE_COMMANDS.SUBSCRIBE, [uid, delay, params]])
 
   def _hibikeGetUID(self, name):
-    try:
-      return self.sensorMappings.get(name)
-    except:
-      raise StudentAPIKeyError()
+    return self.sensorMappings.get(name)
```
yizow: we want to return a studentApiError, not the normal KeyError

yizow: allows us to extend functionality more easily.

runtime/testy/studentAPI.py
```diff
@@ -61,17 +62,16 @@ class Gamepad(StudentAPI):
     "joystick_right_y" : 3
   }
 
-  def __init__(self):
-    super().__init__(self.toManager, self.fromManager)
+  def __init__(self, toManager, fromManager):
```
yizow, line 65: this seems like almost a functional change.
@brandonxxlee

yizow (changes requested):
I like your readibility fixes and cleanups.

Few comments re: style.

runtime/testy/studentAPI.py
kevinji: The code as before wouldn't throw an error in the first place, since we're using `.get` instead of a standard dictionary `[]` access.

runtime/testy/studentAPI.py
kevinji: I guess I personally find it simpler if we have `return` statements since there's no need for the `el` part.

runtime/testy/runtime.py
kevinji: Hmm I was more following the principle of "always use positive `if` conditions with `if`-`else` statements, since they're easier to reason about".

runtime/testy/studentAPI.py
brandonxxlee: From a quick glance it seems to still preserve functionality, but I haven't actually tested it.

runtime/testy/runtime.py
yizow: not always. 
And in this case, its approx the same either way.

runtime/testy/studentAPI.py
yizow: functionally, theyre the same.

I would argue that elif is more clear. One does not need to go "oh this if has a return, so only one of these will execute" 

More conciseness is not always more readibility.

runtime/testy/studentAPI.py
yizow: ah, good point.

yizow (changes requested):
Still would like the `if`/`else` block swapped and the `elif` statements kept

kevinji:
@yizow Changes should be addressed with the latest commit.

yizow (changes requested):
LGTM

yizow (approved):

PR #175, opened 2017-03-02
[DEVOPS] travis: Slightly simplify Travis builds

kevinji:


.travis.yml
```diff
@@ -2,22 +2,24 @@ dist: trusty
 sudo: required
 language: node_js
 node_js:
-- "6"
+- "7"
```
kevinmasd: We use LTS, which is still in the 6's.

kevinmasd (changes requested):

.travis.yml
kevinji: Fair enough, I've reverted that change.

kevinmasd (approved):

.travis.yml
```diff
@@ -36,9 +38,9 @@ script:
 branches:
   only:
   - master
-  - /^(dawn)\/.*$/
-  - /^(runtime)\/.*$/
-  - /^(test)\/.*$/
+  - /^dawn\/.*$/
+  - /^runtime\/.*$/
+  - /^test\/.*$/
 deploy:
```
nikitakit, line 45: The glob on line 48 no longer matches, so this will actually break builds. You can test by tagging a copy of your PR code as `test/dawn/...` or `test/runtime/...` and verifying that artifacts are actually pushed to github.

nikitakit (changes requested):

.travis.yml
nikitakit: There's a chance that I don't know how the deploy globs work (since I haven't used travis much), but I would like to know that you tested the release procedure after making this change.

kevinji: Ugh, of course I'd miss something. I'll back out those changes since there doesn't seem to be an easy way to resolve them.

.travis.yml
```diff
@@ -1,23 +1,26 @@
 dist: trusty
 sudo: required
-language: node_js
-node_js:
-- "6"
+language: python
+python:
+- "3.6"
```
nikitakit: I think we've standardized on python 3.5, at least in terms of what we're deploying to the BBBs. So I don't think it makes sense to use a different version for travis.

kevinji: Fixed.

nikitakit (approved):

PR #176, opened 2017-03-02
[DAWN] Make various UI bug fixes

kevinji:
The specific changes are in the commit messages.


kevinmasd (approved):
While the Joyride fixes are a bit extraneous since it doesn't even work right now, the rest of the fixes look good.

PR #179, opened 2017-03-02
[RUNTIME] Integration/v0.2

yizow:
Will be adding more comments soon.


yizow:
Hibike: 
When hibike_process starts, it first enumerates all devices, and then tells them to stop sending data. 
When hibike receives a message from a device that has not previously enumerated, it will tell it to stop sending. 

Runtime: 
Runtime uses hibike_process.py from Hibike instead of hibikeSim.py

UID's sent to dawn have the lower 24 bits truncated to fit the remaining UID within 64 bits. Since Dawn only ever uses these uid's internally, we do not need to worry about "unmuxing" the truncated uid's

@bodiddles the change in runtime.py from a break to a continue may change some UDP functionality; i'm not sure, this was a merge conflict i tried to fix. I made it follow the other if blocks.

stateManager:
updates setValue/ createKey to take an optional argument for whether or not to send the result to studentCode pipe. 

Disables student code logging because sending couldnt keep up with logging.

Updates hibike/runtime api to send hibike commands and properly update stateManager internal state upon receiving a hibike data packet.

studentApi:
adds a few functions to communicate with hibike. just wrappers.

studentCode:
adds teleop code to drive the motor either high or low
removes obsolete hibike tests.

runtime/testy/Ansible.py
```diff
@@ -111,7 +111,7 @@ def package(state):
                 proto_message.robot_state = state['studentCodeState'][0]
                 for uid, values in state['hibike'][0]['devices'][0].items():
                     sensor = proto_message.sensor_data.add()
-                    sensor.uid = uid
+                    sensor.uid = uid >> 24
```
kevinji: Might want to document why this shift is necessary.

runtime/testy/studentCode.py
```diff
@@ -8,16 +8,53 @@ def autonomous_main():
   pass
 
 def teleop_setup():
-  pass
+  print("setup")
+  devices = Robot.getValue("hibike", "devices")
```
kevinji: This needs to be changed to `Robot._getSMValue` (along with the other calls below) in order to work on `master`.

runtime/testy/stateManager.py
```diff
@@ -152,9 +156,10 @@ def student_upload(self):
     self.processMapping[PROCESS_NAMES.TCP_PROCESS].send([ANSIBLE_COMMANDS.STUDENT_UPLOAD, True])
 
   def send_console(self, console_log):
+    return
```
kevinji: You probably know this already, but this `return` should probably be removed prior to merging.

kevinji (changes requested):
Just a few minor things I saw (not really about functionality).

runtime/testy/Ansible.py
yizow: Fixed

runtime/testy/stateManager.py
yizow: This is deliberate to enable console logging to not crash the robot. see pull request message for details.

runtime/testy/studentCode.py
yizow: Fixed

runtime/testy/stateManager.py
```diff
@@ -129,10 +131,12 @@ def setValue(self, value, keys):
...
       for item in path:
         item[1] = currTime
-      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(value)
+      if send:
+        self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(value)
     except:
       error = StudentAPIKeyError(self.dictErrorMessage(i, keys, currDict))
-      self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
+      if send:
+        self.processMapping[PROCESS_NAMES.STUDENT_CODE].send(error)
```
brandonxxlee, line 139: This looks fine to me, but what is the use case of sending a value to StudentCode after setting it in SM?

runtime/testy/studentAPI.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, toManager, fromManager):
     super().__init__(toManager, fromManager)
     self._createSensorMapping()
     self._coroutines_running = set()
+    self.hibikeEnumerateDevices()
```
brandonxxlee, line 81: Shouldn't this be enumerated when SM is started up, not each time StudentCode is started up?

runtime/testy/Ansible.py
```diff
@@ -213,6 +214,7 @@ def unpackage(data):
             new_state = received_proto.student_code_status
             unpackaged_data["student_code_status"] = [new_state, time.time()]
             if self.control_state is None or new_state != self.control_state:
+                print(new_state, "UDP")
```
brandonxxlee: Do we want to keep this print statement in Ansible?

brandonxxlee (changes requested):

runtime/testy/studentCode.py
```diff
@@ -8,10 +8,35 @@ def autonomous_main():
   pass
 
 def teleop_setup():
-  pass
+  print("setup")
+  devices = Robot._getValue("hibike", "devices")
```
kevinji: I think this is still the wrong function name (should be `_getSMValue`).
On another topic, it'd probably be good to standardize our function naming convention to either be under_scores (my vote) or camelCase, as it's currently inconsistent across the codebase.

runtime/testy/stateManager.py
kevinji: Maybe make a comment about that in the code?

runtime/testy/runtime.py
```diff
@@ -86,7 +86,7 @@ def nonTestModePrint(*args):
           spawnProcess(PROCESS_NAMES.UDP_RECEIVE_PROCESS, startUDPReceiver)
           dawn_connected = False
           controlState = "idle"
-          break
+          continue
```
brandonxxlee: I think this should be break, since we want to kill studentCode if Dawn disconnects. @bodiddles ?

runtime/testy/studentAPI.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, toManager, fromManager):
...
+    # Wait for hibike to respond
+    time.sleep(0.5)
```
kevinji, line 83: This is kinda inelegant; is there a more robust way to wait for Hibike?

kevinji (dismissed):
Mostly about the same things as before.

runtime/testy/studentAPI.py
```diff
@@ -155,3 +158,9 @@ def _print(self, *args):
     print(*args)
     console_string = " ".join([str(arg) for arg in args])
     self.toManager.put([SM_COMMANDS.SEND_CONSOLE, [console_string]]) 
+
+  def hibikeEnumerateDevices(self):
```
brandonxxlee, line 162: Same as above, do we need this function in studentAPI?

runtime/testy/studentAPI.py
```diff
@@ -155,3 +158,9 @@ def _print(self, *args):
...
+    self.toManager.put([HIBIKE_COMMANDS.ENUMERATE, []])
+
+  def hibikeWriteValue(self, uid, params):
```
brandonxxlee, line 165: I have a duplicate function as this + param checking in #180

runtime/testy/stateManager.py
yizow: Done

runtime/testy/stateManager.py
yizow: Student code blocks until it receives a return value. 
That way students don't need to treat their main loop asynchronously

runtime/testy/studentAPI.py
yizow: This will error if hibike isn't spawned yet.
Process spawning order: 
SM -> Hibike -> Student code

putting it in student code is the only way to get the ordering synced up.

We can also put it in hibike. in fact, hibike already does that. This is for additional guarantees in case... hibike break.

runtime/testy/Ansible.py
yizow: It was from you guys. Let me know if i should remove it or not. 
Or you can push to this branch as well.

runtime/testy/studentCode.py
yizow: ah crap. fixing.

the "refactor entire codebase to be consistent" has been on the list of things to do for a while now... 
it's my fault, i started the whole thing in camelCase when it should have been underscores x_x 
(i like underscores more too, but i thought that python style was camelCase. Yes, i know i'm an idiot)

runtime/testy/studentAPI.py
yizow: Nothing that does not require a decidedly non-trivial amount of additional runtime infrastructure changes. 
This is not actually needed i don't think, because hibike does enumeration already but... 
this is a just in case thing.

runtime/testy/studentAPI.py
yizow: we'll keep this in for now, and have #180 delete if necessary.
i believe you also stated that pull request is based on this, right?

runtime/testy/studentAPI.py
yizow: At least for current debugging, i see no reason to remove it right now.

runtime/testy/Ansible.py
bodiddles: We should delete this print statement, it was just used for print line debugging.

runtime/testy/runtime.py
bodiddles: Yes, we should break since if dawn actually disconnects, we should probably stop the robot from moving. In terms of potential issues with shoddy connections, the TCP socket which we use to detect disconnects should have its own timeouts.

runtime/testy/stateManager.py
```diff
@@ -152,9 +156,11 @@ def student_upload(self):
     self.processMapping[PROCESS_NAMES.TCP_PROCESS].send([ANSIBLE_COMMANDS.STUDENT_UPLOAD, True])
 
   def send_console(self, console_log):
+    # Short circuit since console sending can't currently keep up with student code logs.
```
bodiddles: Add this as a TODO

runtime/testy/studentCode.py
```diff
@@ -8,10 +8,35 @@ def autonomous_main():
   pass
 
 def teleop_setup():
-  pass
+  print("setup")
```
bodiddles: Is there a reason this is going to be merged in? If it's for examples, this should be in a separate file documenting how to write this code rather than merging this into master.

runtime/testy/studentCode.py
```diff
@@ -102,14 +127,6 @@ def createKey_main():
   restarts = Robot._getSMValue("Restarts")
   Robot._setSMValue(restarts+1, "Restarts")
 
-def hibikeSubscribeDevice_setup():
```
bodiddles, line 105: Is there a reason we are deleting these tests rather than just modifying them to reflect accurate behavior?

bodiddles (changes requested):
Mostly LGTM but had some questions

brandonxxlee (approved):

bodiddles (approved):

PR #180, opened 2017-03-02
[RUNTIME] Add student API get and set values

brandonxxlee:
Adds get_value and set_value functions into the student API
Now requires students to also pass in a param field, which will be specified in the API.
Adds basis of framework to run parameter/value checking (currently on motors checked)
Adds get_value test

This PR is blocked by #179 .


runtime/testy/stateManager.py
```diff
@@ -73,7 +73,7 @@ def initRobotState(self):
      "list1" : [[[70, t], ["five", t], [14.3, t]], t],
      "string1" : ["abcde", t],
      "runtime_meta" : [{"studentCode_main_count" : [0, t], "e_stopped" : [False, t]}, t],
-     "hibike" : [{"device_subscribed": [0, t], "devices" : [{12345 : [{"sensor0": [1, t]}, t]}, t]}, t],
```
yizow: this is gonna really mess up actual testing....
can we give the uid a clearly fake one? 
like `12345`?

yizow: Or better yet, make the setup() function in the test set these parameters in studentCode.

runtime/testy/studentAPI.py
```diff
@@ -74,11 +74,31 @@ def get_value(self, name, gamepad_number=0):
...
+    "enable" : ["boolean"],
+  }
+
   def __init__(self, toManager, fromManager):
     super().__init__(toManager, fromManager)
     self._createSensorMapping()
     self._coroutines_running = set()
 
+  def get_value(self, device_name, param):
+    uid = self._hibikeGetUID(device_name)
```
yizow: uid has a chance to return `None`

runtime/testy/studentAPI.py
```diff
@@ -74,11 +74,31 @@ def get_value(self, name, gamepad_number=0):
...
+    self._check_params(uid, param)
+    return self._getSMValue('hibike', 'devices', uid, param)
+
+  def set_value(self, device_name, param, value):
+    #TODO Implement parameter checking
+    uid = self._hibikeGetUID(device_name)
```
yizow, line 100: chance to return `None`

runtime/testy/studentAPI.py
```diff
@@ -114,6 +134,19 @@ def is_running(self, fn):
...
 
+  def _check_params(self, uid, param):
+    device_type = uid >> 72
+    valid_params = self.deviceType_to_validParams[device_type]
+    if param not in valid_params:
+      raise StudentAPITypeError("Invalid param passed in, valid parameters for this device are: " + "".join(valid_params))
+
+  def _check_value(self, param, value):
+    valid_values = self.param_to_valid_values[param]
+    if valid_values[0] == "boolean":
```
yizow: so in your dictionary, instead of storing "bool", store the class `bool`
the name of the class as a string: `bool.__name__`
this way, you can do a more robust check of `if isinstance(value, bool)`
http://stackoverflow.com/questions/152580/whats-the-canonical-way-to-check-for-type-in-python

runtime/testy/studentAPI.py
```diff
@@ -114,6 +134,19 @@ def is_running(self, fn):
...
+      return
+    elif valid_values[0] == "float" and not (valid_values[1] < value < valid_values[2]):
```
yizow: i would structure this function as:
```
if {check type}: 
   if {check value range}
elif {check type}:
  if {check value range}
.....
```

yizow (changes requested):
I dont think we should add an actual motor to the state directly.

Some implementation recommendations.

runtime/testy/studentAPI.py
brandonxxlee: Pretty sure this should through an error if the key doesn't exist, and thus can not ever be `None`

runtime/testy/stateManager.py
brandonxxlee: Done

runtime/testy/studentAPI.py
kevinji: `self._hibikeGetUID(device_name)` currently redirects to `self.sensorMappings.get(name)`, which silently returns `None` in the case of a key-not-found instead of throwing an Exception. You can change that behavior if you'd like though.

runtime/testy/studentAPI.py
brandonxxlee: Done

bodiddles (approved):
LGTM

yizow (approved):
LGTM

PR #182, opened 2017-03-03
Implement tab-completion of device names for ./flash.sh

dHutchings:
Tab completion is setup on a linux machine as part of running ./setup.sh

Only works on linux. However, since the setup for compilation of hibike
also only works on linux, I think this is fine.

Done based on this documenation: https://debian-
administration.org/article/317/An_introduction_to_bash_completion_part_2


hibike/utils/flash
```diff
@@ -0,0 +1,26 @@
...
+#
+_flash.sh() 
+{
+    local cur prev opts
+    COMPREPLY=()
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+    opts="$(./utils/device_names.sh)"
+ 
+	COMPREPLY=( $(compgen -W "${opts}" -- ${cur}))
```
nikitakit: Can you fix the inconsistent indentation?

hibike/utils/flash
```diff
@@ -0,0 +1,26 @@
...
+	return 0
+}
+complete -F _flash.sh flash.sh
```
nikitakit: newline at eof

hibike/utils/device_names.sh
```diff
@@ -0,0 +1,10 @@
+#/bin/bash
+
+#returns a list of all device names.  Used in combination with other linux hackery for tab-completion
+
+
+#this function is not meant to be called from the "utils" folder.  It will fail because of pathing problems.
+#it is meant to be called from the "hibike" folder, since that is where flash.sh is located.
+
+
+ls devices/
```
nikitakit: Could you add a newline at the end of the file?

(See how there's a red icon when github displays this line? That means there's no newline at eof. It's standard practice to end a file in a newline, because of how lack of a newline interacts with diffing tools)

nikitakit (changes requested):
I have a few style comments.

There's probably a way to make this more robust and not call `device_names.sh` from random locations (as well as not complete when device names don't exist), but for an internal tool that's not really necessary.

hibike/utils/flash
dHutchings: Done

hibike/utils/flash
dHutchings: done

hibike/utils/device_names.sh
dHutchings: Done

dHutchings:
@nikitakit 

Fixed your style comments.  Also, implemented extra logic so that it won't erroneously attempt to run the device_names script unless the user's terminal is located in the correct filepath.

nikitakit (approved):

PR #183, opened 2017-03-03
Fix UIDs not actually being random.

dHutchings:
Make clean is broken.  Therefore, code won't compile a new RANDOM_UID everytime.

Therefore, smart devices that have been successively loaded will have idential UIDs, which breaks things.

Quick & Dirty fix is just rm -rf the bin file.


hibike/flash.sh
```diff
@@ -1,5 +1,8 @@
 #/bin/bash
-make clean
+
+#make clean doesn't actually work.  Therefore,
+rm -rf ./bin
```
karthik-shanmugam: This might do bad things if ./flash.sh is accidentally run outside of `PieCentral/hibike` but this probably won't be a major issue.

karthik-shanmugam (approved):
hopefully we find a proper fix for this issue later but if this works it's fine

nikitakit (changes requested):
Please do cd dirname $0

nikitakit:
I think accidentally deleting the wrong bin folder can be pretty bad. Thanks @karthik-shanmugam for pointing this out.

Doing
```
cd `dirname $0` && rm -rf ./bin
```

is one possible fix.

Also bash will move to the next line even if the previous one fails, hence use "&&" or disable this behavior with "set -e"

dHutchings:
@nikitakit   Fixed file according to your request.

nikitakit (approved):

PR #184, opened 2017-03-03
[DAWN] Editor: Improve the unsaved changes dialog

kevinji:
The dialog is more succinct, and the program no longer closes on reloading.
Yes, there is a bug where hitting "Save..." won't close the program afterwards, but that bug is already present in the current version on `master`; I just mentioned it in a comment for visibility purposes.


dawn/renderer/components/Editor.js
```diff
@@ -141,6 +124,34 @@ class Editor extends React.Component {
     return text.replace(/[^\x00-\x7F]/g, ''); // eslint-disable-line no-control-regex
   }
 
+  beforeUnload(e) {
+    // If there are unsaved changes and the user tries to close Dawn,
+    // check if they want to save their changes first.
+    if (this.hasUnsavedChanges()) {
+      const clickedId = dialog.showMessageBox(currentWindow, {
```
kevinmasd: Hang for #162, which has some renamings regarding this component.

kevinmasd (changes requested):

dawn/renderer/components/Editor.js
kevinji: Sounds good, I'll rebase on top afterwards.

kevinji:
@kevinmasd Rebased on top of master. I've tried to avoid using `currentWindow.close()`, as that doesn't always make sense (e.g. if you just refresh the editor), and functionally this should work the same as before.

kevinmasd (approved):

PR #185, opened 2017-03-04
[RUNTIME] Add test for Actions.sleep

jiangdaniel:


runtime/testy/studentCode.py
```diff
@@ -203,3 +204,34 @@ def optionalTestsWork_setup():
...
+  if not sleepTestVal['test']:
+    print('Success!')
+  else: 
+    print('Failed to sleep for the correct amount of time')
+
+async def asyncSleepHelper2(sleepTestVal):
+  sleepTestVal['test'] = True
+  await Actions.sleep(.5)
+  sleepTestVal['test'] = False
+
```
kevinji: Really minor thing but I think this is an extra newline.

yizow (changes requested):
Remove one extra new line at end of file, otherwise LGTM

kevinji:
@yizow Just pushed a change that removes the newline, so this should be good to go after Travis finishes.

jiangdaniel:
@mc10 Thanks!

yizow (approved):

PR #189, opened 2017-03-04
[RUNTIME] Minor Bug Fix

bodiddles:
Adds a null check to sending param values which was causing some errors. 

Additionally, changes flag for student code state after studentCode terminates to reflect the actual state of studentCode.


runtime/testy/Ansible.py
```diff
@@ -116,7 +116,8 @@ def package(state):
                     for param, value in values[0].items():
                         param_value_pair = sensor.param_value.add()
                         param_value_pair.param = param
-                        param_value_pair.float_value = value[0]
+                        if value[0]:
```
nikitakit: Do you even want to send the param_value pair if you don't have a value for it?

yizow: I would argue yes. 
It is possible that we do not have any sensors connected (even though its an incorrect configuration because PDB) 

In that scenario, we still want to send the state of the robot via studentCodeState

runtime/testy/runtime.py
```diff
@@ -113,6 +114,7 @@ def nonTestModePrint(*args):
           break
       stateQueue.put([SM_COMMANDS.RESET, []])
       terminate_process(PROCESS_NAMES.STUDENT_CODE)
+      stateQueue.put([SM_COMMANDS.SET_VAL, [runtime_pb2.RuntimeData.STUDENT_STOPPED, ["studentCodeState"]]])
```
yizow: Does this need to be `["studentCodeState"][0]`?

yizow (changes requested):

runtime/testy/runtime.py
kevinji: Wait, that would just return the string `studentCodeState`...

runtime/testy/Ansible.py
brandonxxlee: I think Nikita means just for this for loop (the para value pairs), in which case I agree we could probably just skip that one sensor entirely (not skip sending the whole proto)

brandonxxlee (changes requested):
Should also add `Ansible.py` correctly using the stateManager value of state, not its internally stored value of state.

runtime/testy/runtime.py
```diff
@@ -114,7 +114,7 @@ def nonTestModePrint(*args):
           break
       stateQueue.put([SM_COMMANDS.RESET, []])
       terminate_process(PROCESS_NAMES.STUDENT_CODE)
-      stateQueue.put([SM_COMMANDS.SET_VAL, [runtime_pb2.RuntimeData.STUDENT_STOPPED, ["studentCodeState"]]])
+      stateQueue.put([SM_COMMANDS.SET_VAL, [runtime_pb2.RuntimeData.STUDENT_STOPPED, ["studentCodeState"], False]])
```
yizow: `set=False` will be more readable i feel. 
Personally i'm a fan of always naming keyword args in function calls, because the fact that theyre optional can be confusing on first read.

yizow (changes requested):

brandonxxlee (approved):

yizow (approved):

PR #191, opened 2017-03-04
[RUNTIME] Remove enumeration and resetting SM

brandonxxlee:
Remove enumeration of Hibike in studentAPI
Remove resetting of SM after ever studentCode restart
Update tests to reflect new behavior

I changed the tests simply to how they behave now, however was not super sure what they originally were designed to test so maybe change that aspect @yizow .


bodiddles (approved):
LGTM

PR #192, opened 2017-03-04
[HIBIKE] Implement Heartbeat Feature

tingtingchiang:
Implementing basic heartbeat feature between hibike and sensor devices. Hibike will automatically send a heartbeat response package in response to every data_update package it receives from the sensor. There is also an edit to the method hibike_setup to describe a disable latency, which is how long the sensor will wait to receive a heartbeat response before disabling itself.

All known sensor_devices have been changed to reflect these changes. KNOWN ISSUE: LimitSwitch, LineFollower, Potentionmeter, ServoControl, and PDB all will not compile because of previous incompatibility with current hibike protocol. TeamFlag, YogiBear, and ExampleDevice are tested and will compile.

Though not implemented yet, there is documentation for a Heartbeat Request package. The intent is to implement this package so that heartbeats will not be reliant on Data Update packages in the future.


hibike/README.md
```diff
@@ -152,7 +152,9 @@ Message ID Enumeration:
 |  0x12   |   Subscription Response  |
 |  0x13   |       Device Read        |
 |  0x14   |       Device Write       |
-|  0x15   |      Device Data         |
+|  0x15   |       Device Data        |
+|  0x16   |    Heart Beat Request    |
```
dHutchings: 0x16 as Heart Beat Request won't work, since 0x16 has been allocated to the "Disable" in the newly-merged disable commits.

Please reallocate to 0x16 and 0x17

hibike/README.md
```diff
@@ -339,7 +341,35 @@ Note: These assignments are also fairly random and may not all even be
     Direction:
     BBB <-- SD
 
-7. Error Packet: Sent to indicate an error occured. 
+7. Heart Beat Request: BBB requests SD message for connectivity purposes.
```
dHutchings: According to the rest of the text, both the BBB and the SD may send a heart beat request.  Please update the top line of documentation to accurately match the lines below.

hibike/README.md
```diff
@@ -339,7 +341,35 @@ Note: These assignments are also fairly random and may not all even be
...
+
+        +---------------+
+        |       ID      |
+        |    (8 bits)   |
+        +---------------+
+
+    Direction:
+    BBB --> SD
+
+8. Heart Beat Response: SD sends BBB message for enumeration purposes.
```
dHutchings: Similar comment here.

Also, please explain a little about how heartbeats work.  (I.E. a heart beat requests results in a heart beat response)

hibike/devices/BatteryBuzzer/battery_buzzer.cpp
```diff
@@ -38,7 +38,7 @@ void setup() {
 
   delay(100);  // Wait for the first reading to finish (50ms minimum)
 
-  hibike_setup();
+  hibike_setup(500); //parameter indicateds disable latency in milliseconds
```
dHutchings: "indicateds"?  "Indicates"?  Seems like copy-pasta has spread this spelling mistake around.

hibike/hibike_message.py
```diff
@@ -44,6 +44,8 @@
   "DeviceRead" :           0x13,
   "DeviceWrite" :          0x14,
   "DeviceData" :           0x15,
+  "HeartBeatRequest" :     0x16,
```
dHutchings: Again, please change 0x16 and 0x17 to 0x17 and 0x18

hibike/hibike_process.py
```diff
@@ -104,6 +107,7 @@ def device_read_thread(index, ser, instructionQueue, errorQueue, stateQueue):
             if uid is not None:
                 params_and_values = hm.parse_device_data(packet, hm.uid_to_device_id(uid))
                 stateQueue.put(("device_values", [uid, params_and_values]))
+                instructionQueue.put(("heartResp", [uid]))
```
dHutchings, line 118: I know that this code was implemented this way in order to get a MVP up quickly.  We understood that there'd be a big overhead when we did it this way.

Given the latency testing that yizhe is doing now, I'd advocate moving into a model where Smart Devices send heart beat requests ASAP.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -1,24 +1,24 @@
 #include "hibike_device.h"
 
 message_t hibikeBuff;
-uint64_t prevTime, currTime, heartbeatTime;
+uint64_t prevTime, currTime, heartbeatTime,sentHeartbeat, respHeartBeat;
 uint16_t params, old_params;
 uint32_t value;
-uint16_t subDelay;
+uint16_t subDelay,disableLatency;
```
dHutchings: disableLatceny should not be a uint16.

There is arithmetic done between disableLatency and millis().  millis() returns an unsigned long, (uint32).  Please make sure that disableLatency is also an unsigned long.

In the current implimentation, there will be arithmetic errors after millis - resp heart beet exceeds 65535, otherwise known as 65 seconds.  Given that a match is 2:30 long (excluding disabled periods), it'd totally be feasible for a problem to occur here.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -141,3 +113,12 @@ void toggleLED() {
     led_enabled = true;
   }
 }
+
+uint64_t getLastHeartbeatResponse(){
```
dHutchings: Are these functions used?

From a coder's perspective, it'd be more useful to know how long it's been since we got a heartbeat, not the absolute time that we last recieved our heartbeat at.

hibike/lib/hibike/hibike_message.h
```diff
@@ -28,6 +28,8 @@ typedef enum {
   DEVICE_READ             = 0x13,
   DEVICE_WRITE            = 0x14,
   DEVICE_DATA             = 0x15,
+  HEART_BEAT_REQUEST      = 0x16,
```
dHutchings: Again, please change to 0x17 and 0x18

dHutchings (changes requested):
This is a really useful feature.

Given the latency testing that yizhe is doing, I look forward to having actual Heart Beat Requests sent by the Smart Devices.  Should cut down on the overhead here.

I have a few quibbles with regards to types, etc, below.  Please look them over!

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -99,8 +98,8 @@ void hibike_loop() {
     send_data_update(params);
   }
 
-  if ((currTime - respHeartBeat)>disableLatency){
-    //disable(); //on sensor specific device'
+  if ((currTime - resp_heartbeat) > disable_latency) {
+    // disable(); //on sensor specific device'
```
dHutchings: Now that disable() is implimented and merged to master, actually call disable()

This code is contained in the hibike library, E.G. a Smart Device developer should not be changing it.  If the smart Device developer doesn't want to have any disable functionality, they can instead leave disable() blank,

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -113,12 +112,3 @@ void toggleLED() {
     led_enabled = true;
   }
 }
-
-uint64_t getLastHeartbeatResponse(){
```
dHutchings: TY for removing these functions.

Instead, could we please have a function, like the following:

uint32_t get_time_since_heartbeat(){
return millis() - resp_heartbeat;
}

It might be helpful to have this function lying around for the future.

dHutchings (changes requested):
One requested function for future-proofing.  One uncomment of disable(), since developers shouldn't be touching hibike_device.cpp reguarly.

hibike/lib/hibike/hibike_device.cpp
yizow: done

hibike/lib/hibike/hibike_device.cpp
yizow: Then we'll implement it in the future, is my argument.

dHutchings: Fine.

hibike/lib/hibike/hibike_device.cpp
dHutchings: TY

dHutchings (approved):
Looks good.  Thanks for the work Tiffany & yizhe.

PR #193, opened 2017-03-04
[RUNTIME] API Type Checking and Subscription

brandonxxlee:
Add type checking to write and read params and values.
Add auto subscription in State Manager.

These specific params and values are subject to change dependent on hibike.


runtime/testy/studentAPI.py
```diff
@@ -74,32 +74,45 @@ def get_value(self, name, gamepad_number=0):
...
+    "servo0" : [[int], 0, 180],
+    "enable0" : [[bool]],
+    "servo1" : [[int], 0, 180],
+    "enable1" : [[bool]],
+    "servo2" : [[int], 0, 180],
+    "enable2" : [[bool]],
+    "servo3" : [[int], 0, 180],
+    "enable3" : [[bool]],
+    "duty_cycle" : [[float, int], -1, 1],
+    "enable" : [[bool]],
```
bodiddles: Seems like you're missing some expected values for yogibear params.

runtime/testy/studentAPI.py
```diff
@@ -74,32 +74,45 @@ def get_value(self, name, gamepad_number=0):
...
+  deviceType_to_readParams = {
+    0  : ["switch0", "switch1", "switch2"],
+    1  : ["left", "center", "right"],
+    2  : ["pot0", "pot1", "pot2"],
+    10 : ["duty_cycle", "pid_pos_setpoint", "pid_pos_kp", "pid_pos_ki", "pid_pos_kd", "current_thresh", "enc_pos", "enc_vel"],
   }
   param_to_valid_values = {
-    "duty_cycle" : [float, -1, 1],
-    "enable" : [bool],
+    "servo0" : [[int], 0, 180],
```
bodiddles: Your check values method only expects types to be floats or bools. Setting/getting servo values won't be checked correctly then

bodiddles (changes requested):
Some minor fixes but otherwise LGTM

runtime/testy/studentAPI.py
brandonxxlee: Done

runtime/testy/studentAPI.py
brandonxxlee: Done

runtime/testy/stateManager.py
```diff
@@ -208,6 +221,9 @@ def hibikeReadParams(self, pipe, uid, params):
     pipe.send([HIBIKE_COMMANDS.READ.value, [uid, params]])
 
   def hibikeResponseDeviceSubbed(self, uid, delay, params):
+    if delay == 0:
+      deviceName = runtimeUtil.SENSOR_TYPE[uid >> 72]
+      self.hibikeSubscribeDevice(self.processMapping[PROCESS_NAMES.HIBIKE], uid, 20, self.deviceName_to_subscribeParams[deviceName])
```
yizow: can we change this to be `40` ms delay. 
let's start small and work our way up

runtime/testy/studentAPI.py
```diff
@@ -137,19 +153,25 @@ def is_running(self, fn):
...
     device_type = uid >> 72
-    valid_params = self.deviceType_to_validParams[device_type]
+    valid_params = self.deviceName_to_readParams[SENSOR_TYPE[device_type]]
     if param not in valid_params:
       raise StudentAPITypeError("Invalid param passed in, valid parameters for this device are: " + "".join(valid_params))
 
   def _check_value(self, param, value):
     valid_values = self.param_to_valid_values[param]
-    if not isinstance(value, valid_values[0]):
+    if not type(value) in valid_values[0]:
```
yizow: i vote for `isinstance`
http://stackoverflow.com/questions/1549801/differences-between-isinstance-and-type-in-python/1549854#1549854
The second argument can be a *tuple* of args 
http://stackoverflow.com/questions/33311258/python-check-if-isinstance-any-type-in-list
Note that this means we need `valid_values[0]` to be an argument, not a list.
since we shouldnt be overriding them anyway, i feel changing them to all be tuples is fine.

runtime/testy/stateManager.py
brandonxxlee: Oh I thought this in Hertz, yah we should definitely do that.

runtime/testy/stateManager.py
```diff
@@ -19,8 +19,21 @@ def __init__(self, badThingsQueue, inputQueue, runtimePipe):
...
     self.processMapping = {PROCESS_NAMES.RUNTIME: runtimePipe}
 
+  def makeSubscriptionMap(self):
+    subscriptionMap = {
+      "LimitSwitch"  : ["switch0", "switch1", "switch2"],
+      "LineFollower"  : ["left", "center", "right"],
+      "Potentiometer"  : ["pot0", "pot1", "pot2"],
+      "BatteryBuzzer"  : ["total", "safe"],
+      "ServoControl"  : ["servo0", "servo1"],
+      "YogiBear" : ["duty_cycle", "pid_pos_setpoint", "pid_pos_kp", "pid_pos_ki", "pid_pos_kd", "current_thresh", "enc_pos", "enc_vel"],
```
yizow, line 32: we're gonna need to update this.

runtime/testy/studentAPI.py
```diff
@@ -74,12 +74,31 @@ def get_value(self, name, gamepad_number=0):
     raise StudentAPIKeyError()
 
 class Robot(StudentAPI):
-  deviceType_to_validParams = {
-    10 : ["duty_cycle", "enable"], 
+  deviceName_to_writeParams = {
+    "TeamFlag" : ["s1", "s2", "s3", "s4"],
+    "ServoControl" : ["servo0", "enable0", "servo1", "enable1"],
+    "YogiBear" : ["enable", "command_state", "duty_cycle", "pid_pos_setpoint", "pid_pos_kp", "pid_pos_ki", "pid_pos_kd", "current_thresh", "enc_pos"],
```
yizow: gonna need to update

runtime/testy/studentAPI.py
```diff
@@ -74,12 +74,31 @@ def get_value(self, name, gamepad_number=0):
...
+  deviceName_to_readParams = {
+    "LimitSwitch" : ["switch0", "switch1", "switch2"],
+    "LineFollower" : ["left", "center", "right"],
+    "Potentiometer" : ["pot0", "pot1", "pot2"],
+    "YogiBear" : ["duty_cycle", "pid_pos_setpoint", "pid_pos_kp", "pid_pos_ki", "pid_pos_kd", "current_thresh", "enc_pos", "enc_vel"],
   }
   param_to_valid_values = {
-    "duty_cycle" : [float, -1, 1],
-    "enable" : [bool],
+    "servo0" : [(int), 0, 180],
```
yizow: if you want these as tuples, gonna need to add a comma
`(int,)`

runtime/testy/studentAPI.py
```diff
@@ -134,19 +153,25 @@ def is_running(self, fn):
...
+    valid_params = self.deviceName_to_readParams[SENSOR_TYPE[device_type]]
     if param not in valid_params:
       raise StudentAPITypeError("Invalid param passed in, valid parameters for this device are: " + "".join(valid_params))
 
   def _check_value(self, param, value):
     valid_values = self.param_to_valid_values[param]
     if not isinstance(value, valid_values[0]):
-      raise StudentAPIValueError("Invalid value type passed in, valid types for this param are: " + valid_values[0].__name__)
-    if valid_values[0] == bool:
+      raise StudentAPIValueError("Invalid value type passed in, valid types for this param are: " + valid_values[0][0].__name__)
```
yizow, line 177: why not use a `", ",join(valid_values[0])` here?

yizow (changes requested):

runtime/testy/studentAPI.py
brandonxxlee: So technically each parameter has one valid type, its just that we need to count integers also as a float that we need the original list of both.

runtime/testy/stateManager.py
brandonxxlee: Updated YogiBear values, checked with Sam.

brandonxxlee:
This shouldn't be merged until #216 is merged since we used updated YogiBear Params

yizow (approved):

bodiddles (approved):

PR #197, opened 2017-03-05
Integration/chris

Alramech:
Unifies dawn and runtime robot states.
Not sure if this properly exposes the enums to dawn


yizow:
@bodiddles you or brandon should handle updating the runtime side as well.

kevinmasd:
Will probably conflict with #199 . Also build failure.

brandonxxlee:
@bodiddles Proabably needs to rebuild protos
Will update runtime changes, should probably double check all appropriate enum uses are updated.

kevinji:
Don't worry about merge conflicts with my PR; I can handle those after this gets merged in.

brandonxxlee:
I am pretty sure I have updated all instances in runtime.

nikitakit:
FYI the Makefile in the runtime folder should build protos before it runs any of the tests on travis.

This means that either some of the code hasn't been updated correctly, or the build instructions are wrong (which would need to be fixed for both the CI tests and the release builds)

Alramech:
moved to #211

PR #199, opened 2017-03-05
[DAWN] Refactor Ansible, clean up conns on close

kevinji:
Everything is now nicely encapsulated in classes.
Hopefully this also fixes EADDRINUSE issues.


yizow:
Has this been tested on either an actual robot, or your local machine and running fake runtime/ (i'm not entirely sure how to do that though)

kevinji:
This hasn't been tested yet; I need to either play around with a real robot or make FakeRuntime more complete.

kevinji:
@yizow Just tested this with fake runtime and it seems to work fine. Could you possibly test this with a real robot?

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
- * Sends messages when Gamepad information changes
- * or when 100 ms has passed (with 50 ms cooldown)
- */
-ipcMain.on('stateUpdate', (event, data) => {
-  const message = buildProto(data).encode().toBuffer();
-  console.log('Dawn Sent UDP');
-  sendSocket.send(message, sendPort, runtimeIP, (err) => {
-    if (err) {
-      console.error('UDP Sending Error');
+const LISTEN_PORT = 1235;
```
yizow: move these constants to the top of the file next to each other. 

easily accessible, visible, modifiable.

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
+
+    this.socket.bind(LISTEN_PORT);
+  }
+
+  close() {
+    this.socket.close();
+  }
+}
+
+const SEND_PORT = 1236;
```
yizow: move to top

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
+class SendSocket {
+  constructor() {
+    this.sendGamepadMessages = this.sendGamepadMessages.bind(this);
+    this.ipAddressListener = this.ipAddressListener.bind(this);
+
+    this.runtimeIP = '192.168.0.200';
```
yizow: probably make it a global if we're gonna hardcode it.
or hardocde of "DEFAULT_IP" or something

and move to top of file

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
+
+    this.socket.on('close', () => {
+      RendererBridge.reduxDispatch(ansibleDisconnect());
+      console.log('UDP listening closed');
+    });
+
+    this.socket.bind(LISTEN_PORT);
+  }
+
+  close() {
```
yizow, line 95: i'm curious: how is this different than the `this.socket.on` close above? `

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
+     * IPC Connection with ConfigBox.js' saveChanges()
+     * Receives new IP Address to send messages to.
+     */
+    ipcMain.on('ipAddress', this.ipAddressListener);
+  }
+
+  /*
+   * IPC Connection with sagas.js' ansibleGamepads()
+   * Sends messages when Gamepad information changes
+   * or when 100 ms has passed (with 50 ms cooldown)
```
yizow, line 131: where is this throttling/ timeout being done?

dawn/main/ansible/Ansible.js
```diff
@@ -117,62 +44,212 @@ function buildProto(data) {
...
-  throw err;
-});
+  tryUpload() {
+    const message = new Notification({
+      header: Notification.Type.STUDENT_SENT,
+      console_output: '',
+    }).encode().toBuffer();
+
+    this.socket.write(message, () => {
+      console.log('Runtime notified: try 1');
```
yizow: Remove this block. `waitRuntimeConfirm` does this already.

yizow (changes requested):

dawn/main/ansible/Ansible.js
kevinji: Done.

dawn/main/ansible/Ansible.js
kevinji: Done.

dawn/main/ansible/Ansible.js
kevinji: Done.

dawn/main/ansible/Ansible.js
kevinji: Done.

dawn/main/ansible/Ansible.js
kevinji: I'm actually not sure; I just copied the comment and the corresponding function from before. Should I just remove that section of the comment?

dawn/main/ansible/Ansible.js
kevinji: It's not, but I wasn't sure how to avoid that code duplication, since the sockets are not always the same (between the UDP and the TCP sockets), so the methods you call for closing aren't the same (`.close` vs `.end`).

dawn/main/ansible/Ansible.js
kevinmasd: It's done in the ansibleGamepads function in sagas.js, which bothers updateMainProcess to send a stateUpdate through the IPC.

dawn/main/ansible/Ansible.js
yizow: dont we just... delete this part of the code?

kevinji: Oh, I see what you're referring to. I think it makes more sense to delete it from the `.on('error')` case (which I'll update to do), as you still want `.close()` to work for a normal shutdown.

yizow (approved):
DONT MERGE UNTIL YOU RE-TEST ON ROBOT

kevinji:
Looks like it's working on fake runtime and the robot (minus the one typo fix), so I'm merging it in.

PR #203, opened 2017-03-05
[RUNTIME] Add support for sending different types of values

brandonxxlee:
Add all value types to runtime.proto
Add ability to send ints, bools, and floats to Dawn in Ansible.


bodiddles (commented):
LGTM, have you tested it

bodiddles:
Also these changes seem like they should be in #193

bodiddles (approved):
LGTM

PR #204, opened 2017-03-05
[RUNTIME] Basic Teleop code

yizow:


runtime/testy/studentAPI.py
```diff
@@ -75,7 +80,7 @@ def get_value(self, name, gamepad_number=0):
 
 class Robot(StudentAPI):
   deviceType_to_validParams = {
-    10 : ["duty_cycle", "enable"], 
+    10 : ["duty_cycle", "enable", "enc_pos", "enc_vel", "motor_current"],
```
brandonxxlee, line 83: My pull request #193 should include all the relevant fields that as far as I know should be valid params. It should also auto-subscribe to all the fields we want to.

runtime/testy/studentCode.py
```diff
@@ -9,10 +9,28 @@ def autonomous_main():
   pass
 
 def teleop_setup():
-  pass
+  uids = Robot._getAllUIDs()
```
brandonxxlee, line 12: So is having "testing" teleop code something we want to leave on master, or should we just have a specific branch for this? Also, as above, (still need to test) but we should be able to auto subscribe.

runtime/testy/runtimeUtil.py
```diff
@@ -133,3 +135,7 @@ class TimeoutError(Exception):
 
 class RuntimeError(Exception):
   pass
+
+# Sensor type names are CamelCase, with the first letter capitalized as well
+config_file = open(os.path.join(os.path.dirname(__file__), '../../hibike/hibikeDevices.json'), 'r')
```
brandonxxlee, line 140: I know this was a workaround to find the type of a motor, but this feels kind of jank to merge into master, i.e. we really shouldn't be creating this dictionary.

brandonxxlee (approved):
Mainly LGTM, I'm just not extremely sure whether we want to have all this be added into master if its just for testing, and is subject to frequent changes.

brandonxxlee (changes requested):
I'm going to deny it for now so it doesn't accidentally get merged in with studentCode that is going to get changed very soon anyways.

runtime/testy/studentAPI.py
yizow: I feel we should be using the hibike json file, not hard coding it ourselves.

runtime/testy/studentCode.py
yizow: I think so.
The "test" is "can we drive a robot". 
A separate branch is almost guaranteed to start diverging

runtime/testy/runtimeUtil.py
yizow: The alternative is to 
1) have hibike auto subscribe?
or
2) hard code what to subscribe to in runtime

runtime/testy/studentAPI.py
brandonxxlee: The problem is that we only want to have certain params as being available to the student, so we would have to filter the json file anyways.

runtime/testy/studentCode.py
brandonxxlee: So the counterargument I have is that master is what we will very soon be shipping directly from. We don't want to ship the software with preexisting code in `student_code.py`  This `teleop_setup()` code is exactly the function that students would write to. If they for some reason happened to press play without upload, it would run this code.

runtime/testy/runtimeUtil.py
brandonxxlee: Runtime hard code subscription is being addressed in #193

runtime/testy/studentCode.py
kevinji: We could just have two files: `student_code.py` as a barebones template, and `test_code.py` for our drive test code.

nikitakit: This reminds me... I just opened #208 because I forsee a potential problem.

dHutchings:
Hey all.  Just to jump in here.

Testing various robot issues was made much more difficult tonight because there wasn't a 'known good student code'.  Bob & Brandon had to spent at least an hour re-writing and debugging custom student codes.

I invite you to finish this PR, or an equivilant one, ASAP.

PR #205, opened 2017-03-05
[DAWN] npm/babel: Clean up config, update plugins

kevinji:
Also, remove unneeded npm dependencies.


kevinmasd (approved):
Looks like it still runs right

PR #207, opened 2017-03-06
[DAWN] Make various bug fixes

kevinji:
In particular, IP addresses work again.


dawn/renderer/utils/sagas.js
```diff
@@ -187,10 +187,10 @@ const _timestamps = [0, 0, 0, 0];
 
 function _needToUpdate(newGamepads) {
   return _.some(newGamepads, (gamepad, index) => {
-    if (!_.isUndefined(gamepad) && (gamepad.timestamp > _timestamps[index])) {
+    if (gamepad !== null && (gamepad.timestamp > _timestamps[index])) {
       _timestamps[index] = gamepad.timestamp;
       return true;
-    } else if (_.isUndefined(gamepad) && !_.isNull(_timestamps[index])) {
```
yizow, line 193: undefined and null in JS have distinct meanings, i am told. 

Are we sure this change is correct?

yizow (changes requested):

dawn/renderer/utils/sagas.js
kevinji: I think so; [`navigator.getGamepads()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getGamepads) returns an array of gamepads, where the first is `null` (which isn't being handled correctly right now). I see no reason why any of them should be `undefined`.

yizow (approved):

dawn/renderer/utils/sagas.js
vdonato: not a big deal, but you can also just get away with checking `if (gamepad && ...)`
assuming `typeof gamepad === 'object'` (and I'd be pretty worried if that weren't always true)

same comment with the line below. `else if (!gamepad && _timestamps[index])` is fine unless `_timestamps[index]` could be 0, which seems practically impossible.

PR #209, opened 2017-03-07
Add disable_device() function skeleton to selected devices

dHutchings:
These devices had their disable_device() function handle removed due to
git merge whackyness.


akumaraguru (approved):
SGTM

PR #210, opened 2017-03-07
[RUNTIME] Share student state

bodiddles:
-Changes behavior so that if studentCode crashes, we change the local state of the UDP recv process
which we use for parsing when to send change studentCode state commands.

-Adds a try catch for the main while loop of stateManager.


yizow (approved):
lgtm

runtime/testy/stateManager.py
```diff
@@ -40,7 +40,8 @@ def makeCommandMap(self):
       SM_COMMANDS.SEND_ADDR: self.send_addr,
       SM_COMMANDS.ENTER_IDLE: self.enter_idle,
       SM_COMMANDS.ENTER_TELEOP: self.enter_teleop,
-      SM_COMMANDS.ENTER_AUTO: self.enter_auto
+      SM_COMMANDS.ENTER_AUTO: self.enter_auto,
+      SM_COMMANDS.END_STUDENT_CODE: self.endStudentCode
```
brandonxxlee: Add a comma at the end of every entry

brandonxxlee (approved):
Mainly LGTM, Need to be careful that when we update protos we make sure to update all enums correctly.

PR #213, opened 2017-03-07
[RUNTIME] Have studentAPI only pull from SM once per iteration

brandonxxlee:
studentAPI now pulls entire gamepad and sensor dictionary every loop, and they uses a local copy
to access values.


runtime/testy/studentAPI.py
```diff
@@ -86,11 +90,15 @@ def __init__(self, toManager, fromManager):
     super().__init__(toManager, fromManager)
     self._createSensorMapping()
     self._coroutines_running = set()
+    self._get_all_sensors()
+
+  def _get_all_sensors(self):
+    self.sensors = self._getSMValue('hibike', 'devices')
```
yizow: I suggest a rename to `self.peripherals`

runtime/testy/studentAPI.py
```diff
@@ -64,9 +64,13 @@ class Gamepad(StudentAPI):
 
   def __init__(self, toManager, fromManager):
     super().__init__(toManager, fromManager)
+    self._get_gamepad()
+
+  def _get_gamepad(self):
+    self.all_gamepads = self._getSMValue("gamepads")
```
yizow, line 70: i think we need `[0]` on the end of this line

yizow (changes requested):

runtime/testy/studentAPI.py
brandonxxlee: The current code on master doesn't have one, and I don't think we need one? What would the 0 be for.

yizow (approved):

PR #215, opened 2017-03-08
[HIBIKE] Removed inversion of limit switch value

akumaraguru:
Doug mentioned that the limit switch value was inverted in Dawn, i.e. reading 1 on an open switch and 0 on a closed switch.

Looks like the firmware does a boolean inversion of the value, which has been removed.


hibike/devices/LimitSwitch/limit_switch.cpp
```diff
@@ -45,7 +45,7 @@ uint8_t device_read(uint8_t param, uint8_t* data, size_t len) {
   if (MAX_PAYLOAD_SIZE - len < sizeof(uint8_t) || param >= NUM_SWITCHES || param<0) {
     return 0;
   }
-  data[0] = digitalRead(pins[param]) == LOW;
+  data[0] = digitalRead(pins[param]);
```
dHutchings: Digital Read returns HIGH or LOW, which is enumed to 0 and 1. 

That's not a boolean.  Please compare so that this returns a boolean, or change the spec so the paramter expected is an int.

dHutchings (changes requested):
Please don't typecast when sending things into hibike.

hibike/devices/LimitSwitch/limit_switch.cpp
nikitakit: Note that making the param an int will impact the way it's presented to the students. It has generally been our policy to convert params to accessible units in firmware (i.e. analog sensors are returned as floats and not integers from 0 to 1024). So I wouldn't recommend making it an int.

akumaraguru: Added back the comparison

dHutchings (approved):

PR #217, opened 2017-03-08
[HIBIKE] Add deadband feature

JHGZhou:


hibike/README.md
```diff
@@ -223,6 +223,7 @@ Device Type Enumeration:
 |         |                | 12           | enc_pos             | float      | yes   | yes    |
 |         |                | 13           | enc_vel             | float      | yes   | no     |
 |         |                | 14           | motor_current       | float      | yes   | no     |
+|         |                | 15           | deadband            | float      | no    | yes    |
```
yizow: i feel this should be both read and writeable.

hibike/devices/YogiBear/YogiBear.cpp
```diff
@@ -240,6 +245,9 @@ uint8_t device_read(uint8_t param, uint8_t* data_update_buf, size_t buf_len) {
       return sizeof(float);
       break;
 
+    case DEADBAND:
```
yizow, line 219: add ability to read.

hibike/devices/YogiBear/motor.cpp
```diff
@@ -4,6 +4,7 @@
 //tab to handle all controls issued to the motor including driving and braking
 
 bool motorEnabled = false;
+float deadBand = 0;
```
yizow: can we default this to something reasonable. 
lets say .05?

hibike/devices/YogiBear/motor.cpp
```diff
@@ -64,4 +67,8 @@ void clearFault() {
   digitalWrite(enable_pin, HIGH);
 }
 
+void setDeadBand(float range) {
+  deadBand = range;
+}
+
```
yizow: remove extra newline

hibike/hibikeDevices.json
```diff
@@ -75,7 +75,8 @@
                                                           {"number": 11  , "name": "current_thresh"     , "type": "float"    , "read": false , "write": true  },
                                                           {"number": 12  , "name": "enc_pos"            , "type": "float"    , "read": true , "write": true  },
                                                           {"number": 13  , "name": "enc_vel"            , "type": "float"    , "read": true , "write": false  },
-                                                          {"number": 14  , "name": "motor_current"      , "type": "float"    , "read": true , "write": false  }
+                                                          {"number": 14  , "name": "motor_current"      , "type": "float"    , "read": true , "write": false  },
+                                                          {"number": 15  , "name": "deadband"           , "type": "float"    , "read": false , "write": true  }
```
yizow: let deadband be both read and write

hibike/hibike_tester.py
```diff
@@ -29,5 +30,15 @@ def disable(self):
 if __name__ == '__main__':
     h = Hibike()
     h.enumerate()
+    uid = h.stateQueue.get()[1][0]
```
yizow: remove these changes. 
hibike_tester should be general, not specific to a device.

yizow (changes requested):
mainly curious why the deadband isn't readable.

hibike/devices/YogiBear/motor.cpp
```diff
@@ -38,13 +39,15 @@ float readCurrent() {
...
+  if (target < -deadBand) { 
     digitalWrite(INA, LOW);
     digitalWrite(INB, HIGH);
     target = target * -1;
-  } else if (target > 0) {
+  } else if (target > deadBand) {
     digitalWrite(INA, HIGH);
     digitalWrite(INB, LOW);
+  } else {
+    target = 0;
```
yizow, line 59: should we call motordisable here?

hibike/README.md
JHGZhou: In general, we left values by default not readable, but yea i agree to make deadband readable. Will make necessary changes for this.

hibike/devices/YogiBear/YogiBear.cpp
JHGZhou: see first comment

hibike/devices/YogiBear/motor.cpp
JHGZhou: Sounds reasonable to me. I believe doug wanted it to default to 0, but I think in this case we will change it to 0.05 so that by default the controller in its resting position won't drive the motor because of a small tilt.

hibike/devices/YogiBear/motor.cpp
JHGZhou: done

hibike/hibikeDevices.json
JHGZhou: see first comment

hibike/hibike_tester.py
JHGZhou: sorry forgot about that file, done now

hibike/devices/YogiBear/motor.cpp
JHGZhou: no, motor disable resets all values and states and should only be used when hibike issues a disable command. In this case, we want the robot to continue its operation as it was before stopping so we will just set the PWM to 0

hibike/hibike_tester.py
```diff
@@ -1,4 +1,5 @@
 import hibike_process
+import time
```
yizow, line 2: dont think we need this import

dHutchings: we did to test specific functionality.

That functionality has been reverted for the commit.  But given the number of times that we've had to put time.sleep() in the while loop, and given how important time.sleep() is to having loops run at real rates, I suggest we leave this import in.

Possibly even have a time.sleep() in the while true loop to demonstrate its use

JHGZhou: Personally agree with Doug on this, the time import seems useful for testing. For now I have not made a change

yizow:
Once the other yogibear change is in, rebase this off master and then ping me

yizow (approved):

PR #221, opened 2017-03-09
[Dawn] Fixes gamepad null checks

bodiddles:
Reverts changes for null checking gamepads so dawn doesn't error on build.


dawn/renderer/utils/sagas.js
```diff
@@ -187,10 +187,10 @@ const _timestamps = [0, 0, 0, 0];
 
 function _needToUpdate(newGamepads) {
   return _.some(newGamepads, (gamepad, index) => {
-    if (gamepad !== null && (gamepad.timestamp > _timestamps[index])) {
+    if (!_.isUndefined(gamepad) && (gamepad.timestamp > _timestamps[index])) {
```
kevinji: Try `gamepad != null` here (should warn against both `null` and `undefined`).

dawn/renderer/utils/sagas.js
```diff
@@ -187,10 +187,10 @@ const _timestamps = [0, 0, 0, 0];
...
       _timestamps[index] = gamepad.timestamp;
       return true;
-    } else if (gamepad === null && _timestamps[index] !== null) {
+    } else if (_.isUndefined(gamepad) && !_.isNull(_timestamps[index])) {
```
kevinji: Try `gamepad == null && _timestamps[index] !== null` here (again, the `gamepad` check should warn against both `null` and `undefined`, and the RHS check is the same).

kevinji (changes requested):
These changes should be a bit more robust against weird `null` issues.

kevinji (approved):

PR #222, opened 2017-03-09
[RUNTIME] Re-enable Student Error Printing

bodiddles:
Re-enable console logging for just errors from studentCode but still does not support regular printing from studentCode.


brandonxxlee (approved):
LGTM

PR #225, opened 2017-03-09
Re-enable student code upload

kevinmasd:
correctText moved to static, so Dawn was actually throwing errors rather than creating valid ssh sessions.


ArjunSridhar1 (approved):
nice job

kevinji (approved):
Thanks for fixing my bugs

PR #236, opened 2017-03-09
Add better handling of unable to open com ports

dHutchings:


brandonxxlee (approved):
LGTM

karthik-shanmugam:
lol you forgot to actually use working_ports instead of ports when spawning the threads. I think this will actually crash now because it tries to zip serials with ports.

Issue #115, opened 2017-02-09
[Dawn] Status Label + Battery Status


kev-chien:
As title. Was implemented (I think) last year.

kevinmasd:
Done in #241

PR #241, opened 2017-03-10
[Dawn] Nav Bar, Status Label, React Joyride Fix

### LINKED ISSUE #106: [Dawn] React-Joyride
React-Joyride, the interactive tours mechanism, still has issues loading in Dawn.

Also think of new information to show to students.

kev-chien:
PR Changes
- Reimplemented Nav Bar + Status Label from old dawn, yay pretty things (addresses #115 )
- Moved UpdateBox and ConfigBox buttons to DNav top right corner
- Fixed React Joyride and added steps (fixes #106 )


kevinji (commented):
Hmm, the `.babelrc` changes shouldn't be necessary... do you know what the issue was? And were your `node_modules` up to date? I bumped up a few versions and added a few dependencies.

kev-chien:
**Nvm, issue was indeed bc I did not run npm install, rolled back changes.**

dawn/renderer/components/Editor.js
```diff
@@ -300,136 +284,107 @@ class Editor extends React.Component {
...
+            <DropdownButton
+              title="Theme"
+              bsSize="small"
+              id="choose-theme"
+            >
+              {_.map(this.themes, (theme, index) => (
+                <MenuItem
+                  active={theme === this.props.editorTheme}
+                  onClick={_.partial(this.changeTheme, theme)}
+                  key={index}
```
kevinji: Don't use `index` as a key; use the theme name or something unique per theme instead. (I believe React also complains in the console about this issue.)

.gitignore
```diff
@@ -393,3 +393,6 @@ $RECYCLE.BIN/
 
 # Hibike dynamic files
 hibike/virtual_devices.txt
+
+# Webstorm project settings
+.idea
```
kevinji: Missing a newline after.

dawn/renderer/components/DNav.js
```diff
@@ -0,0 +1,147 @@
...
+
+class DNav extends React.Component {
+  static saveAddress(currentAddress) {
+    const prompt = smalltalk.prompt(
+      'Enter the IP address of the robot:',
+      'Examples: 192.168.13.100, 127.0.0.1',
+      currentAddress,
+    );
+    prompt.then((value) => {
+      if (/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/.test(value)) {
```
kevinji: There's already a `getValidationState` (yes, this should be given a better name like `validateIP`, feel free to rename) in renderer/utils/utils.js which has a more precise regex.

dawn/renderer/components/DNav.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  updateAddress() {
+    storage.has('runtimeAddress').then((hasKey) => {
+      if (hasKey) {
+        storage.get('runtimeAddress').then((data) => {
+          this.saveAddress(data.address);
+        });
+      } else {
+        this.saveAddress('192.168.13.100');
```
kevinji: Not sure what relevance this specific IP address has, but it should probably be moved to a `const` at the top of the file. (Also, there's another default in ConfigBox's placeholder *and* one in Ansible as `DEFAULT_IP`... so more consolidation would be good?)

dawn/renderer/components/Editor.js
```diff
@@ -125,7 +117,7 @@ class Editor extends React.Component {
   }
 
   getEditorHeight(windowHeight) {
-    return `${String(windowHeight - 160 - (this.props.showConsole * (this.consoleHeight + 40)))}px`;
+    return `${String(windowHeight - 231 - (this.props.showConsole * (this.consoleHeight + 40)))}px`;
```
kevinji: If possible, maybe give this magic number a name and turn it into a `const`?

kevinji (changes requested):
Also, not sure if you've looked at `ConfigBox` (which should honestly also be renamed), but there's some existing logic in `disableUploadUpdate` that should probably be revisited. My guess is that you can remove the `this.props.runtimeStatus || this.props.isRunningCode` section, as `DNav` should already take care of whether to show/enable the TooltipButton or not.

kev-chien:
Thanks for the reviews, I'll try to address them in the morning haha

kevinji (approved):

kevinji:
Looks okay to me; would still wait for @kevinmasd to review this as well.

nikitakit:
I just sent off a build and tried to test this code with an actual robot.

The peripherals display in Dawn has the following issues:
* I only displays a single device (I'm looking into whether this is a Dawn bug or a runtime bug)
* __It does not display UIDs for devices__. Without knowing the UIDs, students will be unable to write working code.

I understand that these issues are orthogonal to the very-much-necessary changes in this PR, but please ensure we have a working version of Dawn ready asap.

PR #243, opened 2017-03-10
[Dawn] Display UID instead of Names; Throw Redirection

kevinmasd:
UID are now displayed in replacement of NameEdit, due to N.E's broken nature currently. Throws are also sent into stdout rather than through blocking modal windows. There are also fixes for multi-param values in peripheral outputs.


dawn/renderer/components/UpdateBox.js
```diff
@@ -46,44 +46,47 @@ class UpdateBox extends React.Component {
...
+                      });
                     }
+                    stream.write(`${defaults.PASSWORD}\n`);
+                    stream.on('exit', (code) => {
+                      if (code === 0) {
+                        conn.end();
+                      }
+                    });
+                    console.log('Hello');
+                    this.setState({ isUploading: false });
```
yizow: does this need to be inside the `exit` block?

yizow (changes requested):

yizow:
I think a test was hardcoded to expect 1000 ms timeout from runtime. we should change to use the new timeout global.

yizow (approved):

PR #246, opened 2017-03-11
[DevOps] Added student/field wifi switching

Alramech:


yizow:
Not actually working?

yizow (changes requested):
Blocking until fixed

jonathan-j-lee:
Fixed in #320.

PR #248, opened 2017-03-11
[HIBIKE] MAX_PAYLOAD_LEN deleted from device_read comparisons

tingtingchiang:
Similar to recent PDB pull request, comparison used to assume that len was amount already written in the buffer rather than amount left to be written to. Changed in LineFollower and LimitSwitch.


karthik-shanmugam (approved):
LGTM as long as this has been tested to work lol

PR #250, opened 2017-03-11
[RUNTIME] Adds support for disabling Hibike

bodiddles:
Adds commands for Hibike to enter disabled mode whenever the the robot goes into idle mode or when emergency stop is called.


brandonxxlee (approved):
LGTM

PR #264, opened 2017-03-11
[DAWN] Fix non-general param handling

nikitakit:
Dawn would error if the battery buzzer did not report exactly the
two params that were expected from it.


bodiddles (approved):
LGTM

PR #269, opened 2017-03-16
[HIBIKE] Servo was improperly merged. Fixed and reintegrated

JHGZhou:


hibike/hibike_tester.py
```diff
@@ -30,5 +30,13 @@ def disable(self):
 if __name__ == '__main__':
     h = Hibike()
     h.enumerate()
```
yizow: Testing file changes?

hibike/devices/ServoControl/servo_control.cpp
```diff
@@ -20,12 +21,6 @@ void enable(int num) {
   servos[num].attach(pins[num]);
 }
 
-void setup() {
-  hibike_setup(500); // Time in milliseconds before timeout on heartbeat
-  disableAll();
```
yizow, line 25: Are we removing disable all?

yizow (changes requested):

hibike/hibike_tester.py
JHGZhou: Oops, reverted to original state

hibike/devices/ServoControl/servo_control.cpp
JHGZhou: Added disable all back in

yizow (approved):

PR #270, opened 2017-03-16
Dawn: Warn against turning off robot during Runtime update

kevinmasd:
Handles #266

PR #272, opened 2017-03-16
[Dawn] Update Joyride for new upload situation

kevinmasd:


nikitakit:
Does Dawn give a signal when it's safe to restart the robot? It might be better to say "please don't restart your robot. When X happens you know that the update has succeeded"

kevinmasd:
Handling that in #270, in that we warn students to not restart the robot. As for checking if the update succeeded, then can look at the version numbering.

yizow (approved):

PR #276, opened 2017-03-18
Hibike/arduino script

tingtingchiang:
Modified setup.sh to extract and move the arduino library from given (command-line argument) path to /opt. Also modifies Makefile to look for the arduino library in /opt, specifically Arduino version 1.8.1. This means that when this is merged, people flashing must first download the .tar.xz (if they do not have it already) at https://www.arduino.cc/en/Main/Software and run setup.sh with the first command line argument as the path to the .tar.

Also fixed a setup.sh spacing error.


yizow:
You should put that link to the download to be printed if the script errors and can't find the arduino files.

tingtingchiang:
Hmm. Clarification: The script "errors" when the path file is not accurate, not when there isn't even an optional argument right?

hibike/setup.sh
```diff
@@ -9,6 +9,13 @@ cp utils/flash /etc/bash_completion.d/flash.sh
 echo "Setup Tab-complete"
 echo "Please restart your terminals"
 
+echo "Update Arduino library: First command line argument as file path to .tar (or .tar.xz)"
+echo "Update to Arduino 1.8.1"
+if [ $# -eq 1 ] && [ $(grep -c -i "arduino" <<< $1) != 0 ] && [ $(grep -c ".tar" <<< $1) != 0 ]; then
```
yizow: Use regex for the grep
.*arduino.*\.tar\.xz
Should work

hibike/setup.sh
```diff
@@ -9,6 +9,13 @@ cp utils/flash /etc/bash_completion.d/flash.sh
 echo "Setup Tab-complete"
 echo "Please restart your terminals"
 
+echo "Update Arduino library: First command line argument as file path to .tar (or .tar.xz)"
```
yizow: Do we want to print this all the time?

yizow (changes requested):

yizow:
Ah I see the library is optional
Do we want to tell the user somewhere where to get it from?

hibike/setup.sh
tingtingchiang: Are you saying $(grep -c -i "arduino..tar.xz" <<< $1)  will work?

yizow: Nikita caught this. 
http://stackoverflow.com/questions/16045139/redirector-in-ubuntu

oh oops my regex didn't format correctly. 
`.*arduino.*\.tar\.xz`

grep uses regex to look for things

hibike/setup.sh
tingtingchiang: How else should we remind/prompt user to update the Arduino library?

hibike/setup.sh
```diff
@@ -9,6 +9,17 @@ cp utils/flash /etc/bash_completion.d/flash.sh
 echo "Setup Tab-complete"
 echo "Please restart your terminals"
 
+sudo usermod -a -G dialout $(whoami)
+
+if [ $# -eq 1 ] && [ $(echo $1 | grep -c -i ".*arduino.*\.tar") != 0 ]; then
```
yizow: this will not match .tar.xz
`".*arduino.*(\.tar|\.tar\.xz)"` will

Either change it here, or in the print statement below (by removing the .tar.xz)

yizow (approved):

PR #279, opened 2017-03-22
[RUNTIME] Change runtime to pass linter

brandonxxlee:
Pylint has been added to travis, and the main runtime files (runtime.py, stateManager.py, studentAPI.py, Ansible.py) have been fixed to pass the linter

There are quite a few manually disabled warnings that should be fixed

I tried really hard to make sure nothing broke and only style was changed. 

This PR is mainly to show what it looks like as of now, and discuss how we can improve it.


brandonxxlee:
I don't even know if this is something we want to deploy this season, in case something breaks, but I guess that is up for discussion.

kevinji:
Can you modify [runtime/.editorconfig](https://github.com/pioneers/PieCentral/blob/master/runtime/.editorconfig) to use 4 spaces instead of 2?

yizow:
Does "tried really hard to make nothing broke" include trying to drive a robot with this code/ upload code/ upgrade runtime?

yizow (approved):
I think this is far too big to parse every line, but it still drives a robot so....

Issue #277, opened 2017-03-21
[Hibike] Wrong Gitignore.


dHutchings:
the hibike/lib folder is gitignored.

I think that someone added 'lib' to the gitignore for their own project, and it accidentally also gitignored the hibike lib folder.

PR #289, opened 2017-03-23
[Hibike] .gitignore allows lib

kevinmasd:
Addresses #277


kevinji:
Maybe it'd make more sense to just whitelist it in [Hibike's .gitignore](https://github.com/pioneers/PieCentral/blob/master/hibike/.gitignore)?

yizow (approved):

PR #290, opened 2017-03-23
[DEVOPS] Add a slack notifier

onibrow:
This should start pinging build statuses to #things_broke channel in slack


onibrow:
Please and thank.

kevinji:
Is this value [encrypted](https://docs.travis-ci.com/user/notifications/#Configuring-slack-notifications)? If not, you should encrypt and force push over this commit.

onibrow:
@mc10 How about now? I think I used the token correctly.

.travis.yml
```diff
@@ -52,3 +52,9 @@ deploy:
   on:
     tags: true
     condition: $DEPLOY_ARTIFACTS = 1
+
+notifications:
+  slack:
+    rooms:
+      - secure: "4NsPGDKKtiUSdqjl3bxUMlbw"
+    on_success: always
```
kevinji: I think you want `on_failure: always` if this bot's supposed to ping when things break. Also, perhaps add `on_pull_requests: false` to avoid spurious notifications for PRs, and `on_success: change` for when builds change from failing to passing?

.travis.yml
```diff
@@ -52,3 +52,9 @@ deploy:
   on:
     tags: true
     condition: $DEPLOY_ARTIFACTS = 1
+
+notifications:
+  slack:
+    rooms:
+      - secure: "4NsPGDKKtiUSdqjl3bxUMlbw"
```
kevinji: Really minor nit: since the hyphens aren't indented in the rest of this file, you probably want to unindent this one level.

kevinji (dismissed):
Mainly just comments about notification settings, but everything else looks good to me.

.travis.yml
onibrow: I want it ping for every single push, so the pusher gets notified immediately if it fails. I've fixed the indent issue and added the on_failure line.

brandonxxlee (approved):
lgtm

PR #291, opened 2017-03-24
[HIBIKE] switch communication protocol from serial to HID

karthik-shanmugam:
- replace serial communication with hid in python and firmware
- completely switch `hibike_process.py` to support hid only
- add [this HID library](https://github.com/NicoHood/HID) to the hibike/lib folder
- add a C++ preprocessor variable to switch firmware between HID and serial mode at compile time
  - hard-code it to HID mode for now
- add a test script to enumerate all devices using HID
  - does not use `hibike_process`
- minor changes/fixes
  - `hibike_process`'s `write_thread` now flushes its queue when it gets too large
    - it will never flush a `disable` request though
  - example devices no longer continually increment two of their params
    - these are now for testing writing
  - `hibike_process` now keeps count of packets read/second and average time spent parsing packets


hibike/Makefile
```diff
@@ -18,13 +18,13 @@ ARDMK_DIR         = $(PROJECT_DIR)/Arduino-Makefile
 
 ### ARDUINO_DIR
 ### Path to the Arduino application and ressources directory.
-ARDUINO_DIR       = /usr/share/arduino
+ARDUINO_DIR       = /opt/arduino-1.8.1
```
dHutchings: Doesn't a different, already merged PR, take care of this?

dHutchings: Doesn't a different, already merged PR, take care of this?

hibike/Makefile
```diff
@@ -68,7 +68,7 @@ endif
 RANDOM := 0x$(shell head -c 8 /dev/urandom | xxd -p)
 CXXFLAGS += -D UID_RANDOM=$(RANDOM)
 
-CXXFLAGS += -D DESCRIPTOR=$(shell python escape_json.py $(PROJECT_DIR)/devices/$(DEVICE)/*.json)
+# CXXFLAGS += -D DESCRIPTOR=$(shell python escape_json.py $(PROJECT_DIR)/devices/$(DEVICE)/*.json)
```
dHutchings, line 72: Is this a change needed for this device only, or for all makefiles in the future?

If it's a generic change for all devices, make a seperate branch and PR it in seperately.

hibike/devices/ExampleDevice/example_device.cpp
```diff
@@ -38,8 +38,8 @@ void loop() {
   // so exessive delays here will affect hibike.
   //don't change Kumiko.
 
-  hazuki += 0;
-  sapphire += 1;
+  // hazuki += 0;
```
dHutchings, line 41: hm?  I'd prefer to see these lines uncommented again.

hibike/hibike_message.py
```diff
@@ -57,9 +57,11 @@
   "GenericError"        : 0xFF
 }
 
+MAX_PACKET_SIZE = 64
```
dHutchings, line 60: is MAX_PACKET_SIZE really 64?

I agree that MAX_PAYLOAD_SIZE is 64.  (4 bytes per float, 16 maximum variables, 4*16 = 64).  However, wouldn't that mean that MAX_PACKET_SIZE = Message ID (1) + Payload Length (1) + max payload (64) +  Cheksum (1) = 67.

Or is it 68, because there's then a '0' added as a packet deliminator (cobs encoding).

Does this figure also include COBs encoding?  The packet size can therefore get much longer if there just happen to be a bunch of 0's than need to be excluded.

hibike/hibike_message.py
```diff
@@ -333,6 +324,17 @@ def parse_bytes(bytes):
     return None
   return HibikeMessage(messageID, payload)
 
+# Sends this message
+# Computes the checksum
+# Then sends each byte of the message, and finally sends the checksum byte
+def send(serial_conn, message):
```
dHutchings, line 330: It's not really a 'serial_conn' anymore, is it?

dHutchings: More importantly.  Why did this block of code move?  Last time, it was located at line 114

hibike/hibike_message.py
dHutchings: Or is this the observed maximum size that our HID library can handle?

hibike/hibike_message.py
```diff
@@ -432,6 +434,43 @@ def read(serial_conn):
 
   return HibikeMessage(messageID, payload)
 
+# Sends this message
+# Computes the checksum
+# Then sends each byte of the message, and finally sends the checksum byte
+def hid_send(hid, message):
```
dHutchings, line 440: As far as I can tell, hid_send is the same as send, (from the python side), except that it can return true/false.

why not unify them?  Or does it have to do with the size check?

hibike/hibike_message.py
```diff
@@ -432,6 +434,43 @@ def read(serial_conn):
...
+  encoded = cobs_encode(m_buff)
+  out_buf = bytearray([0x00, len(encoded)]) + encoded
+  if (len(out_buf) > MAX_PACKET_SIZE):
+    return False
+  else:
+    hid.write(out_buf)
+    return True
+
+# behold the beauty of sending packets over a packet based communication protocol
+def hid_blocking_receive(hid):
```
dHutchings, line 460: why are both blocking_recieve and blocking_recieve_generator implimented?  Do we use both?

hibike/hibike_process.py
```diff
@@ -14,41 +15,18 @@
...
-    for port in ports:
-        try:
-            serials.append(serial.Serial(port, 115200))
-            working_ports.append(port)
-
-            #this implimentation ensures that as long as the cannot open serial port error occurs while opening the serial port, a print will appear, but the rest of the ports will go on.
-
-        except serial.serialutil.SerialException as e:
-            print("Cannot Open Serial Port: " + str(port))
+    devices = hid.open_all()
```
dHutchings, line 19: I approve of this line change.  makes it simpler.

hibike/hibike_process.py
```diff
@@ -14,41 +15,18 @@
...
 
 
     # each device has it's own write thread, with it's own instruction queue
```
dHutchings, line 22: change "it's" to "its".  English FTW.

hibike/hibike_process.py
```diff
@@ -83,32 +61,54 @@ def hibike_process(badThingsQueue, stateQueue, pipeFromChild):
...
+            hm.hid_send(dev, hm.make_disable())
         elif instruction == "heartResp":
             uid = args[0]
-            hm.send(ser, hm.make_heartbeat_response(hm.uid_to_device_id(uid)))
+            hm.hid_send(dev, hm.make_heartbeat_response())
+
+# global counters for tracking statistics
 
+# number of packets read
+ctr = 0
```
dHutchings, line 94: This smells like debugging / testing code that wasn't removed.

hibike/hibike_process.py
```diff
@@ -83,32 +61,54 @@ def hibike_process(badThingsQueue, stateQueue, pipeFromChild):
...
+start_time = time.time()
+# total time spent parsing received packets
+parse_time = 0
 
-def device_read_thread(index, ser, instructionQueue, errorQueue, stateQueue):
+def device_read_thread(index, dev, instructionQueue, errorQueue, stateQueue):
     uid = None
+    global ctr
```
dHutchings, line 101: 'ahem'.  global.  There are N threads, all trying to access the same globals.

We could be causing the slowdown issues here simply by having so many different threads attempt to access the same globals at once.  Remove this.

hibike/hibike_process.py
```diff
@@ -119,12 +119,14 @@ def device_read_thread(index, ser, instructionQueue, errorQueue, stateQueue):
...
                 else:
-                    print("[HIBIKE] Port %s received data before enumerating!!!" % ser.port)
+                    print("[HIBIKE] Device %s received data before enumerating!!!" % dev)
                     print("Telling it to shut up")
-                    hm.send(ser, hm.make_subscription_request(1, [], 0))
+                    hm.hid_send(dev, hm.make_subscription_request(1, [], 0))
             elif message_type == hm.messageTypes["HeartBeatRequest"]:
                 if uid is not None:
                     instructionQueue.put(("heartResp", [uid]))
+            parse_time += time.time() - before_time
```
dHutchings, line 128: Remove tsting code.

hibike/hid.py
```diff
@@ -0,0 +1,102 @@
...
+        raise usb.USBError("Could not find any device (vendor_id={}, product_id={})".format(vendor_id, product_id))
+
+    @classmethod
+    def open_all(cls, vendor_id=VENDOR_ID, product_id=PRODUCT_ID):
+        res = []
+        for args in cls.find_hid_endpoints(vendor_id, product_id):
+            try:
+                res.append(HIDWrapper(*args))
+            except usb.USBError as e:
+                print("WARNING: Skipping device due to error:", e)
```
dHutchings, line 35: Would be nice to know which device... should be possible to extract that information.

hibike/hid.py
```diff
@@ -0,0 +1,102 @@
...
+                    # We'll need the interface number and endpoint address to use
+                    # the device
+                    in_endpoint = interface[0]
+
+                    yield device, interface_num, in_endpoint.bEndpointAddress
+
+    def write(self, data):
+        while True:
+            try:
+                return self.device.ctrl_transfer(0x21, 0x09, 0x0200, self.interface_num, data)
```
dHutchings, line 65: Not wild about using ctrl_transfers for everything.  Well, I guess it makes sense b/c writes from runtime currently aren't batched.

hibike/hid.py
```diff
@@ -0,0 +1,102 @@
...
+            except usb.core.USBError as e:
+                time.sleep(0.01)
+
+    def read(self, maximum_len):
+        return self.device.read(self.endpoint_address, maximum_len)
```
dHutchings, line 70: Not wild about this implimentation.

hibike/hid_enumeration_test.py
```diff
@@ -0,0 +1,31 @@
+import threading
```
dHutchings, line 1: I feel like we need to start organizing our tests better.

Hibike developers have had to delete or revert our tests in the name of keeping few code files in the hibike folder..  We should make a /tests folder or something.  I advocate keeping thsi code around, though.

hibike/lib/hibike/hibike_message.cpp
```diff
@@ -1,5 +1,25 @@
 #include "hibike_message.h"
 
+
+
+uint8_t hid_recv_buffer[COBS_DELIM_BYTES+COBS_LENGTH_BYTES+MAX_PAYLOAD_SIZE+MESSAGEID_BYTES+PAYLOAD_SIZE_BYTES+CHECKSUM_BYTES+COBS_OVERHEAD_BYTES];
```
dHutchings, line 5: Were all these constants somewhere already?

hibike/lib/hibike/hibike_message.cpp
```diff
@@ -8,11 +28,32 @@ uint8_t checksum(uint8_t data[], int length) {
   return chk;
 }
 
+#ifdef HIBIKE_HID_MODE
+int send_message_hid(message_t *msg) {
+  uint8_t data[MAX_PAYLOAD_SIZE+ MESSAGEID_BYTES+PAYLOAD_SIZE_BYTES+CHECKSUM_BYTES];
+  // TODO karthik-shanmugam: check buffer size
```
dHutchings, line 34: todo?

hibike/lib/hibike/hibike_message.cpp
```diff
@@ -26,17 +67,111 @@ int send_message(message_t *msg) {
...
+// convenience function for read_message_hid
+int hid_clear() {
+  size_t hid_len = RawHID.available();
+  for (int i=0;i<hid_len;i++){RawHID.read();}
+  return -1;
+}
+// TODO karthik-shanmugam : debug this and investigate if we can use it
+// this is supposed to be a function that reads directly from the RawHID buffer
+// currently the other function, read_message_hid_stream, which uses RawHID's byte-stream-like API, is the one used
+int read_message_hid(message_t *msg) {
```
dHutchings, line 83: Don't include code that isn't used.

hibike/lib/hibike/hibike_message.h
```diff
@@ -1,13 +1,21 @@
 #ifndef HIBIKE_H
 #define HIBIKE_H
+#define HIBIKE_HID_MODE
```
dHutchings, line 3: let's get some comments going here, since you have lots of #ifdefs going on.  we need to know that this is an important option.

hibike/lib/hibike/hibike_message.h
```diff
@@ -1,13 +1,21 @@
...
 #include "Arduino.h"
 #include "devices.h"
 #include "cobs.h"
 
-#define MAX_PAYLOAD_SIZE    100
+#ifdef HIBIKE_HID_MODE
+#include "HID-Project.h"
+#endif
+
+#define MAX_PAYLOAD_SIZE    60
```
dHutchings, line 12: is it?  I thought it was 4*16 = 64.

hibike/serial_enumeration_test.py
```diff
@@ -0,0 +1,28 @@
...
+read_threads = [threading.Thread(target=print_uid, args=(i, device)) for (i, device) in enumerate(devices)]
+
+for read_thread in read_threads:
+    read_thread.start()
+
+# ping each device
+for device in devices:
+    hm.send(device, hm.make_ping())
+
+# wait until each thread enumerates its device and exits
```
dHutchings, line 26: Seems werid for a test to not have an if name == __main__ method.   Also seems wierd to join read threads?

hibike/setup.sh
```diff
@@ -9,6 +9,17 @@ cp utils/flash /etc/bash_completion.d/flash.sh
 echo "Setup Tab-complete"
 echo "Please restart your terminals"
 
+sudo usermod -a -G dialout $(whoami)
```
dHutchings: Woah, hold on.  This is implimented by another PR.

hibike/setup.sh
```diff
@@ -41,7 +52,7 @@ if [ "$(dpkg-query -W -f='${Status}' python3-pip 2>/dev/null | grep -c "ok insta
     apt-get --assume-yes install python3-pip
 fi
 
-if [ "$(pip3 freeze | grep -c "pyserial")" != 1]; then
+if [ "$(pip3 freeze | grep -c "pyserial")" != 1 ]; then
```
dHutchings: As is this.

dHutchings (changes requested):
Lots of changes & comments.

Some things that should be moved to seperate PR.
Some debugging things that should be removed.
Other issues.

I'd appreciate it if yizhow / nikita could read through it once these are fixed.

hibike/hibike_message.py
karthik-shanmugam: I moved it closer to the read function. This function is meant to use a serial object. There's a separate `hid_send` function for sending over hid. I kept the function name as "send" so it doesn't break any python program that uses this.

hibike/hibike_message.py
karthik-shanmugam: the other send is meant to send over a serial connection.

hibike/hibike_message.py
karthik-shanmugam: one reads a single packet and one continually yields packets and never returns. `hibike_process` uses the generator but the other function may be useful for writing tests that want to receive a finite number of packets.

hibike/hibike_process.py
karthik-shanmugam: It logs the number of packets read. I think this statistic will be useful in the future.

hibike/hibike_process.py
karthik-shanmugam: the slowdowns were definitely present before I added this to debug them. There are no locks on this global; at worst race conditions cause it to fail to increment at times, and at best python itself has some locking mechanism for this but the locks are released after very short increment operations.

hibike/hibike_process.py
karthik-shanmugam: I think these statistics are useful to keep around

hibike/hid.py
karthik-shanmugam: I honestly don't understand how the usb low level stuff works or why we use `ctr_transfer` to write but `read` to read... @nikitakit halp

hibike/lib/hibike/hibike_message.cpp
karthik-shanmugam: no I just added them as part of this pr

hibike/lib/hibike/hibike_message.cpp
karthik-shanmugam: lol this is a todo from last year that got copy pasted

hibike/serial_enumeration_test.py
karthik-shanmugam: the read threads exit as soon as they successfully enumerate their device. The joins are so the test can exit if it enumerates successfully on every port

hibike/Makefile
karthik-shanmugam: github claims there are no merge conflicts with the makefile

hibike/devices/ExampleDevice/example_device.cpp
karthik-shanmugam: I stopped two of the parameters from auto-incrementing so they can be used to test writing instead. `hibike_process.py` will now write to these two parameters if you run it directly

hibike/hibike_message.py
karthik-shanmugam: I vaguely remember @nikitakit mentioning 64 bytes was some kind of limit but I may be wrong. If we wanted to support a 64 byte payload we need 1 + 1 + 1 + 1 + 64 + 1 + 1 = 70 bytes. This is because cobs encoding adds a 0 byte, a redundant length field, and 1 byte of overhead.

hibike/hibike_process.py
karthik-shanmugam: also breaks support for virtual devices ;_;

hibike/hid.py
karthik-shanmugam: I'm not sure you'll be able to extract any useful information. At this point all we have are the low level pyusb objects. What information do those contain?

hibike/hid.py
karthik-shanmugam: I can't think of a better workaround without a better understanding of how usb works.

hibike/setup.sh
karthik-shanmugam: This is because we cherry picked some commits from the branch that pr came from. What's the best way to resolve these conflicts now?

hibike/setup.sh
karthik-shanmugam: This is because we cherry picked some commits from the branch that pr came from. What's the best way to resolve these conflicts now?

hibike/hibike_process.py
dHutchings: Well then. Shouldn't that mean that this commit also fixes virtual devices so they work again?

(Pottery Barn Rule)

hibike/Makefile
dHutchings: Huh.  Ok then.

hibike/devices/ExampleDevice/example_device.cpp
dHutchings: Not sure how I feel about leaving it this way.

Espescially since we've decided that running hibike_process.py as a tester is deprecated.

hibike/hibike_message.py
dHutchings: 64 bytes is the 'limit' in that back when we were testing arduinos w/ c libraries on the computer, we couldn't get more than 64 bytes (to the arduino) working.

No one tested it now that we're in python.  It may have been a c library problem, and seeing as we've now removed the library, we should test it.

hibike/hibike_message.py
dHutchings: Oic.  git diff :(  sgtm.

hibike/hibike_message.py
dHutchings: See my above comment.  Untill it's tested, I'm not convinced that there should be a packet limit size anymore.

hibike/hibike_process.py
dHutchings: .... I disagree with leaving so much testing code in a PR to master.

hibike/hibike_process.py
dHutchings: See my above comment.  I disagree with leaving so much testing code in a PR to master.

hibike/hibike_process.py
dHutchings: see above

hibike/hid.py
dHutchings: http://www.beyondlogic.org/usbnutshell/usb4.shtml

I doubt we can get to this immediately w/out some serious work (and open question if it'll help).  I'm fine with this particular comment being ignored moving fwd.

hibike/lib/hibike/hibike_message.cpp
dHutchings: kk

hibike/lib/hibike/hibike_message.cpp
dHutchings: remove it/?

hibike/hid.py
nikitakit: @dHutchings What are you concerned about? I see no problem here.

This class is a wrapper around PyUSB that abstracts usb-specific functionality (such as endpoint numbers). And calling `device.read` is exactly how you read from HID. No idea what else could be done

hibike/hid.py
nikitakit: Using `self.device.write` here doesn't appear to be an option because the HID device only presents an IN endpoint, but no OUT endpoint. The C demo code you had handled both cases depending on whether an OUT endpoint was present. Seeing as the arduino code doesn't appear to present an OUT endpoint, I only ported the ctrl_transfer version to python. I'm not familiar enough with the embedded side to know if this is easy to change or not.

yizow:
Rebase off master before pull requesting, as per instructions in the master PieCentral readme

yizow:
And i have rebased.

hibike/hid.py
karthik-shanmugam: I think the comment was in reference to line 67

hibike/hibike_message.py
nikitakit: I believe the arduino RawHID library has a 64-byte payload limit, which is where this comes from. It may be possible that HID supports bigger payloads, but that would need to be tested. And I definitely remember seeing something like `#define MAX_LENGTH 64` in the RawHID source code.

dHutchings: You're right that there's a bit limit.  Unfortunately, it's unidirectional.  (We tested this last night, using pyusb).

There's a 63 byte limit in packets from the HID Smart Device to the BBB.  From the BBB to the HID smart device, there seems to be no practical limit in packet size.  (We tested this up to 100 bytes).

Therefore, this limit needs to go away, and a similar limit needs to be implimented on the embedded side.

hibike/hibike_message.py
dHutchings: See above.  The packet size limit is in the _other_ way

hibike/hibike_process.py
dHutchings: Additional comment: we've determined that print statements can slow robots down.  Please, remove this code.

hibike/hid.py
dHutchings: After more testing in 101, I'm not sure if this is needed.

We've run limit_flag.py and runtime on the robot.  The queue tends to overflow when the robot writes many device parameter changes at one time.  For example:
        comms.write(team_flag, [("led1", True),("led2",False),("led3",True),("led4",False)])
sends 1 packet setting 4 parameters, whereas
        comms.write(team_flag, [("led1", True)]
        comms.write(team_flag, [("led2", False)]
        comms.write(team_flag, [("led3", True)]
        comms.write(team_flag, [("led4", False)]

writes 4 packets immediately after one another.  The system is much more likely to buffer overflow in the second case.  We actually ran limit_flag.py without having to drain the buffer at all, but this functionality should still be supported.

dHutchings: Additionally, this discovery may provoke a runtime update that changes functionality to batch writes.

dHutchings:
We can close this issue for the time being.  HID is not being deployed.

PR #292, opened 2017-03-28
[DEVOPS] Add screenfetch, help menus, and aliases to chrommunal setup scripts

onibrow:
screen fetch and help menu


DevOps/frankfurter/chrommunal/bashrc
```diff
@@ -6,3 +6,11 @@ fi
 alias dawn='sudo startxiwi -n trusty /opt/dawn/dawn'
 alias term='sudo enter-chroot'
 alias ubuntu='sudo startunity -X xorg'
+alias screenfetch='/./usr/local/bin/screenfetch-dev -E'
```
ArjunSridhar1: Why is there a '/' before the './'? 
Is this correct?

onibrow: Eeerrrr its been too long, I'll look at this sometime during the Summer.

ArjunSridhar1 (approved):

DevOps/frankfurter/chrommunal/bashrc
rmandla: I don't think there is supposed to be a /

rmandla: nvm, it's fine. but i don't think that file is executable, so you got to run it with bash. Also, when i tried to run it on my computer, I had to use sudo. Don't know if this is true for chromebooks

kevinji: If `/` should be there, the `./` is definitely redundant.

PR #298, opened 2017-04-01
Slowed down heartbeat times to lighten load on hibike python stack.

dHutchings:
Slowed to every 200 ms, and timeout after 1 sec.
Was every 100ms, timeout after 0.5 sec.


dHutchings:
Added new functionality so that there's a default timeout.

Tested to work in both default & non-default modes.

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -2,19 +2,27 @@
 
 message_t hibikeBuff;
 uint16_t params, old_params, subDelay;
-uint32_t value, disable_latency, heartbeatDelay, prevHeartTimer;
-uint64_t prevSubTime, currTime, sent_heartbeat, resp_heartbeat;
+
+//Heartbeat settings.  
+uint32_t disable_latency, heartbeat_Delay;
+uint32_t default_disable_latency = 1000; //number of ms without heartbeat untill disable.  default
+uint32_t default_heartbeat_Delay = 200; //default time between sending heartbeat requests.
```
nikitakit: I think you can should mark these as `const`

hibike/lib/hibike/hibike_device.cpp
```diff
@@ -2,19 +2,27 @@
 
 message_t hibikeBuff;
 uint16_t params, old_params, subDelay;
-uint32_t value, disable_latency, heartbeatDelay, prevHeartTimer;
-uint64_t prevSubTime, currTime, sent_heartbeat, resp_heartbeat;
+
+//Heartbeat settings.  
+uint32_t disable_latency, heartbeat_Delay;
```
nikitakit: if you're going to be renaming the variables, could you standardize the capitalization in `heartbeat_Delay`. Either camelCase or snake_case is fine, but a mix seems odd.

nikitakit (changes requested):

hibike/lib/hibike/hibike_device.cpp
dHutchings: done.

hibike/lib/hibike/hibike_device.cpp
dHutchings: done.

nikitakit (approved):
diff lgtm (but please makes sure that this compiles and runs on hardware)

dHutchings:
It does.  Thank you!

PR #302, opened 2017-04-05
[HIBIKE] Batches device value updates

yizow:
All read threads first copy their data into a batching thread. The
batching thread sends using interprocess communication to Runtime at 25
Hz


hibike/hibike_process.py
```diff
@@ -126,6 +130,11 @@ def device_read_thread(index, ser, instructionQueue, errorQueue, stateQueue):
                 if uid is not None:
                     instructionQueue.put(("heartResp", [uid]))
 
+def batch_data(data, stateQueue):
+    while True:
+        time.sleep(.04)
```
nikitakit: This is a hardcoded constant hiding here. Moreover, the value of this constant has impact on performance.

It might be good to have a comment explaining why batch reads are required, and some brief tips on how this constant value was determined.

nikitakit (approved):
@yizow I think you should address my comment earlier, but otherwise this looks good.

PR #307, opened 2017-04-08
[DevOps] Moved and updated chrommunal

jonathan-j-lee:
* The `chrommunal` directory is now in the `DevOps` directory, not `DevOps/frankfurter`.
* The URLs for Dawn and `chrommunal` resources have been updated for this year.


DevOps/chrommunal/setup-ubuntu.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-DAWN_URL=https://storage.googleapis.com/pie-software-builds/latest/dawn-linux-x64.zip
+DAWN_URL=https://github.com/pioneers/PieCentral/releases/download/test%2Fdawn%2F2017-03-11c/dawn-linux-x64.zip
```
ArjunSridhar1: This appears to be a change in functionality. Maybe we should set up some sort of url forwarding on the website so we can have a download link to the latest dawn release.

jonathan-j-lee: You're right. Thanks for catching that.

ArjunSridhar1 (approved):

PR #309, opened 2017-04-08
[Dawn] Peripheral Display Filter

kevinmasd:
TeamFlag has been edited out of peripheral display. It had been conflicting with any other sensor in the default case (i.e. not in `cleanerNames`).

Additionally, can somebody check if all the displayable peripherals are enumerated in `cleanerNames`?


yizow:
I'm not quite sure what this is supposed to do.
Currently, team flag does not show up as a sensor.

This pull request... also makes it not show up?

kevinmasd:
What's happening currently is that the accordion/panel for ExampleSensor and TeamFlag are conflicting one another, since Dawn doesn't recognize either. As a result, TeamFlag would never be seen because Electron only displays the panel that came earliest (alphabetically being ExampleSensor).

dawn/renderer/components/PeripheralList.js
```diff
@@ -18,13 +18,18 @@ cleanerNames[PeripheralTypes.ServoControl] = 'Servo Controllers';
 cleanerNames[PeripheralTypes.RFID] = 'RFID';
 cleanerNames[PeripheralTypes.YogiBear] = 'Yogi Bear';
 
+const filter = [PeripheralTypes.TeamFlag];
```
kevinji: Can you make this a set (`new Set([PeripheralTypes.TeamFlag])`)?

dawn/renderer/components/PeripheralList.js
```diff
@@ -18,13 +18,18 @@ cleanerNames[PeripheralTypes.ServoControl] = 'Servo Controllers';
...
+
 const handleAccordion = (array) => {
   const peripheralGroups = {};
+
   array.forEach((p) => {
```
kevinji: You can extract the filter includes check with
`array.filter(p => !filter.has(p.device_type)).forEach(p => /* ... */)`
assuming you converted the filter to a Set.

kevinji (changes requested):
Really minor stylistic differences.

kevinmasd:
Changes done.

kevinji (approved):

Issue #297, opened 2017-04-01
[DAWN] Runtime-Side State Information Display in Dawn


dHutchings:
Hi.

There's currently no visual feedback of any kind showing what the "Robot state" is.  Is the robot disable?  In teleop?  In autonomous?  Currently, there's no way to tell at a glance.

I understand that making the drop-down menu explicitly change is hard, so maybe that's not the best way forward.  But could we at least have an indicator, something like the battery voltage indicator?  It'd be better than exactly nothing.

Thank you!

kevinmasd:
Dawn displays what it sends according to #274 . As for what Runtime is sending back to Dawn, this isn't addressed but can be done since the state does hold this information.

PR #310, opened 2017-04-08
[Dawn] Display Runtime-Side Robot State

kevinmasd:
Addresses #297 

Modifies `DNav.js` and `utils.js`:
* Append Runtime State after Runtime Versioning
* Hides both if robot isn't running so students recognize that v0.0.0 shouldn't exist.


dawn/renderer/utils/utils.js
```diff
@@ -41,11 +41,17 @@ export const robotState = {
...
   STUDENT_RUNNING: 1,
+  1: 'Running',
   STUDENT_STOPPED: 2,
+  2: 'Stopped',
   TELEOP: 3,
+  3: 'Tele-Operated',
   AUTONOMOUS: 4,
+  4: 'Autonomous',
   ESTOP: 5,
+  5: 'E-Stop',
```
yizow, line 54: Just for the record, this will likely never get printed except momentairly. After receiving e-stop, runtime kills itself (this is deliberate)

yizow (approved):

dawn/renderer/utils/utils.js
kevinmasd: Noticed that. I'll just leave it here if we need it elsewhere.

Issue #300, opened 2017-04-02
[Dawn] Log File


kevinmasd:
Much information can be gathered through all the logging we do, but this goes directly to terminals and thus hard to debug externally. We need to redirect this information into a log file of some sort.

PR #311, opened 2017-04-09
[Dawn] Logging to File

kevinmasd:
Addresses #300 

Edits to `Ansible.js`:
* Initiates Ansible logging in `Ansible` object.

Edits to `utils.js`:
* Logger class created
  * Initiated in App.js, which is guaranteed to be the earliest file loaded.

Edits to `.eslintrc`:
* To prevent multiple calls to `Logger()`, we have to use `export let` rather than `export const`, since `const` does not allow for variable to be initialized other than the declaration line.


kevinji:
Does the last commit work? I feel like you'd need to call `logging().log(str)` with your current implementation.

dawn/main/ansible/Ansible.js
```diff
@@ -177,28 +180,29 @@ class SendSocket {
 }
 
 class TCPSocket {
-  constructor(socket) {
+  constructor(socket, logger) {
     this.waitRuntimeConfirm = this.waitRuntimeConfirm.bind(this);
     this.tryUpload = this.tryUpload.bind(this);
+    this.logger = logger;
```
yizow, line 185: not exactly sure what binding does; does it mean the bound function can start being scheduled? if so, we may want `this.logger` to be set before bidning.

kevinmasd: Logger is always set before this part. It is initialized at the very bottom, right before any of the other classes are running.

dawn/renderer/utils/utils.js
```diff
@@ -60,3 +62,31 @@ export const timings = {
...
+    if (output !== this.lastStr) {
+      this.log_file.write(`\n${output}`);
+      this.lastStr = output;
+    } else {
+      this.log_file.write('*');
+    }
+  }
+}
+
+export const logging = () => {
```
kevinji: Per my comment from earlier: This is a function, so if you wanted to access the logger you'd probably need `logging().log` with your current implementation; it'd probably make more sense to just set `logging = new Logger();` and remove `_log`.

Also, could you possibly remove the duplicate version of `Logger` from `Ansible`, export this version, and give this one a constructor argument for file name?

dawn/renderer/utils/utils.js
```diff
@@ -60,3 +62,31 @@ export const timings = {
   TELEOP: 120,
   SEC: 1000,
 };
+
+let _log = null;
+class Logger {
+  constructor() {
+    const d = new Date();
+    this.log_file = fs.createWriteStream(`./${d.valueOf()}-dawn.log`, { flags: 'w' });
+    this.log_file.write(d.toString());
```
kevinji: Is this really necessary, considering it's part of the file name already?

[EDIT] On second thought, maybe it'd be good to log the time of every message next to the message, like so:
`[2017-04-09T18:55:57.325Z] ERROR: Could not connect to TCP server.`

dawn/renderer/utils/utils.js
```diff
@@ -60,3 +62,31 @@ export const timings = {
   TELEOP: 120,
   SEC: 1000,
 };
+
+let _log = null;
+class Logger {
+  constructor() {
+    const d = new Date();
+    this.log_file = fs.createWriteStream(`./${d.valueOf()}-dawn.log`, { flags: 'w' });
```
kevinji: I think `d.getTime()` (or even `Date.now()` if the below line isn't needed) is more commonly used.

dawn/main/ansible/Ansible.js
```diff
@@ -212,15 +216,15 @@ class TCPSocket {
 
   waitRuntimeConfirm(message, count) {
     if (count > 3) {
-      console.error('Runtime failed to confirm');
+      this.logger.error('Runtime failed to confirm');
```
kevinji: This doesn't appear like it'll work, as the `Logger` class has no `error` method.

kevinji (changes requested):
Left a few comments. Also, this is probably more of a long-term thing, but it might make sense to better differentiate log messages between `info`, `warning`, and `error` to make it easier to distinguish what is and isn't important.

kevinmasd:
For some reason thought the logging function would replace itself with the interior. Creating the export must be separate from variable initialization (`export const logger = new Logger()` causes a new Logger to be created on every message).

kevinji:
Could you do this instead:
```
const logger = new Logger();
export logger;
```
I forget exactly how export rules work.

kevinmasd:
Throws parsing errors, so probably doesn't work.

kevinji (approved):

PR #314, opened 2017-04-11
[DAWN] Tweak logging functionality

kevinji:
It makes more sense for logging to _append_ to a file, not completely replace an existing one. Also, we shouldn't need to recreate a WriteStream every time the page is refreshed.


kevinmasd (changes requested):
Just tested it. Looks like while the Ansible debugger works, the main renderer-side doesn't due to not being able to access logger. Perhaps starting the logger from main-develop does not allow the renderer to see it?

kevinji:
@kevinmasd Backed out that change until I have time to debug it further.

kevinmasd (approved):

PR #318, opened 2017-04-13
[Dawn] 0.6.0 Log Directory Fix

kevinmasd:
Due to Node and how __dirname is handled, this right now is the only clear way to handle pathing given current scheduling.

### DIFF
--- a/dawn/renderer/utils/utils.js
+++ b/dawn/renderer/utils/utils.js
@@ -81,7 +81,17 @@ export const windowInfo = {
 
 export class Logger {
   constructor(processname, firstline) {
-    this.log_file = fs.createWriteStream(`${__dirname}/${Date.now()}-${processname}.log`, { flags: 'w' });
+    let path = '';
+    if (processname === 'ansible') {
+      path = __dirname;
+      while (path.includes('dawn.')) {
+        path = path.substring(0, path.lastIndexOf('/'));
+      }
+    } else {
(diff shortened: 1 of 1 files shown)


kevinmasd:
Looking into that.

kevinmasd:
Decided to shift to desktop location, one of the options definitely known to exist cross-platform. (The others are the temp folders or the download folder)

dawn/renderer/utils/utils.js
```diff
@@ -81,7 +81,14 @@ export const windowInfo = {
 
 export class Logger {
   constructor(processname, firstline) {
-    this.log_file = fs.createWriteStream(`./${Date.now()}-${processname}.log`, { flags: 'w' });
+    let path;
+    if (processname === 'dawn') {
+      path = require('electron').remote.app.getPath('desktop'); // eslint-disable-line global-require
```
kevinji, line 86: Question: Is there a reason this isn't imported at the top of the file?

kevinmasd: Main vs renderer difference. One throws an error if used in one process, the other throws an error if used in the other process.

kevinji (approved):
Looks okay to me. I think #314 should also be merged in after this to use append mode instead of write mode for log files.

kevinmasd:
It's already released with this feature. Any other changes will not be in 0.6.0

PR #320, opened 2017-04-14
[DevOps] Field control

jonathan-j-lee:
* Implemented prioritization of network selection with `wpa_supplicant`
    * Team router is connected to over field control router
* Implemented field control router configuration
    * Beaglebone connects to `192.168.128.2**`, where `**` is replaced by the team number
* Reenabled Ethernet (visible as `192.168.6.2`)
* Reimplemented `interfaces.py` as `interfaces.sh` (easier to `scp`)
* Changed targets for `runtime.service` and `update.service`
    * With the implementation of the USB fix, `multi-user.target` fails

This branch is intended to replace [devops/whyfi](https://github.com/pioneers/PieCentral/pull/246) with a much cleaner solution for router selection.


AravindK95:
SGTM as long as we do a deployment test

akumaraguru (dismissed):

ArjunSridhar1 (dismissed):

ArjunSridhar1 (approved):

PR #327, opened 2017-04-23
[Dawn] Cleaner Logging

kevinmasd:
`utils.js`:
* Move away from two log files per run to two log files in total
  * This means replacing older information, assuming that bugs can be found in the most recent run
* Move log files into ~/Desktop/Dawn (or equivalent locations)


kev-chien (approved):
Looks good. Maybe tell students about this feature so they know to save bug reports elsewhere when bugs arise.

dawn/renderer/utils/utils.js
```diff
@@ -88,7 +88,12 @@ export class Logger {
       const { app } = require('electron'); // eslint-disable-line global-require
       path = app.getPath('desktop');
     }
-    this.log_file = fs.createWriteStream(`${path}/${Date.now()}-${processname}.log`, { flags: 'a' });
+    try {
+      fs.statSync(`${path}/Dawn`);
+    } catch (err) {
+      fs.mkdirSync(`${path}/Dawn`);
```
kevinji: It might be easier just to `try` the `mkdirSync` call and ignore the error if `err.code === 'EEXIST'`, removing the `statSync` call entirely, and `console.log`-ing the error otherwise.

PR #328, opened 2017-04-23
[Dawn] Download Robot's studentCode

kevinmasd:
`EditorContainer.js`, `Editor.js`, `EditorActions.js`:
* Added `DOWNLOAD_CODE` action, initiated as another button.

`sagas.js`:
* Added `downloadStudentCode` to handle file download through promise, which uses a switch over resolve values for error output as redux-sagas does not work in callbacks. Actual placement of code into editor is done in the success case, basing off `openFile`


dawn/renderer/components/Editor.js
```diff
@@ -439,6 +439,12 @@ class Editor extends React.Component {
               onClick={this.upload}
               glyph="upload"
             />
+            <TooltipButton
```
kev-chien: I think maybe add button disabled case, because the generator that handles `onDownloadCode` doesn't "run" unless
`stateSlice.runtimeStatus && stateSlice.ipAddress !== defaults.IPADDRESS`.
This might be somewhat helpful for the student UI-wise.

kev-chien, line 442: I think maybe add button disabled case, because the generator that handles `onDownloadCode` doesn't "run" unless
`stateSlice.runtimeStatus && stateSlice.ipAddress !== defaults.IPADDRESS`.
This might be somewhat helpful for the student UI-wise.

kev-chien (changes requested):
I think if tested with robot, looks good.

dawn/renderer/components/Editor.js
kevinmasd: Makes sense. Originally I was going to make it try to get the file without any protocol (i.e. just assume the robot is there, since SFTP doesn't care if runtime's up), but that can be done later.

kev-chien (approved):

PR #329, opened 2017-04-23
[Dawn-FC] Field Control for 2017

kevinmasd:
This is the active branch for field control modifications to Dawn for the 2017 competition, in a bid to prevent constant rewrites every season.

Features:
* Ability to Configure station number and field control master address
* Ability to toggle between Field Control mode and normal mode


dawn/main/MenuTemplate/DawnMenu.js
```diff
@@ -14,6 +15,14 @@ const DawnMenu = {
         app.quit();
       },
     },
+    {
+      label: 'Field Control Mode',
+      click() {
+        RendererBridge.reduxDispatch({
+          type: 'TOGGLE_FIELD_CONTROL',
```
brandonxxlee, line 22: Is this toggle going to be a public facing toggle? What is the use case of students toggling between Field Control Mode and not. (Correct me if I'm completely misreading the code.

brandonxxlee (approved):
As a whole, really great job! I left one comment, and noticed that you had a lot of todos so please also make relevant github issues for the todos so that we can track them and not forget about them. 

I'm not extremely familiar with javascript or the Dawn codebase, so ideally someone else looks over it once too but it lgtm. (I trust you that it works if you've tested it, so I'll approve it anyways, up to you whether you merge it or not)

dawn/main/MenuTemplate/DawnMenu.js
kevinmasd: Copying from Slack for future reference. This toggle is public; it becomes very obvious that Dawn is set to Field Control mode when toggling. Otherwise, it acts much the same was as regular Dawn, only that it may be overridden if it is connected to an active FC master station. Perhaps later it could be made into a "hidden" feature, but I don't really think that to be necessary.

kevinmasd:
Lots of those TODOs were from porting code over and seeing what needs to be kept. For the most part, they will be removed in a later PR, mostly to keep changes small.

kevinmasd:
Will pull in by Wednesday if no one else approves or makes comments

dawn/renderer/utils/lcm_ws_bridge.js
```diff
@@ -0,0 +1,148 @@
...
+                 [true, true, false],
+                 [true, true, false]],
+        servos: [0, 0, 0, 0, 0, 0, 0, 0]
+      });
+    });
+*/
+import WebSocket from 'ws';
+
+
+const CircularJSON = require('circular-json');
```
nikitakit, line 50: I've been curious for a while now: why are to using circular json? The python bridge server certainly doesn't understand any JSON extensions, so sending it circular-json json annotations seems unnecessary.

kevinmasd: I don't quite remember the details, but there was some issue serializing data to pass through Websockets due to some quirk in the system causing the object to reference itself. I actually have no idea how the main station parses the data; if anything, the circular part of the JSON would probably be truncated off somewhere along the line.

nikitakit: Would you perhaps be willing to spend some time to get to the bottom of this issue? Maybe it would be possible to leave a comment explaining why this is required, or even to get rid of whatever is creating the circular objects in the first place.

The serialization issue you're describing sounds very surprising to me, and if you no longer remember the details then I'm not sure anyone in PiE understands why this code works. I'm of the opinion that it's better to spend time once to figure out what's going on, rather than punting the rediscovery of this issue down to new staff at some point in the future.

Edit: to be clear, I'm commenting about looking into this in general (not necessarily as part of this PR)

kev-chien (approved):
Looks good to me too.

kev-chien:
One concern I do have is readability and maintanability. First of all, understanding how field control is toggled requires understanding Redux and how actions and reducers work together, and also how (and why) sagas are used. This means for new staff a little bit of a learning curve to get started, and possible confusion as it involves jumping around several files. Is there a workaround to make understanding this easier, i.e. an IDE or workflow that makes this easy to understand or debug? (I assume Webstorm makes this easier but haven't really used it that much)

I think in overall I'd like to suggest more documentation, either in code comments or in the wiki. For simpler things I think it'd be fine/ maintainable to leave comments in the code, but there is a lot of code that can't be easily understood unless explained in a wiki. For example,
- it isn't clear why ipcRenderer is used, and why the function calls have to switch back and forth to `maindevelop,js`. A short Google search shows that it has something to do with Electron's async calls, but this might not be immediately clear to new staff. In general it might be helpful to have more documentation to explain why ipcRenderer or ipcMain is used.
- There is a lot of switching between files to try and figure out Field Control. Perhaps using the wiki might be useful to "tag" which files are Field Control related and what they do, etc. (and for other parts of Dawn as well)
UPDATE: I see now that ipcRenderer and Field Control have been updated in the wiki. There is still more that could be explained; I could try doing so myself.

Also, `LCM.js` could be easily confused with `lcm_ws_bridge.js` (the library). It seems like `LCM.js` is meant strictly for field control related aspects, and uses lcm_ws_bridge. So why is LCM.js located in the ansible folder? It doesn't use Ansible.js as well. Perhaps it might be less confusing if the ansible folder were renamed "networking", and if LCM.js were renamed to `LCM_FC.js` (suggesting Field Control) or even `FieldControlLCM.js`

Similarly I think the name `main-develop.js` is not memorable/ is vague. In a comment it says "Entrypoint for dawn's main process" which doesn't say too much either about the entire file, but at least "main process" gives the notion that it initializes the browser window and FC components. So I think it should be renamed `main-process.js`.

kevinji:
Gonna review this when I get home today.

dawn/main/ansible/Ansible.js
```diff
@@ -57,7 +61,7 @@ function buildProto(data) {
   return new DawnData({
     student_code_status: status,
     gamepads,
-    team_color: 0,
+    team_color: (stationNumber < 2) ? DawnData.TeamColor.BLUE : DawnData.TeamColor.GOLD,
```
kevinji: This `stationNumber` (as well as the others) should be given constants if possible.

dawn/main/ansible/LCM.js
```diff
@@ -0,0 +1,119 @@
...
+    const connected = data.connectionStatus;
+    const ok = data.runtimeStatus;
+    const msg = {
+      __type__: 'StatusLight',
+      red: false,
+      yellow: false,
+      green: false,
+      buzzer: false,
+    };
+    if (connected) {
```
kevinji: Can condense this loop into
```
if (connected && ok) {
  msg.green = true;
} else {
  msg.red = true;
}
```

dawn/package.json
```diff
@@ -73,6 +76,7 @@
     "seedrandom": "2.4.2",
     "smalltalk": "2.2.0",
     "ssh2": "0.5.4",
-    "superagent": "3.4.1"
+    "superagent": "3.4.1",
+    "ws": "^2.2.3"
```
kevinji: These package installs should ideally be exact versions (without any `^` or `=` prefix). I believe `yarn add [pkg] --exact` (or `-E`) will do what you want.

dawn/renderer/components/App.js
```diff
@@ -173,6 +184,10 @@ const mapDispatchToProps = dispatch => ({
   onIPChange: (ipAddress) => {
     dispatch(ipChange(ipAddress));
   },
+  onFCUpdate: (ipAddress) => {
+    dispatch(updateFieldControl(ipAddress));
+  },
+
```
kevinji: Extra newline here.

dawn/renderer/utils/sagas.js
```diff
@@ -281,12 +283,20 @@ const gamepadsState = state => ({
   gamepads: state.gamepads.gamepads,
 });
 
+const lcmState = state => ({
+  connectionStatus: state.info.connectionStatus,
+  runtimeStatus: state.info.runtimeStatus,
+});
+
```
kevinji, line 290: Extra newline.

dawn/main/ansible/LCM.js
```diff
@@ -0,0 +1,119 @@
...
+import RendererBridge from '../RendererBridge';
+
+class LCMInternals {
+  constructor(stationNumber, bridgeAddress) {
+    this.lcm = null;
+    this.lcmReady = false;
+    this.queuedPublish = null;
+    this.stationNumber = stationNumber;
+    this.bridgeAddress = bridgeAddress;
+    console.log(`LCM SN:${this.stationNumber} BA:${this.bridgeAddress}`);
```
kevinji: This `console.log` (as well as other calls elsewhere) should probably be `logging.log` instead.

dawn/renderer/utils/lcm_ws_bridge.js
```diff
@@ -0,0 +1,148 @@
...
+        if (callback !== undefined) {
+          callback(request.data.msg);
+        }
+        break;
+      default:
+        throw 'Invalid request!'; // eslint-disable-line no-throw-literal
+    }
+  }
+
+  generate_uuid() { // eslint-disable-line
```
kevinji, line 103: Is there a reason why we don't just use [node-uuid](https://github.com/kelektiv/node-uuid) here?

kevinji (changes requested):
Everything is mostly reasonable, just some minor fixes.

dawn/renderer/utils/lcm_ws_bridge.js
kevinmasd: Old code; might revisit later if needed.

dawn/main/ansible/LCM.js
kevinmasd: Typo. Should be yellow in the middle case

dawn/package.json
kevinmasd: Bumping to v3. No interface changes, so should be fine

dawn/main/ansible/Ansible.js
kevinmasd: Forgot to update to use changeable value

kevinmasd:
@kev-chien Documentation is slowly coming (due to busy schedule currently). Learning curve could be made better with more uniform code practices (see future PR) that the linter can't really enforce. Your file-renaming arguments are valid and have been addressed in this PR

dawn/renderer/utils/lcm_ws_bridge.js
kevinmasd: Quirk might have been when communicating with station, since I cannot reproduce it right now (i.e. never tested FC setup on Mac). On quick check now, it looks slight different and without circularity, but not sure.

dawn/renderer/utils/sagas.js
kevinmasd: This has been resolved.

dawn/renderer/utils/lcm_ws_bridge.js
nikitakit: The original version of this file, [still found in the forseti2 repo](https://github.com/pioneers/forseti2/blob/gh-pages/lcm_ws_bridge.js), was written to be used inside the browser. The web UI for managing the field was written as a static HTML page for easy editing and non-reliance on web hosting.

As a result it had to be dependency-free, which is why there is handcoded uuid generation here. So between needing to maintain browser compatibility ~3 years ago when the file was written, as well as my lack of familiarity with web programming at the time, the code certainly seems odd by modern standards. Though it has been modernized quite a bit after getting merged into Dawn.

I'm also curious what will end up happening to the field UI, since the current version of this file has diverged from something that can be used compiler-free in the browser.

kevinji (approved):

PR #330, opened 2017-04-27
[Dawn] Better Debug Output

kevinmasd:
`utils.js`:
* Added way for things to by pass console and go straight to Debug log

`Ansible.js`:
* Log things sent and received from Runtime using new Logger capability


kev-chien (approved):
Looks good.

PR #333, opened 2017-05-11
Make staff-bots directory and add staff infection's code.

vdonato:


brandonxxlee (approved):
lgtm, probably would be useful to use this eventually as a place to just store sample code/staff code

PR #335, opened 2017-05-27
[HIBIKE] Delete unused files

brandonxxlee:
Deleted unused files, Karthik approved files that can be deleted but someone should double check.


akumaraguru (changes requested):
Looks good so far. There are some other files I think can go away as well.

reset.py - legacy code for a quirky bootloader when we made our own smartsensors. Currently this is redundantly called 3 times every time we run flash.sh. Make sure to remove the relevant lines from flash.sh as well.

hibike_architecture.pdf - how badly do we need this diagram? At the very least we should make something better designed, ideally as an svg

\_\_init\_\_.py - why does this import hibike_simulator.py? I don't see this anywhere in piecentral

brandonxxlee:
I removed the hibike_architecture.pdf because its kinda jank, and its not even fully correct anymore. A new one should be made somehow at some point anyways that's up to date.

akumaraguru (approved):

akumaraguru:
Could clean out reset.py as well, but I won't nitpick

brandonxxlee:
I would, but I don't really want to touch something that I don't know at all, just in case. I'll merge this for now, and make an issue for that.

Issue #337, opened 2017-05-31
[HIBIKE] Clean up reset.py and __init__.py


brandonxxlee:
reset.py - legacy code for a quirky bootloader when we made our own smartsensors. Currently this is redundantly called 3 times every time we run flash.sh. Make sure to remove the relevant lines from flash.sh as well.

__init__.py - why does this import hibike_simulator.py? I don't see this anywhere in piecentral

tingtingchiang:
Fixed

PR #342, opened 2017-06-07
Remove old reset code from Hibike

baby-bell:
As per #337, this code is old and can be removed. I am not knowledgeable enough to touch `__init__`.


baby-bell:
Doug has told me that this code is necessary, so I will close this PR.

Issue #340, opened 2017-06-06
Move files out of testy to runtime folder


brandonxxlee:

jonathan-j-lee:
Addressed in #347.

PR #347, opened 2017-06-24
Removed testy

jonathan-j-lee:
* Moved all files in `runtime/testy` to `runtime` (addresses #340)
* Updated Dawn, runtime, and DevOps sources and documentation to use `runtime` rather than `runtime/testy`


brandonxxlee (approved):
lgtm

PR #344, opened 2017-06-20
[DAWN] Update deps, make lint fixes

kevinji:
Most exciting is Webpack 3.0.0!

PR #349, opened 2017-06-26
upgraded to webpack 3

kev-chien:
webpack 3 is out. I suppose upgrading is better to keep up with new (possibly useful) features?


kevinji:
I have this included in #344, which is waiting for one package to get updated.

kev-chien:
Oh okay, in that case I'll close my PR and wait on yours; it has more features.

PR #365, opened 2017-07-15
[HIBIKE] Fix hang in device detection

### LINKED ISSUE #363: [HIBIKE] Device detection hangs when some ports are not writeable
Hibike attempts to detect devices based on whether they return valid packets after a subscription request. Unfortunately, if the serial port does not accept writes, the process hangs forever. This may not be a big issue in practice.

baby-bell:
Previously, Hibike would hang if a serial port did not accept writes. This PR imposes a timeout on device responses. Fixes #363


hibike/hibike_process.py
```diff
@@ -75,8 +75,16 @@ def recv_subscription_requests(conn, uid_queue, stop_event):
     thread_list = []
     event_list = []
     read_queues = []
+
     for conn in serial_conns:
-        hm.send(conn, hm.make_ping())
+        old_timeout = conn.write_timeout
+        conn.write_timeout = IDENTIFY_TIMEOUT
```
yizow, line 81: The definition of `IDENTIFY_TIMEOUT` shouldnt be hiding halfway through this file.

brandonxxlee (commented):
lgtm just fix Yizhe's comment and I'll approve it

baby-bell:
I've moved the constant to the top of the file, below the batch time.

brandonxxlee (approved):
Lgtm

PR #367, opened 2017-07-20
[HIBIKE] Handle device disconnect

baby-bell:
Previously, a device disconnect would not send any indication to runtime and not clean up associated resources. This PR creates a new message type, `device_disconnect`, indicating that a device has disconnected, and cleans up resources when such a disconnect occurs. This is (almost) half of hotplug, the other being scanning for new devices.


runtime/stateManager.py
```diff
@@ -271,6 +272,14 @@ def hibike_response_device_values(self, data):
             for key, value in params:
                 self.set_value(value, ["hibike", "devices", uid, key], send=False)
 
+    # pylint: disable=invalid-name
+    def hibike_response_device_disconnect(self, uid):
+        """
+        Delete any history of the device at UID.
+        """
+        devs = self.state["hibike"][1]["devices"]
```
bodiddles: To actually access the device dictionary, you want to do self.state["hibike"][0]["devices"][0] to parse through the timestamps.

bodiddles (changes requested):
Other than the change in stateManager, this should work. You should fully test it with dawn and see if causing a device disconnect would actually make the device disappear in Dawn.

bodiddles (approved):
lgtm from runtime side.

PR #369, opened 2017-07-27
[HIBIKE] Hotplug

baby-bell:
This addresses #354. It is currently based on polling, with a period of 250ms. It attempts to make sure that a device is really disconnected by disconnecting it only after two periods. In the future, it might be easier to switch to a dmesg-based version.


akumaraguru (changes requested):
line 54: grammar error in comment
line 79: should be recv_subscription_response()
line 240: In hibike_process, accesses to devices dict should be wrapped in a try/catch. The current guarding method of `if uid in devices` is not atomic, and could (very rarely) crash the thread.

In general: I see multiple uses of `while True: <do i/o stuff>` in multiple threads. This is not ideal, and should be replaced with a mechanism that allows for more graceful shutdown of a thread. Consider using threading.Events. This isn't something that needs to be fixed in this PR, but definitely sometime soon.

akumaraguru (approved):

baby-bell:
I'd like to wait for one more review before I merge this, to be extra sure

hibike/hibike_process.py
```diff
@@ -85,25 +85,123 @@ def recv_subscription_requests(conn, uid_queue, stop_event):
...
+        time.sleep(HOTPLUG_POLL_INTERVAL)
+        scan_for_new_devices(devices, state_queue, batched_data, error_queue)
+        remove_disconnected_devices(error_queue, devices, clean_up_queue, state_queue)
+
+
+def scan_for_new_devices(existing_devices, state_queue, batched_data, error_queue):
+    """
+    Scan for devices not in EXISTING_DEVICES, and add any if found.
+    """
+    ports, names = get_working_serial_ports(map(lambda d: d.serial_port.name,
```
yizow: Will this error if called multiple times? 
`serial.Serial()` immediately attempts to open a port on object creation. 
What happens if the port is already opened?
[docs](http://pyserial.readthedocs.io/en/latest/pyserial_api.html)

hibike/hibike_process.py
```diff
@@ -20,22 +20,28 @@
 BATCH_SLEEP_TIME = .04
 # Time in seconds to wait until reading from a potential sensor
 IDENTIFY_TIMEOUT = 1
+# Time in seconds to wait between checking for new devices
+# and cleaning up old ones.
+HOTPLUG_POLL_INTERVAL = 0.25
```
yizow: I recommend lowering this to ~1 second. 4 Hz seems a bit oversensitive.

This is something that should be looked at later in combination with user testing

hibike/hibike_process.py
```diff
@@ -85,25 +85,123 @@ def recv_subscription_requests(conn, uid_queue, stop_event):
...
+        try:
+            error = error_queue.get(block=False)
+            pack = devices[error.uid]
+            if not error.accessed:
+                # Wait until the next cycle to make sure it's disconnected
+                error.accessed = True
+                next_time_errors.append(error)
+                continue
+            elif error.serial_port.name != pack.serial_port.name:
+                # We found it again
```
yizow: Maybe expand on this comment with a few more lines.
Had to think about this for a while before finally getting it.

(ex. "scan_for_new_devices will add a serial port to `devices` if reconnected) etc.

yizow: What happens if device gets readded on the same port?

hibike/hibike_process.py
```diff
@@ -85,25 +85,123 @@ def recv_subscription_requests(conn, uid_queue, stop_event):
...
+                continue
+            uid = error.uid
+            pack = devices[uid]
+            del devices[uid]
+            clean_up_queue.put(pack)
+            state_queue.put(("device_disconnected", [uid]))
```
yizow, line 208: I'd move this logging line into clean_up_queues, and then either remove this altogether, or change it to something like "Attempting to disconnect uid"

yizow (commented):
I don't want to be the one approving.

A change of this magnitude should be reviewed by at least one other active, non-alumni staff member. 

So Not me, nikita, karthik (he'll be at fb next semeseter), etc

hibike/hibike_process.py
baby-bell: Logging line? This tells state manager that the device was disconnected.

hibike/hibike_process.py
baby-bell: Opening a serial port multiple times is not an issue. The issue is two instances issuing instructions at the same time.

hibike/hibike_process.py
baby-bell: I have tested this, and devices don't get added on the same port.

yizow (changes requested):
Blocking while this is in testing

tingtingchiang (approved):
+1 previous comments, LGTM if the tests with the robot in a box are fine.

hibike/hibike_process.py
```diff
@@ -85,25 +90,124 @@ def recv_subscription_requests(conn, uid_queue, stop_event):
...
+        except queue.Empty:
```
tingtingchiang, line 209: What's the difference between this logic and looping while not queue.Empty + moving this exception part to outside of the while loop?

baby-bell: [Queue docs](https://docs.python.org/3/library/queue.html?highlight=queue#module-queue) state that `.empty()` could return true even if there are still items in the queue, so easier just to keep trying until there's nothing.

hibike/hibike_process.py
yizow: Oops, my bad. 
My comment still stands though. A device has not actually been disconnected here; the cleaner thread that does the disconnecting has merely been notified.

hibike/hibike_process.py
yizow: That is a quirk/ coincidence/ implementation detail of the current kernel/ OS/ etc.
Do not rely on that functionality.

yizow: Unless there is an explicit doc on Linux man page or something stating that sequential device numbers are the expected, *explicitly specified* behavior.

hibike/hibike_process.py
yizow: Is there a chance that the queue will never be empty, because something else keeps adding to it?
if so, nothing will ever get put on the error_queue, correct?

baby-bell: At most, there can be as many messages in the error queue as there are devices. A device has to be active to send a message to the queue, and adding devices doesn't happen concurrently, so the queue cannot always have items in it (at this point).

hibike/hibike_process.py
baby-bell: I am unclear as to what you mean by "disconnected". Do you mean:

1. the serial port no longer exists OR
2. the resources associated with the device do not exist.

The cleanup thread just closes the serial port and shuts down the read/write threads. StateManager is responsible for removing device data, and we tell it to do so via `state_queue`.

hibike/hibike_process.py
baby-bell: I have changed this to check an instance id instead.

hibike/hibike_process.py
yizow: I mean whatever `state_queue.put(("device_disconnected", [uid]))` is trying to do.

currently, it is possible for StateManager to get the message to remove data *before* the read/write threads stop sending. I dont think that's the behaviour we want.

My proposal is to move that state_queue.put line to the clean up queue, where we  can guarantee that cleanup is finished.

baby-bell: The error condition is triggered inside `read_thread`; I don't think it's possible for read/write threads to be active by the time we reach this point, since they have already crashed. It is possible to move this into the clean-up thread, but that might mean the message doesn't send for 30 or more seconds.

yizow (approved):

Issue #345, opened 2017-06-20
[DAWN] Upgrade protobuf.js to v6


kevinji:
This is non-trivial to do, as the API was basically rewritten for v6. See the [wiki docs](https://github.com/dcodeIO/protobuf.js/wiki) for more info.

PR #371, opened 2017-07-30
[Dawn] Bump protobufjs to v6

kevinmasd:
Addresses #345.

* While protobufjs's documentation indicates that it should be used asynchronously, the synchronous way looks visually simpler and makes code conversion less difficult
  * Note: This does not mean that we can't change to more async-code later
* The `(new protobuf.Root())` syntax is due to a minor inconsistency between default `protobuf.loadSync` and `Root().loadSync`. The latter allows exact variable naming (i.e. `int_param` instead of being converted to `intParam`
* Due to the way the parser [works](https://github.com/dcodeIO/protobuf.js/issues/631), `oneof` only works if all of the message containing it explicitly calls `{Message}.create`. Since Runtime's side has not seen this change/I haven't tested with actual Runtime, I instead made attributes explicit in display
  * Note: We may keep this functionality later (i.e. Acts as a type checking)


dawn/package.json
```diff
@@ -61,7 +61,7 @@
     "mousetrap": "1.6.1",
     "numeral": "2.0.6",
     "object-assign": "4.1.1",
-    "protobufjs": "5.0.1",
+    "protobufjs": "^6.8.0",
```
kevinji: Can you make this an exact dep (like `yarn add protobufjs -E`)?

kevinji (approved):
Looks reasonable. I don't know `protobufjs` too well, but it kinda sucks that you need to keep specifying `keepCase: true` everywhere. If there's a place to globally set this, that'd be great, but if not, it's fine.

kevinmasd:
There's no clean way to set it globally, so might as well be verbose about it.

PR #372, opened 2017-07-30
[DAWN] yarn: Upgrade deps, use frozen-lockfile

kevinji:
Always prefer exact versions if possible.


.travis.yml
```diff
@@ -15,6 +15,8 @@ cache:
     - $HOME/.cache/yarn
 before_install:
   - nvm install --lts
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.27.5
```
kevinmasd, line 18: Do we care about the specific version of the shell?

kevinji: Not *that* much, but the [Yarn docs](https://yarnpkg.com/lang/en/docs/install-ci/#travis-tab) recommend leaving it in to avoid potential breakage. I mainly needed this line to upgrade Yarn to a version with `--frozen-lockfile`.

kevinmasd: Sounds fine

kevinmasd (approved):

PR #376, opened 2017-08-01
editorconfig: Merge individual files

kevinji:
Now that Hibike and Runtime use the same indentation style, there's no need for individual .editorconfig files.


brandonxxlee (approved):
lgtm

PR #378, opened 2017-08-01
[Dawn] Component Code Cleanup

kevinmasd:
Going to be one in a series to cleanup code before the semester starts

* The old Container Model is switched into the Component model for naming, since its the wrapper we are specifically referencing
  * Files that separate the two are merged together, with exception of Editor due to size and NameEdit due to questionable future
* Initial Changes to add [PropTypes](https://github.com/facebook/prop-types)
* Functions made simpler
  * i.e. No reason to make function if functionality fits in one line, not used elsewhere
* Initial Props cleanup (Haven't removed branches due to unknown efficiency changes in tree reconciliation)


dawn/package.json
```diff
@@ -50,8 +50,9 @@
   },
   "dependencies": {
     "brace": "0.10.0",
-    "circular-json": "0.3.1",
-    "create-react-class": "15.6.0",
+    "bufferutil": "3.0.2",
+    "circular-json": "^0.3.1",
+    "create-react-class": "^15.6.0",
```
kevinji: These two should be exactly-versioned.

dawn/webpack.config.babel.js
```diff
@@ -24,6 +24,7 @@ const plugins = [
   new webpack.DefinePlugin({
     VERSION: JSON.stringify(version),
   }),
+  new webpack.IgnorePlugin(/vertx/),
```
kevinji: Is there a reason this is needed?

dawn/renderer/components/peripherals/Motor.js
```diff
@@ -1,37 +1,36 @@
...
-import { ProgressBar } from 'react-bootstrap';
 import numeral from 'numeral';
 
+import { ProgressBar } from 'react-bootstrap';
+
 /**
- * A component representing a motor.
+ *  Motor Component
  */
 const Motor = props => (
```
kevinji: Instead of `props`, you can probably just spread out the attributes: `{ id, param, device_name }`.

dawn/renderer/components/peripherals/GenericPeripheral.js
```diff
@@ -1,36 +1,34 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import _ from 'lodash';
 import numeral from 'numeral';
 
 /**
- * A generic peripheral, used when the peripheralType is unknown.
+ * Generic Peripheral for General Case
  */
 const GenericPeripheral = props => (
```
kevinji: Instead of `props`, you can probably just spread out the attributes: `{ id, param, device_name }`.

dawn/renderer/components/peripherals/BooleanSensor.js
```diff
@@ -1,51 +1,45 @@
...
+import PropTypes from 'prop-types';
 import _ from 'lodash';
 import { PeripheralTypes } from '../../constants/Constants';
 
 /**
- * A component representing a boolean sensor,
- * for example a LimitSwitch.
+ * Boolean Sensor Component
  */
 class BooleanSensor extends React.Component {
```
kevinji: Since this has no state, can you convert this to a functional-style component, and convert `formatBoolean` to a helper function?

dawn/renderer/components/StatusLabel.js
```diff
@@ -14,19 +16,29 @@ const StatusLabel = (props) => {
...
+const mapStateToProps = state => ({
+  batteryLevel: state.peripherals.batteryLevel,
+  batterySafety: state.peripherals.batterySafety,
+});
+
+
+const StatusLabel = connect(mapStateToProps)(StatusLabelComponent);
+
 export default StatusLabel;
+
```
kevinji: Stray newline.

dawn/renderer/components/StatusLabel.js
```diff
@@ -14,19 +16,29 @@ const StatusLabel = (props) => {
...
+  batterySafety: PropTypes.bool.isRequired,
 };
 
+
+const mapStateToProps = state => ({
+  batteryLevel: state.peripherals.batteryLevel,
+  batterySafety: state.peripherals.batterySafety,
+});
+
+
```
kevinji: Stray newline.

dawn/package.json
```diff
@@ -76,6 +78,7 @@
     "smalltalk": "2.4.1",
     "ssh2": "0.5.5",
     "superagent": "3.5.2",
+    "utf-8-validate": "3.0.3",
```
kevinji, line 81: Is this used anywhere?

dawn/package.json
```diff
@@ -50,8 +50,9 @@
   },
   "dependencies": {
     "brace": "0.10.0",
-    "circular-json": "0.3.1",
-    "create-react-class": "15.6.0",
+    "bufferutil": "3.0.2",
```
kevinji: Is this used anywhere?

kevinji (changes requested):
Seems mostly reasonable. There are a bunch of places that use `PropTypes` without making them required; it'd be nice if you did a full pass and made the actually required props `isRequired`.

dawn/package.json
kevinmasd: Recommended dependency for ws.

dawn/package.json
kevinmasd: Recommended dependency for ws.

dawn/package.json
kevinmasd: Will change

dawn/webpack.config.babel.js
kevinmasd: Sometimes there's a weird bug when `make watch` stops due to finding a missing item. This was an attempt to prevent that, but it seems there are other bugs.

kevinmasd: Specifically I get an issue similar to https://github.com/redux-saga/redux-saga/issues/197. I'll debug on a separate branch.

dawn/renderer/components/peripherals/BooleanSensor.js
```diff
@@ -1,51 +1,42 @@
+/* eslint-disable camelcase */
```
kevinji, line 1: Why is this necessary?

dawn/renderer/components/peripherals/BooleanSensor.js
```diff
@@ -1,51 +1,42 @@
...
-    if (peripheralType === PeripheralTypes.LimitSwitch) {
-      return (sensorValue) ? 'Closed' : 'Open';
-    }
-    return sensorValue;
+const formatBoolean = (peripheralType, sensorValue) => {
+  if (peripheralType === PeripheralTypes.LimitSwitch) {
+    return (sensorValue) ? 'Closed' : 'Open';
   }
+  return sensorValue; // TODO: Verify if int_value or bool_value
+};
```
kevinji, line 15: This function seems a bit strange; the `if` condition expects a boolean (or truthy `sensorValue`) and returns a string, while the `else` branch expects an int (I think)?

dawn/renderer/components/peripherals/GenericPeripheral.js
```diff
@@ -1,36 +1,35 @@
+/* eslint-disable camelcase */
```
kevinji, line 1: Same comment as other file: why is this necessary?

dawn/renderer/components/peripherals/Motor.js
```diff
@@ -1,37 +1,37 @@
+/* eslint-disable camelcase */
```
kevinji, line 1: Ditto.

dawn/webpack.config.babel.js
kevinji: Can you leave a comment here mentioning that this is an issue?

kevinji (changes requested):
Just a few more minor changes.

dawn/renderer/components/peripherals/BooleanSensor.js
kevinmasd: It flags on the `device_name`, `device_type`, etc variables when unpacking from the object.

dawn/renderer/components/peripherals/BooleanSensor.js
kevinmasd: `sensorValue` as far as I can tell is always numerical. It just happens that for one sensor, we have semantics that allow us to translate value to word meaning.

kevinmasd:
The vertx thing is going to be removed, since it seems to hit a different bug preventing recompilation.

kev-chien (approved):
Looks good to me.

kevinji (approved):

PR #384, opened 2017-08-07
[HIBIKE] Recursively lint all Python files

baby-bell:
Just what the title says. This lints files in `hibike/` and all its subdirectories, as opposed to solely `hibike/`.


hibike/travis/Makefile
```diff
@@ -12,7 +12,7 @@ artifacts-install:
 	$(nop)
 
 lint:
-	cd .. && pylint *.py
+	cd .. && find | grep "\\.py$$" | xargs pylint
```
kevinji: Does something like `pylint **/*.py` not work?

baby-bell: No, that only works for files one directory down. I would like to lint all Python files in the directory tree. There might be a way to remove the `grep`.

kevinji: Ah, I didn't realize that `**` only globs recursively by default on zsh; on bash, you'd need to set `globstar`. Does `find . -name "*.py" | xargs pylint` work?

baby-bell: Yes, that works. Thank you.

kevinji (approved):
lgtm

PR #385, opened 2017-08-07
[HIBIKE] Some unit tests

baby-bell:
Previously, Hibike had no unit tests, although some testing infrastructure was in place (`hibike_tester` and `virtual devices`). This commit adds a few, nonexhaustive unit tests for functions in `hibike_message` and `hibike_process`. As of now, most tests are based on randomized input data, rather than specific inputs.


baby-bell:
These tests can be added to Travis, if preferred.

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,385 @@
...
+import random
+import queue
+import time
+import threading
+
+import serial
+
+import hibike_message
+import spawn_virtual_devices
+DEVICE_TYPES = [dev_id for dev_id in hibike_message.DEVICES]
```
tingtingchiang: Maybe just a stylistic choice, but list(hibike_message.DEVICES) reads better for me, or hibike_message.DEVICES.keys(). Both create a new list of keys.

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,385 @@
...
+
+def test_with_random_data(func, arg_func, kwarg_func=lambda: {}, times=5, *args, **kwargs):
+    """
+    Test FUNC with random arguments generated by ARG_FUNC and KWARG
+    func, repeating TIMES.
+    """
+    for _ in range(times):
+        # Do not expand list arguments out
+        generated_args = arg_func()
+        if isinstance(generated_args, tuple):
```
tingtingchiang: Does generated_args change from being a tuple to another type of instance between loops? If not, we should just generated the arguments and check once before deciding how to call the func. Otherwise, this for loop may become very expensive if arg_func is expensive.

tingtingchiang (changes requested):
Is it possible to also add a README file with descriptions on how to run tests? Or is there an example of how to run them somewhere?

hibike/unit_tests/hibike_message_tests.py
baby-bell: Can do!

hibike/unit_tests/hibike_message_tests.py
baby-bell: Yeah, I think you're right about this, especially since the data we are generating should always be of the same type.

baby-bell:
Just a warning: the waiting period for virtual devices is not enough for running tests on the Beaglebone, so it will usually fail `hibike_process_tests`. I can modify this if you like.

hibike/unit_tests/hibike_process_tests.py
```diff
@@ -0,0 +1,65 @@
...
+import serial
+from spawn_virtual_devices import spawn_device, get_virtual_ports
+from hibike_process import identify_smart_sensors
+import hibike_message as hm
+
+class IdentifySmartSensorsTests(unittest.TestCase):
+    """
+    Tests for `identify_smart_sensors`.
+    """
+    VIRTUAL_DEVICE_TYPES = ["LimitSwitch", "YogiBear", "YogiBear",
```
yizow: Will this ever be used / synced with something else? If so, it should be in a central file that all things are importing.

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,386 @@
...
+import random
+import queue
+import time
+import threading
+
+import serial
+
+import hibike_message
+import spawn_virtual_devices
+DEVICE_TYPES = list(hibike_message.DEVICES)
```
yizow: Why convert this to a list? `hibike_message.devices` is originally a dictionary that should have all the info you need. The way youre using `DEVICE_TYPES` later also suggests you want access to that data

hibike/hibike_message.py
```diff
@@ -405,17 +415,18 @@ def parse_bytes(msg_bytes):
...
     packets_buffer = bytearray()
-    while stop_event is None or not stop_event.is_set():
-
+    # Switch to nonblocking mode so that we don't get stuck reading
+    old_timeout = serial_conn.timeout
+    serial_conn.timeout = 0
+    while not stop_event.is_set():
         # Wait for a 0 byte to appear
-        while packets_buffer.find(zero_byte) == -1:
+        while packets_buffer.find(zero_byte) == -1 and not stop_event.is_set():
```
yizow: why are we duplicating this is_set here?
If this is an edge case that causes problems, then you should be able to restructure the code to address that logic.

hibike/hibike_message.py
```diff
@@ -104,6 +106,14 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    # pylint: disable=protected-access
+    def __eq__(self, other):
+        if not isinstance(other, HibikeMessage):
+            return False
+        return self._message_id == other._message_id and\
```
yizow: For multi line, preferred style is implicit continuation via wrapping in parentheses, and putting the boolean operator at the start of the line. 

see [here](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

hibike/hibike_message.py
```diff
@@ -104,6 +106,14 @@ def __str__(self):
...
+               self._length == other._length and\
```
yizow: does a `HibikeMessage` have a is_valid check or similar? If so, no need to compare the length with the payload

yizow: And if not, maybe its time to write one : P

hibike/hibike_message.py
```diff
@@ -437,10 +448,13 @@ def blocking_read_generator(serial_conn, stop_event=None):
                 packets_buffer = packets_buffer[new_packet:]
             # Otherwise, there might be more incoming bytes for the current packet,
             # so we do a blocking read and try again
-            else:
+            elif not stop_event.is_set():
```
yizow: i feel like putting this check *everywhere* is masking some underlying code structure issue. if you're worried about synchronization, then its a great time to
1) think hard about exactly how the threads/ processes can interleave and interfere
2) design the best, easiest, cleanest way to solve it
3) learn about python multi-threading/processing synchronization primitives

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def test_with_random_data(func, arg_func, kwarg_func=lambda: {}, times=5, *args, **kwargs):
+    """
+    Test FUNC with random arguments generated by ARG_FUNC and KWARG
```
yizow: i dont like changing the case of variable names, especially in a case-sensitve language. 

also `/s/KWARG\nfunc/KWARG_FUC`

yizow: I'd also replace "Test" with "Execute" or "Run"

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,386 @@
...
+    func, repeating TIMES.
+    """
+    # Do not expand returned lists out
+    multiple_args = isinstance(arg_func(), tuple)
```
yizow: Do you require arg_func() to be idempotent?

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,386 @@
...
+    if multiple_args:
```
yizow: get rid of the if-branch:
`multiple_args = tuple(arg_func())`

Or, just specify that arg_func must return an iterable in your docstring

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,386 @@
...
+        for _ in range(times):
+            func(*args, arg_func(), **kwargs, **kwarg_func())
+
+
+def random_uid():
+    """
+    Generate a random UID.
+    """
+    uid_type = random.choice(DEVICE_TYPES)
+    return uid_type << 72 | random.getrandbits(8) << 64 | random.getrandbits(64)
```
yizow: why not:
`uid_type << 72 |  random.getrandbits(72)`

hibike/virtual_device.py
```diff
@@ -89,19 +105,22 @@ def main():
         if msg.get_message_id() in [hm.MESSAGE_TYPES["SubscriptionRequest"]]:
             # Update the delay, subscription time,
             # and params, then send a subscription response
-            print("Subscription request recieved")
+            if args.verbose:
```
yizow: Hmm i feel like we could restructure this to get rid of arg.verbose check multiple times.
Eh probably a future pull request

hibike/virtual_device.py
```diff
@@ -26,11 +26,15 @@ def main():
...
+    parser.add_argument('-v', '--verbose',
+                        help='print messages when sending and receiving packets',
+                        action="store_true")
     args = parser.parse_args()
 
     device = args.device
     port = args.port
-    print(device, port)
+    if args.verbose:
+        print(device, port)
```
yizow: Add a more descriptive string

yizow (changes requested):
Will finish review of more in depth additional classes, etc. later

Also I'm gonna block till i get a good commit message : P

hibike/hibike_message.py
baby-bell: Actually, this particular check is not necessary anyways, since reads are nonblocking. I will remove that.

hibike/unit_tests/hibike_message_tests.py
baby-bell: Good point.

hibike/hibike_message.py
baby-bell: The goal is to be able to stop reads when the event is set. It's possible to get stuck in this loop if we don't check the event.

hibike/unit_tests/hibike_message_tests.py
baby-bell: No, it is supposed to generate random data.

hibike/unit_tests/hibike_message_tests.py
baby-bell: `random.choice`, which I am using a few times, only works with sequences. This doesn't include dictionaries.

hibike/unit_tests/hibike_message_tests.py
baby-bell: Doesn't look like there's ambiguity wrt variable names here.

hibike/hibike_message.py
kevinji: You can change this into an `if` and stick a `continue` at the end of this `while` loop to avoid repeating the `not stop_event.is_set()` condition.

hibike/virtual_device.py
kevinji: We can just have a function called `log` that contains the `args.verbose` logic, and then use `log("MESSAGE")` instead.

hibike/virtual_device.py
```diff
@@ -89,19 +105,22 @@ def main():
         if msg.get_message_id() in [hm.MESSAGE_TYPES["SubscriptionRequest"]]:
             # Update the delay, subscription time,
             # and params, then send a subscription response
-            print("Subscription request recieved")
+            if args.verbose:
+                print("Subscription request recieved")
```
kevinji: s/recieved/received

hibike/virtual_device.py
```diff
@@ -89,19 +105,22 @@ def main():
...
             params, delay = struct.unpack("<HH", msg.get_payload())
 
             subscribed_params = hm.decode_params(device_id, params)
             hm.send(conn, hm.make_subscription_response(device_id, subscribed_params, delay, uid))
             update_time = time.time()
         if msg.get_message_id() in [hm.MESSAGE_TYPES["Ping"]]:
             # Send a subscription response
-            print("Ping recieved")
+            if args.verbose:
+                print("Ping recieved")
```
kevinji: Ditto.

hibike/virtual_device.py
```diff
@@ -117,7 +136,8 @@ def main():
         if msg.get_message_id() in [hm.MESSAGE_TYPES["DeviceWrite"]]:
             # Write to requested parameters
             # and return the values of the parameters written to using a device data
-            print("Device write recieved")
+            if args.verbose:
+                print("Device write recieved")
```
kevinji: Ditto.

hibike/virtual_device.py
```diff
@@ -89,19 +105,22 @@ def main():
...
             # Send a subscription response
-            print("Ping recieved")
+            if args.verbose:
+                print("Ping recieved")
             hm.send(conn, hm.make_subscription_response(device_id, subscribed_params, delay, uid))
         if msg.get_message_id() in [hm.MESSAGE_TYPES["DeviceRead"]]:
             # Send a device data with the requested param and value tuples
-            print("Device read recieved")
+            if args.verbose:
+                print("Device read recieved")
```
kevinji: Ditto.

hibike/unit_tests/hibike_process_tests.py
baby-bell: This is just an example set of sensors that might be put on a robot. If there's anything that might be factored out, it would be `run_with_random_data`

hibike/hibike_message.py
baby-bell: I'm sorry, I don't quite understand what you mean by that. Which part am I changing into an `if`?

kevinji:
Can you add these tests to Travis as well? You might need to modify the `requirements.txt` to include `unittest` as a dep.

hibike/hibike_message.py
kevinji: I was imagining something like the following:
```python
while not stop_event.is_set():
    # Wait for a 0 byte to appear
    if packets_buffer.find(zero_byte) == -1:
        # ...
        continue

    # ...
```
Although I'm not sure what the correct behavior should be if a stop event has occurred. Should the last "read" packet still be `yield`-ed, like the current code does?

hibike/unit_tests/hibike_message_tests.py
kevinji: You could probably just use `hibike_message.DEVICES.keys()` where necessary, to avoid instantiating a copy of the keys.

baby-bell: I'm afraid that still doesn't work. `dict.keys` gives an iterator, not a sequence.

hibike/hibike_message.py
baby-bell: It's OK to stop immediately when the event is set, rather than yield additional packets. Once calling code sets the event, it's assumed that it doesn't care about any other data. I'll try using this fix, thanks.

tingtingchiang (approved):

baby-bell:
Is there anything else that needs to be resolved, @yizow ?

hibike/TESTS.md
```diff
@@ -0,0 +1,6 @@
+# Running and Writing Tests
+
+Tests are written using Python's `unittest` module. You can read
+more about it [here](https://docs.python.org/3/library/unittest.html).
+All unit tests are located in the `unit_tests` directory. You can run
+them yourself with `python3 -m unittest unit_tests/*.py`.
```
yizow: from what directory/

hibike/hibike_message.py
```diff
@@ -6,6 +6,8 @@
 import struct
 import os
 import json
+import threading
+
```
yizow: extra new line. remove

hibike/hibike_message.py
```diff
@@ -405,19 +407,21 @@ def parse_bytes(msg_bytes):
     return HibikeMessage(message_id, payload)
 
 
-def blocking_read_generator(serial_conn, stop_event=None):
+def blocking_read_generator(serial_conn, stop_event=threading.Event()):
```
yizow, line 409: Interesting quirk with python arg defaults: 
One object is created when the module is imported, and that is used for every single function call. 

So when one threading.Event() for a blocking_read_generator called with default arguments triggers, every single blocking_read_generator called with default arguments will trigger. 

I recommend keeping it None, then have an `if stop_event == None` etc.

hibike/hibike_message.py
```diff
@@ -405,19 +407,21 @@ def parse_bytes(msg_bytes):
...
+    Yield packets from SERIAL_CONN, stopping if STOP_EVENT is set.
     """
     zero_byte = bytes([0])
     packets_buffer = bytearray()
-    while stop_event is None or not stop_event.is_set():
-
+    # Switch to nonblocking mode so that we don't get stuck reading
+    old_timeout = serial_conn.timeout
+    serial_conn.timeout = 0
+    while not stop_event.is_set():
```
yizow, line 418: will this cause a busy wait?

hibike/hibike_message.py
```diff
@@ -405,19 +407,21 @@ def parse_bytes(msg_bytes):
...
-    and is set.
+    Yield packets from SERIAL_CONN, stopping if STOP_EVENT is set.
     """
     zero_byte = bytes([0])
     packets_buffer = bytearray()
-    while stop_event is None or not stop_event.is_set():
-
+    # Switch to nonblocking mode so that we don't get stuck reading
+    old_timeout = serial_conn.timeout
+    serial_conn.timeout = 0
```
yizow, line 417: Is there really a point to manipulating the timeout like this? Who else would change it/ when do we ever want a timeout?

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
+"""
+Unit tests for functions in hibike_message.
```
yizow: maybe change the name of this directory to `hibike_tests` or something similar? I feel like the current name may cause namespace confusion (from the programmer's perspective)

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+import queue
+import time
+import threading
+
+import serial
+
+import hibike_message
+import spawn_virtual_devices
+from unit_tests.utils import run_with_random_data
+DEVICE_TYPES = list(hibike_message.DEVICES)
```
yizow: add new line between imports and object initializations.

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+
+
+def random_uid():
+    """
+    Generate a random UID.
```
yizow: Pep style guide says single line should have quotes on same line
[here](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    uid_type = random.choice(DEVICE_TYPES)
+    return uid_type << 72 | random.getrandbits(72)
+
+
+def random_params(device_id):
+    """
+    Get a random sampling of parameters for a particular DEVICE_TYPE.
```
yizow: see above

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    all_params = hibike_message.all_params_for_device_id(device_id)
+    sample_length = random.randrange(len(all_params) + 1)
+    params = random.sample(all_params, sample_length)
+    return params
+
+
+def random_values(device_id, params):
+    """
+    Generate random valid values for PARAMS.
```
yizow: see above

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    all_params = hibike_message.DEVICES[device_id]["params"]
+    values = []
+    for param in params:
+        param_index = 0
+        for (idx, param_pack) in enumerate(all_params):
+            if param_pack["name"] == param:
+                param_index = idx
+                break
+        param_type = all_params[param_index]["type"]
```
yizow: I'd add an if statement here: 
`if param_index == 0`
better to be explicit, and more readable
add a corresponding `else` case that should never happen, and just throw a generic exception

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        values.append(ParsingTests.VALUE_GENERATORS[param_type]())
+
+    return values
+
+
+class CobsTests(unittest.TestCase):
+    """
+    Tests for COBS encoding and decoding.
```
yizow: See docstring comment above

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+            encoded = hibike_message.cobs_encode(data)
+            decoded = hibike_message.cobs_decode(encoded)
+            self.assertEqual(data, decoded,
+                             "{} encoded incorrectly".format(data))
+        run_with_random_data(assert_encode_decode_equal, CobsTests.gen_cobs_data, times=100)
+
+
+class FakeSerialPort(object):
+    """
+    A fake serial port that acts as a queue.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    def __init__(self):
+        self._buf = queue.Queue()
+
+    def write(self, byte_buf):
+        """
+        Write BYTE_BUF into the queue.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        for byte in byte_buf:
+            self._buf.put(byte)
+
+    def read(self, length=1):
+        """
+        Read out LENGTH bytes from the queue.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        buf = bytearray()
+        for _ in range(length):
+            buf.append(self._buf.get())
+        return buf
+
+    def drain(self):
+        """
+        Read the contents of the queue into a bytearray.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        contents = bytearray()
+        while not self._buf.empty():
+            contents.append(self._buf.get())
+        return contents
+
+    @property
+    def in_waiting(self):
+        """
+        The number of bytes in the queue.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        return self._buf.qsize()
+
+
+class ParamsTests(unittest.TestCase):
+    """
+    Tests for encoding and decoding device parameters.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+            self.assertEqual(sorted(decoded), sorted(param_list),
+                             "encoded/decoded param list {} not identical".format(param_list))
+
+        run_with_random_data(assert_params_same,
+                             self.gen_random_device_id_and_params, times=1000)
+
+
+class ParsingTests(unittest.TestCase):
+    """
+    Tests for parsing Hibike messages.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        cls.VALUE_GENERATORS["uint32_t"] = lambda: random.randrange(2 ** 32)
+        cls.VALUE_GENERATORS["int32_t"] = lambda: random.randrange(-(2 ** 31), 2 ** 31)
+        cls.VALUE_GENERATORS["uint64_t"] = lambda: random.randrange(2 ** 64)
+        cls.VALUE_GENERATORS["int64_t"] = lambda: random.randrange(-(2 ** 63), 2 ** 63)
+        cls.VALUE_GENERATORS["float"] = random.random
+
+    @staticmethod
+    def encode_packet(packet):
+        """
+        Turn a HibikeMessage into an encoded packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        fake_port = FakeSerialPort()
+        hibike_message.send(fake_port, packet)
+        return fake_port.drain()
+
+    @staticmethod
+    def gen_random_sub_request():
+        """
+        Generate a random subscription request packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+    def encode_packet(packet):
+        """
+        Turn a HibikeMessage into an encoded packet.
+        """
+        fake_port = FakeSerialPort()
+        hibike_message.send(fake_port, packet)
+        return fake_port.drain()
+
+    @staticmethod
+    def gen_random_sub_request():
```
yizow: would it perhaps be useful to make these family of functions take arguments, with default values set to something? (maybe `None`, though if you did that, to get the same cleanless you'd have to create another function to wrap the `if None, else random()` check)

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        device_id = random.choice(DEVICE_TYPES)
+        params = random_params(device_id)
+        delay = random.randrange(100)
+        msg = hibike_message.make_subscription_request(device_id, params, delay)
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_sub_response():
+        """
+        Generate a random subscription response packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        device_id = hibike_message.uid_to_device_id(uid)
+        params = random_params(device_id)
+        delay = random.randrange(100)
+        msg = hibike_message.make_subscription_response(device_id, params, delay, uid)
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_device_read():
+        """
+        Generate a random device read packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        device_id = random.choice(DEVICE_TYPES)
+        params = random_params(device_id)
+        msg = hibike_message.make_device_read(device_id, params)
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_device_write():
+        """
+        Generate a random device write packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        device_id = random.choice(DEVICE_TYPES)
+        params = random_params(device_id)
+        values = random_values(device_id, params)
+        msg = hibike_message.make_device_write(device_id, zip(params, values))
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_device_data():
+        """
+        Generate a random device data packet.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        params = random_params(device_id)
+        values = random_values(device_id, params)
+        params_and_values = [(param, value) for param, value in zip(params, values)]
+        msg = hibike_message.make_device_data(device_id, params_and_values)
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_heartbeat_response():
+        """
+        Generate a heartbeat response with a random ID.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        msg = hibike_message.make_heartbeat_response(random.randrange(256))
+        return (ParsingTests.encode_packet(msg), )
+
+    @staticmethod
+    def gen_random_error():
+        """
+        Generate an error packet with a random error code.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+                                 "valid packet {} parsed as None".format(valid_packet))
+
+        for func in self.MESSAGE_GENERATORS.values():
+            if func is not None:
+                run_with_random_data(assert_parse_is_not_none,
+                                     func, times=100)
+
+    def test_parse_bad_checksum(self):
+        """
+        Packets with bad checksums should not be parsed.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+            """
+            if valid_packet[-2] < 255:
+                valid_packet[-2] += 1
+            else:
+                valid_packet[-2] -= 1
+            return valid_packet
+
+        def assert_parse_is_none(invalid_packet):
+            """
+            Assert that parsing an invalid packet results in None.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+
+        for func in self.MESSAGE_GENERATORS.values():
+            if func is not None:
+                # pylint: disable=cell-var-from-loop
+                wrapped_func = lambda: (screw_up_checksum(*func()), )
+                run_with_random_data(assert_parse_is_none, wrapped_func, times=100)
+
+    def test_parse_idempotence(self):
+        """
+        Check that parsing an encoded packet is idempotent.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+                             "parse not idempotent for {}".format(valid_packet))
+
+        for func in self.MESSAGE_GENERATORS.values():
+            if func is not None:
+                run_with_random_data(assert_parse_idempotent, func, times=100)
+
+
+class BlockingReadGeneratorTests(unittest.TestCase):
+    """
+    Tests for blocking_read_generator.
```
yizow: See docstring comment

hibike/unit_tests/hibike_message_tests.py
```diff
@@ -0,0 +1,373 @@
...
+        start_time = time.time()
+        stop_event = threading.Event()
+        for _ in hibike_message.blocking_read_generator(self.dummy_device, stop_event):
+            stop_event.set()
+            self.assertLessEqual(time.time() - start_time, self.STOP_TIMEOUT,
+                                 "blocking_read_generator didn't stop when stop_event triggered")
+
+    def test_read_device_data(self):
+        """
+        Check that DeviceData packets are received and decoded.
```
yizow: See docstring comment

hibike/unit_tests/hibike_process_tests.py
```diff
@@ -0,0 +1,65 @@
...
+        Assert that IDENTIFIED_DEVICES contains all devices in
+        VIRTUAL_DEVICE_TYPES.
+        """
+        device_ids = map(hm.uid_to_device_id, identified_devices.values())
+        found_types = map(lambda dev: hm.DEVICES[dev]["name"], device_ids)
+        self.assertListEqual(sorted(self.VIRTUAL_DEVICE_TYPES), sorted(found_types), msg)
+
+    def test_detect_devices(self):
+        """
+        Test detection of valid devices.
```
yizow: See docstring comment

hibike/unit_tests/hibike_process_tests.py
```diff
@@ -0,0 +1,65 @@
...
+        for vdev_type in self.VIRTUAL_DEVICE_TYPES:
+            virtual_devices.append(serial.Serial(spawn_device(vdev_type)))
+        # Wait for virtual devices to spin up
+        time.sleep(self.VIRTUAL_DEVICE_STARTUP_TIME)
+        found = identify_smart_sensors(virtual_devices)
+        self.assert_all_devices_identified(found, "did not identify all sensors")
+
+    def test_detect_no_devices(self):
+        """
+        Make sure that we don't detect empty serial ports as sensors.
```
yizow: See docstring comment

hibike/virtual_device.py
```diff
@@ -26,11 +26,22 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('-d', '--device', required=True, help='device type')
     parser.add_argument('-p', '--port', required=True, help='serial port')
+    parser.add_argument('-v', '--verbose',
+                        help='print messages when sending and receiving packets',
+                        action="store_true")
     args = parser.parse_args()
 
+    def verbose_log(fmt_string, *fmt_args):
```
yizow, line 34: I like this.
we should think about pulling this out into some general utilities shared between hibike + runtime. although that may be too much effort.

yizow (dismissed):

hibike/hibike_message.py
baby-bell: It's useful if you want to do blocking reads, then `timeout = None`. I don't want to add more preconditions.

hibike/hibike_message.py
baby-bell: This shouldn't be an issue, because the event is never triggered within `blocking_read_generator`,

hibike/hibike_message.py
baby-bell: If there is no data coming down the serial pipeline, then yes. There isn't really a way to fix this without reworking Hibike's design.

hibike/unit_tests/hibike_message_tests.py
baby-bell: Done

hibike/unit_tests/hibike_message_tests.py
baby-bell: Hm, not quite sure why this would be better. More flexibility?

baby-bell:
Okay, I think that's pretty much everything. Is there anything else that needs to be resolved, @yizow? I appreciate all the constructive criticism so far.

PR #392, opened 2017-08-16
[HIBIKE] Add note for changing ard-reset-arduino

onibrow:
Add a note that the change in `ard-reset-arduino` might not be necessary depending on what Python version is installed.


onibrow:
jk this is useless ignore me [this](https://github.com/sudar/Arduino-Makefile/issues/407)

Issue #391, opened 2017-08-15
[HIBIKE] add a --help flag to ./flash.sh that prints the device names


viviehn:
Because it's bothersome when you forget what they are, and it would be nice to not have to return to the README every time you forget

dHutchings:
Last year, I got so fed up with hibike's flash script that I implemented tab-completion:

https://github.com/pioneers/PieCentral/pull/182

Just mash tab & it'll tell you the options.  Works on ubuntu, should work on arch (for seiya et al) (but I can't recall if it did or not).  You'd have to re-run the install script, if you last ran it before March 3 2017.

Should be quite easy to add a -h option to the flashing script, and have it give you the output of ./utils/device_names.sh

baby-bell:
Should be addressed by #393

PR #393, opened 2017-08-16
[HIBIKE] More friendly flasher

baby-bell:
A nice help message for the flashing script. Also checks that the device type is valid, and gives a usage message. Fixes issue #391.


hibike/flash.sh
```diff
@@ -1,10 +1,57 @@
-#/bin/bash
+#!/bin/bash
 
+NAME="$0"
+SENSOR_TYPES="$(ls devices/)"
```
dHutchings: there's a script to list sensor types, actually: /utils/device_names.sh.

It's true that it does the same thing as line 4.  But it's the script that ubuntu call when it tries to tab-complete.

probably should use the same script.

baby-bell: Thanks for the feedback. Apparently it's bad form to rely on the output of `ls` anyways, so I will change this script to use your script, and your script to use globs.

viviehn (approved):

PR #394, opened 2017-08-19
[HIBIKE] Cleanly shutdown Hibike

baby-bell:
Instead of terminating the process, it's possible to have Hibike monitor an event and shut itself down nicely. This makes testing and profiling somewhat easier, albeit at the cost of some added complexity.


baby-bell:
I am a little bit uncomfortable with the amount of complexity this adds.

PR #402, opened 2017-09-09
[HIBIKE] Fixing flash.sh to actually find the correct sensors

tingtingchiang:
flash.sh actually did not scan the list of devices to see if something was flash-able.


jonathan-j-lee (approved):
LGTM

PR #404, opened 2017-09-14
[RUNTIME/HIBIKE] Use pipenv instead of requirements.txt

kevinji:
pipenv manages dependencies more like npm.


baby-bell:
I'm afraid that you've broken `pylint` for Hibike. Another question: does switching to Python 3.6 break anything? I'm still running Python 3.5, and would prefer to keep Hibike working on it.

baby-bell (changes requested):
Need to update Hibike makefile as well. Also, what are the benefits of this change?

kevinji:
Addressing your comments:

- I don't think Python 3.6 adds anything significant; I'm just running that on my machine, so I think Pipfile just defaulted to that.
- pipenv is now the [official recommended package manager](https://packaging.python.org/new-tutorials/installing-and-using-packages/) according to the docs. The nice thing about `pipenv` is that it automatically creates a different `virtualenv` for every Python project that uses it, so if two projects require different versions of the same library, those can co-exist without being installed as a global Python dependency.

I'll see if I can fix the build issues; I was trying to debug the issue earlier.

baby-bell:
lgtm, I'm not amazing w/ bash so I will ask someone else for review on that

tingtingchiang (approved):
Not familiar with travis or pipfiles but the bash scripts look good to me.

baby-bell (approved):

PR #408, opened 2017-09-23
[HIBIKE] Fix flashing script, remove reset.py

baby-bell:
The reset code has broken, and it hasn't been needed for quite a while now. Flashing works without it
as long as you run `setup.sh`.


tingtingchiang (approved):
LGTM

PR #409, opened 2017-09-23
[RUNTIME] Changes Version for Stack Trace Change

bodiddles:
Increments patch number of the version to indicate that runtime has changed what it reports in its stacktrace if there is an error in studentCode.


brandonxxlee:
To be honest, maybe we should just increment the MAJOR number for this entire year. Alternatively we don't forget to change it before deployment

runtime/runtimeUtil.py
```diff
@@ -14,7 +14,7 @@ class RUNTIME_CONFIG(Enum):
     TEST_OUTPUT_DIR             = "test_outputs/"
     VERSION_MAJOR               = 1
```
brandonxxlee, line 15: Lets just change major to 2, in preparation for full release next season.

brandonxxlee (changes requested):

runtime/runtimeUtil.py
bodiddles: Should keep for internal use to distinguish between runtime changes to make sure newest code is utilized

brandonxxlee (approved):

PR #410, opened 2017-09-23
[RUNTIME] Change runtimeUtil to have BAD_EVENTS include timestamps mssgs

renegamino012:
Modified runtimeUtil.py as to include timestamp events - TIMESTAMP_DOWN & TIMESTAMP_UP - and their messages - "Latency test ... - ... this stack".

### DIFF
--- a/runtime/runtimeUtil.py
+++ b/runtime/runtimeUtil.py
@@ -35,6 +35,8 @@ class BAD_EVENTS(Enum):
     ENTER_IDLE                = "Dawn says enter Idle"
     NEW_IP                    = "Connected to new instance of Dawn"
     DAWN_DISCONNECTED         = "Disconnected to Dawn"
+    TIMESTAMP_DOWN             = "Latency test down, down this stack"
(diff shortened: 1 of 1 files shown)


runtime/runtimeUtil.py
```diff
@@ -35,6 +35,8 @@ class BAD_EVENTS(Enum):
     ENTER_IDLE                = "Dawn says enter Idle"
     NEW_IP                    = "Connected to new instance of Dawn"
     DAWN_DISCONNECTED         = "Disconnected to Dawn"
+    TIMESTAMP_DOWN             = "Latency test down, down this stack"
```
tingtingchiang: we probably don't need the second down, so just "Latency test down this the stack" would be clearer. Also try and make the "=" line up with the rest if possible

tingtingchiang:
Also FYI, convention for creating branches is <group-name>/<specific-issue-or-fix-name>. So in this case, you would want to create the branch "runtime/bad-events-messages" or something similar.

tingtingchiang (approved):
LGTM

PR #412, opened 2017-09-28
[RUNTIME] Handle timestamp_UP/DOWN in runtime

mrduston:
This PR enables runtime to attach timestamps to timestamp packets and print timestamps when a timestamp up packet is received.


baby-bell (approved):
It's best not to disable lints in general, but probably worth it in this situation. LGTM

PR #416, opened 2017-10-07
[Hibike] add timestamps to Hibike 

jamesleng100:


baby-bell (approved):
Looks good to me, in the future, we should refactor Hibike to use the constants in `runtimeUtil` in the future

PR #420, opened 2017-10-07
Aggregate Buttons

ArjunSridhar1:
Aggregated the file operation buttons except for upload and download from the robot buttons. Reduces the total number of buttons.


kevinmasd (approved):
Seems to work

PR #423, opened 2017-10-12
[SHEPHERD] Add utils file for headers

brandonxxlee:
Each header is autoenumerated

### DIFF
--- a/shepherd/Utils.py
+++ b/shepherd/Utils.py
@@ -2,49 +2,49 @@ from enum import Enum, unique, auto
 
 @unique # pylint: disable=invalid-name
 class SHEPHERD_HEADER(Enum):
-    GOAL_SCORE = auto()
-    GOAL_BID = auto()
-    CODE_INPUT = auto()
+    GOAL_SCORE = "goal_score"
+    GOAL_BID = "goal_bid"
+    CODE_INPUT = "code_input"
 
-    START_MATCH = auto()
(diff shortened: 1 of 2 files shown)


shepherd/Utils.py
```diff
@@ -0,0 +1,42 @@
+from enum import Enum, unique
+
+class Header(Enum):
+    # Used to autoenumerate
```
FIREdog5: Seems like this should be described with a doc string, but it really doesn't matter.

shepherd/Utils.py
```diff
@@ -0,0 +1,42 @@
...
+    def __new__(cls):
+        value = len(cls.__members__) + 1 # pylint: disable=no-member
+        obj = object.__new__(cls)
+        obj._value_ = value # pylint: disable=protected-access
+        return obj
+
+@unique # pylint: disable=invalid-name
+class SHEPHERD_HEADER(Header):
```
FIREdog5: It seems like there should be some sort of documentation associated with these headers... again tho, if it's just for us then it doesn't really matter.

FIREdog5 (dismissed):
looks good enough to me.

bodiddles (dismissed):
lgtm

JHGZhou (dismissed):
Might have to consider not using auto, but is usable for now

JHGZhou (approved):
I'm lowkey approving some of my own changes. But thats fine

PR #430, opened 2017-10-14
[RUNTIME] Implement full stack timestamps

mrduston:
This PR adds a timestamp command that goes from Dawn to Hibike and back. Currently the timestamp is implemented using `time.monotonic`, may be better idea to change to something else with more precision in the future. Runtime itself is technically not included.

### DIFF
--- a/runtime/runtimeUtil.py
+++ b/runtime/runtimeUtil.py
@@ -35,8 +35,6 @@ class BAD_EVENTS(Enum):
     ENTER_IDLE                = "Dawn says enter Idle"
     NEW_IP                    = "Connected to new instance of Dawn"
     DAWN_DISCONNECTED         = "Disconnected to Dawn"
-    TIMESTAMP_DOWN            = "Latency test down the stack"
-    TIMESTAMP_UP              = "Latency test up the stack"
 
 restartEvents = [BAD_EVENTS.STUDENT_CODE_VALUE_ERROR, BAD_EVENTS.STUDENT_CODE_ERROR,
                  BAD_EVENTS.STUDENT_CODE_TIMEOUT, BAD_EVENTS.END_EVENT, BAD_EVENTS.EMERGENCY_STOP]
--- a/runtime/runtimeUtil.py
+++ b/runtime/runtimeUtil.py
@@ -73,6 +73,8 @@ class HIBIKE_RESPONSE(Enum):
 class ANSIBLE_COMMANDS(Enum):
     STUDENT_UPLOAD = "student_upload"
     CONSOLE        = "console"
+    TIMESTAMP_UP = "timestamp up"
+    TIMESTAMP_DOWN = "timestamp down"
 
 @unique
 class SM_COMMANDS(Enum):
--- a/ansible-protos/ansible.proto
+++ b/ansible-protos/ansible.proto
@@ -24,4 +24,5 @@ message DawnData {
     repeated Gamepad gamepads = 2;
     repeated string peripheral_names = 3;
     TeamColor team_color = 4;
+    float dawn_timestamp = 5;
 }
--- a/ansible-protos/ansible.proto
+++ b/ansible-protos/ansible.proto
@@ -24,5 +24,4 @@ message DawnData {
     repeated Gamepad gamepads = 2;
     repeated string peripheral_names = 3;
     TeamColor team_color = 4;
-    float dawn_timestamp = 5;
 }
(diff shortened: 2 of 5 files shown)


runtime/stateManager.py
```diff
@@ -288,9 +289,11 @@ def hibike_response_device_disconnect(self, uid):
         del devs[uid]
 
     def hibike_response_timestamp_up(self, *data):
+        print("We have received timestamp_up")
```
baby-bell: Might be a good idea to remove this, since we already know it's working when dawn receives a timestamp back.

baby-bell (commented):
Looks alright to me. I defer to Bob on most of the changes.

runtime/Ansible.py
```diff
@@ -403,6 +403,21 @@ def package_confirm(confirm):
                         str(e),
                         event=BAD_EVENTS.TCP_ERROR,
                         printStackTrace=True))
+        def package_timestamp(timestamp):
```
bodiddles: Change argument name to 'timestamps' to indicate that this variable is a list of multiple timestamps instead of a singular timestamp.

runtime/Ansible.py
```diff
@@ -452,6 +469,11 @@ def unpackage(data):
                             printStackTrace=False))
                     break
                 unpackaged_data = unpackage(recv_data) # pylint: disable=unused-variable
+                if unpackaged_data.header == notification_pb2.Notification.TIMESTAMP_DOWN:
+                    timestamps = list(unpackaged_data.timestamps)
+                    timestamps.append(time.monotonic())
+                    state_queue.put([HIBIKE_COMMANDS.TIMESTAMP_DOWN, timestamps])
+                    continue
                 state_queue.put([SM_COMMANDS.STUDENT_UPLOAD, []])
```
bodiddles: Since we are handling messages that are not just messages informing runtime that student code was uploaded, we should add a check for this line and only send to state manager if we actually receive a student upload command. We should check if the header is equivalent to the STUDENT_SENT enum defined in notification.proto.

runtime/fake_dawn.py
```diff
@@ -43,13 +43,14 @@ def receiver(port, receive_queue):
         runtime_message = runtime_pb2.RuntimeData()
         runtime_message.ParseFromString(msg)
         receive_queue[0] = msg
-        print(runtime_message)
+        #print(runtime_message)
```
bodiddles: Should remove this line or keep it. Shouldn't just keep it commented out on master

runtime/runtimeUtil.py
```diff
@@ -73,6 +71,8 @@ class HIBIKE_RESPONSE(Enum):
 class ANSIBLE_COMMANDS(Enum):
     STUDENT_UPLOAD = "student_upload"
     CONSOLE        = "console"
+    TIMESTAMP_UP = "timestamp up"
```
bodiddles: I think it would be worth it to copy the original strings informing future viewers what these actually mean.

bodiddles (changes requested):
Some minor changes for clarity but otherwise looks good.

runtime/Ansible.py
```diff
@@ -403,6 +403,21 @@ def package_confirm(confirm):
                         str(e),
                         event=BAD_EVENTS.TCP_ERROR,
                         printStackTrace=True))
+        def package_timestamps(timestamp):
```
bodiddles: Sorry if this wasn't clear but we should change the argument name from timestamp to timestamps and change the rest of the function to reflect this change

runtime/runtimeUtil.py
```diff
@@ -73,6 +71,8 @@ class HIBIKE_RESPONSE(Enum):
 class ANSIBLE_COMMANDS(Enum):
     STUDENT_UPLOAD = "student_upload"
     CONSOLE        = "console"
+    TIMESTAMP_UP = "Get timestamps going up the stack"
```
kevinji: This looks a bit off unaligned.

bodiddles (approved):

runtime/fake_dawn.py
```diff
@@ -43,21 +43,26 @@ def receiver(port, receive_queue):
...
+    for i in range(10):
+        msg = notification_pb2.Notification()
+        msg.header = notification_pb2.Notification.TIMESTAMP_DOWN
+        msg.timestamps.append(time.perf_counter())
+        msg = msg.SerializeToString()
+        msgqueue.put(msg)
+    return msgqueue
 
-def tcp_relay(port):
+def tcp_relay(port, msgqueue=queue.Queue()):
```
kevinji, line 56: This default parameter seems like a possible source of bugs, as it's sharing the same queue among multiple calls to `tcp_relay`; is the optional parameter necessary?

PR #434, opened 2017-10-14
Update Packages; Lint Auto-Fix; DevTools Added

kevinmasd:
* Before any more development, updated the Node packages (i.e. `yarn upgrade --latest`)
  * Tested to make sure nothing broke
* Let `eslint` fix some lint mismatches
  * Fixed unused state problems in `Editor.js` and `ConfigBox.js`
* Added `electron-devtools-installer` to help install Dev Tools for React and Redux


.travis.yml
```diff
@@ -11,8 +11,9 @@ env:
   - TEST_DIR=runtime
 cache:
   pip: true
+  yarn: true
```
kevinji, line 14: I don't think this works, as the language used for Travis is `python`; that's why the yarn directory was previously explicitly spelled out.

kevinmasd: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

ArjunSridhar1 (approved):
hopefully you didn't break anything

.travis.yml
kevinji: I know they've made that post, but that only applies if your language is `node_js`; if it's not, then the `yarn: true` does nothing.

Before change: https://travis-ci.org/pioneers/PieCentral/jobs/294230172
After change: https://travis-ci.org/pioneers/PieCentral/jobs/294243830

Note that the
```
adding /home/travis/.cache/yarn to cache
creating directory /home/travis/.cache/yarn
```
lines have disappeared in "after change" build logs, under the "Setting up build cache" section.

PR #438, opened 2017-10-26
[SHEPHERD] Change constants back to numbers

brandonxxlee:


shepherd/Utils.py
```diff
@@ -35,11 +35,12 @@ class SCOREBOARD_HEADER(Enum):
...
-    ZERO_X_COOLDOWN = "zero_x_cooldown"
-    STEAL_COOLDOWN = "steal_cooldown"
-    CODE_COOLDOWN = "code_cooldown"
-    BID_INCREASE_CONSTANT = "bid_increase_constant"
+    # Check with GDC for actual numbers
+    TWO_X_COOLDOWN = 5
+    ZERO_X_COOLDOWN = 5
+    STEAL_COOLDOWN = 5
+    CODE_COOLDOWN = 5
+    BID_INCREASE_CONSTANT = 5
```
kevinji, line 42: I think you're going to need unique numbers here, or you'll have to remove the `@unique` before the function; otherwise, Python will throw a ValueError.

brandonxxlee: Good catch, fixed

brandonxxlee:
Will be fixed by @FIREdog5 closing this PR

PR #447, opened 2017-11-02
[HIBIKE] Updated Enable pin and implemented error state

bhargavi99:
- Changed the Enable pin to 21 and enableAnalog to A3
- implemented error state by adding new error_state parameter and checking value of voltage


hibike/devices/YogiBear/YogiBear.cpp
```diff
@@ -222,6 +222,16 @@ uint8_t device_read(uint8_t param, uint8_t* data_update_buf, size_t buf_len) {
       return sizeof(float);
       break;
 
+    case ERROR_STATE:
+      bool errorState = false;
+      int voltageReading = analogRead(current_pin);
```
dHutchings: Hi!  My name's Doug.  I'm old.  I used to do a bunch of stuff for PiE, and I still occasionally poke around PRs.

I'd be happy to get into a proper review if you'd like.

But IRRC I'm pretty sure that it shouldn't be "current Pin" on line 227.  Shouldn't it be the enable pin?

dHutchings: Alright, now to get technical.

I don't think that you can do an analogRead on the enable pin when it's configured to be an output.  You'll have to configure it as an analog input before you can actually read the voltage.  The Arduino IDE won't yell at you if you do, you'll just get (possibly) unsensical data.  This is something that should be tested.

hibike/devices/YogiBear/motor.cpp
```diff
@@ -22,12 +24,12 @@ void motorSetup() {
 
 void motorEnable() {
   clearFault();
-  digitalWrite(enable_pin, HIGH);
+  analogWrite(enable_pin, analogHIGH); /* digital to analog where HIGH corresponds to 255*/
```
dHutchings: So why did we change these lines from digitalWrite to analogWrite?

There's no reason to.  If you analogWrite to a pin, you'll have to still configure it as an input before you analog read it.  You can't configure the pin to be an analog/digital pin...

dHutchings (dismissed):
Hi.

I don't have too many comments about the specific code (although I do have a few, see below.  Mainly that as implimented, the measurement of if we're in an error state potentially obliterates any previous disabling of the motor driver.).

I have far more questions about the intent & design of this feature.  I'm sorry that I only picked up on this now that I saw the commit, and i'm sorry that I'm not there to ask these questions in person.

Specifically, I have questions how the error state is integrated into the rest of the yogibear, and how the error state is integrated into the larger control system.

For example, the yogibear is an intelligent thing that periodically pays attention to the encoder / red driver board on its own volition.  For example, it does current limiting by itself at 1000 Hz.  So why does the yogibear only check if the motor is in an error state if it is explicitly requested to do so as part of device_read?  (Yogibear.cpp, line 229).

Furthermore, if there is indeed an error... should the yogibear do nothing about it?  Do we just inform the higher-level control system and expect the user to deal with it?  Or should the yogibear disable itself?  Should the yogibear adjust several of the LEDs to inform them that something bad is happening?  We already use the LEDs to display if the yogibear is disabled (All three lights coming on).

Additionally, the implementation of physically measuring the error state is suspect.  Currently, this is done by taking the pin that is enabling or disabling the device, and using it as an analog measurement.

Therefore, it means if we are disabling the motor (which is done by writing the enable pin low), we'd have to stop disabling the motor to read the analog voltage on the enable pin to check if we're in an error state.  Something about that plan sounds suspect.


Other questions: will state manager be polling the yogibear is_error state?  If so, at what rate?

hibike/devices/YogiBear/motor.cpp
dHutchings: I should also point this out.

The enable_pin is pulled up with a resistor on the red board.  So I think we should actually put this pin into a High Impedance state (aka make it an analog/digital input) to enable the pin, not write the value high.

Let's pretend that we could both do analog_writes and analog_reads at the same time (we can't).  If we could, once we enabled the motor, what analog value would we see?  It'd always be high: regardless of what the motor's trying to do, even if it's trying to pull the line low to signal it's in an error state.

dHutchings (commented):
Also would point this out.

I think that on the electrical level, the PCBs have been set?  Dunno.

I would like to point out that the code would most likely be simplified if we could actually get _two_ Arduino Pins hooked up to the same line from the red board.  One serving as an analog input to check if we're in error or not, the other serving as a digital output low / High-Z (aka open collector) output to do the enable / disable.

Not sure if the PCBs have been locked down.

baby-bell:
@dHutchings do you mind taking another look at this? I'd like to make sure major changes are all merged by the end of the year.

tingtingchiang:
@dHutchings can you take a look at the changes when you have the time? We are looking to try and get things merged in a week or two so we do tests with the new code too.

baby-bell:
Okay, it's been long enough that I am going to override Doug's review and we will test with the new code.

hibike/devices/YogiBear/motor.cpp
```diff
@@ -67,13 +67,14 @@ void drive(float target) {
...
   digitalWrite(INA, !digitalRead(INA));
   digitalWrite(INB, !digitalRead(INB));
-  
+
   digitalWrite(INA, !digitalRead(INA));
   digitalWrite(INB, !digitalRead(INB));
 
   Timer1.pwm(PWM, 0);
+  pinMode(enable_pin, OUTPUT);
   digitalWrite(enable_pin, LOW);
```
dHutchings, line 76: you may want to throw in a few millisecond's worth of delay here.  an experimental test will tell you.

dHutchings (approved):
I know you've been waiting for month for this, but I *htink * lgtm.

baby-bell:
wonderful, thanks doug!

PR #450, opened 2017-11-04
[DevOps] Log download script

Benji410:
Finished script for obtaining systemd log scripts from beagle bone. It will require two inputs of the beagle bones password, one to ssh into the beagle bone and the second to secure copy the file over to the host computer.

### DIFF
--- a/DevOps/frankfurter/scripts/tools/sendlogs.sh
+++ b/DevOps/frankfurter/scripts/tools/sendlogs.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+mkdir temp && cd temp
+
+journalctl -e -u update.service > update.log
+journalctl -e -u runtime.service > service.log
+journalctl -e -u network.service > network.log
+
+
--- a/DevOps/frankfurter/scripts/tools/sendlogs.sh
+++ b/DevOps/frankfurter/scripts/tools/sendlogs.sh
@@ -1,9 +1,15 @@
 #!/bin/bash
 
+#make sure to install zip command using sudo apt install.. 
+
 mkdir temp && cd temp
 
 journalctl -e -u update.service > update.log
 journalctl -e -u runtime.service > service.log
-journalctl -e -u network.service > network.log
+journalctl -e -u networking.service > networking.log
+
+cd $HOME 
+
+
 
 
--- a/DevOps/frankfurter/scripts/tools/sendlogs.sh
+++ b/DevOps/frankfurter/scripts/tools/sendlogs.sh
@@ -1,14 +1,22 @@
 #!/bin/bash
 
 #make sure to install zip command using sudo apt install.. 
+#added zip pkg to master_frankfurter setup
 
 mkdir temp && cd temp
 
+#get the logs from these services and log them in temp directory
 journalctl -e -u update.service > update.log
 journalctl -e -u runtime.service > service.log
 journalctl -e -u networking.service > networking.log
 
 cd $HOME 
+
+#zip log files and print out zip file if it was made
+zip -r service-logs.zip temp && ls *.zip
+
+#send files to host computer
+
 
 
 
--- a/DevOps/frankfurter/README.md
+++ b/DevOps/frankfurter/README.md
@@ -28,7 +28,7 @@ Alternatively,
 1. Download a [stock Ubuntu prebuilt image](http://elinux.org/BeagleBoardUbuntu#eMMC:_All_BeagleBone_Varients_with_eMMC) (preferably 16.04).
 2. Follow the instructions given by the link in step 1 to bring up a Beaglebone running stock Ubuntu.
 3. Be sure you can `ssh ubuntu@192.168.7.2`. See the [networking section](#networking) for setup instructions.
-4. Connect the Beaglebone to the internet. You should be able to get a response from `ping 8.8.8.8`, but not necessarily from `ping google.com`. One way to do this is to run  `frankfurter/scripts/toos/usb-fwd.sh`, possibly as `sudo`.
+4. Connect the Beaglebone to the internet. You should be able to get a response from `ping 8.8.8.8`, but not necessarily from `ping google.com`. One way to do this is to run  `frankfurter/scripts/tools/usb-fwd.sh`, possibly as `sudo`.
 5. On your machine, run `frankfurter/scripts/install/start_frank_install.sh`. Type in `ubuntu`'s password as many times as is necessary.
 6. `ssh` in, and run `start_tmux_job.sh`. Again, enter `ubuntu`'s password as necessary.
 7. After the kernel is upgraded (to ensure `linux-headers-$(uname -r)` is available), the Beaglebone will automatically reboot. As instructed, reconnect the Beaglebone to the internet after the reboot, `ssh` in, and run `tmux new-session -d '~/PieCentral/DevOps/frankfurter/scripts/install/master_frank_setup.sh'`.
--- a/DevOps/frankfurter/scripts/install/master_frank_setup.sh
+++ b/DevOps/frankfurter/scripts/install/master_frank_setup.sh
@@ -27,7 +27,7 @@ alias sudo='sudo -H'
 
 # Install apt packages
 sudo apt update -y && sudo apt upgrade -y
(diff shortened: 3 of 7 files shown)


DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+#make sure to install zip command using sudo apt install.. 
+#added zip pkg to master_frankfurter setup
+
+HOST=hostname
```
jonathan-j-lee: This doesn't look it's being used, and should be removed.

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,40 @@
...
+journalctl -e -u update.service > update.log
+journalctl -e -u runtime.service > service.log
+journalctl -e -u networking.service > networking.log
+bzip2 * 
+
+cd 
+
+find temp -not -name *bz2 | xargs tar -jcvf archivelogs.tar.bz2 
+
+exit "
```
jonathan-j-lee: Doesn't look like `temp` is deleted, so it will reside in `~` forever, which causes `mkdir` to fail if you run the script twice on the same machine (`mkdir -p` succeeds if a directory already exists).

My suggestion is to use a `/tmp/log-fetch` subdirectory when preparing the file, since Ubuntu cleans up old files in `tmp` for you.

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,40 @@
...
+#exit ssh
+
+ssh $HOME@$USER "
+mkdir temp
+cd temp
+
+journalctl -e -u update.service > update.log
+journalctl -e -u runtime.service > service.log
+journalctl -e -u networking.service > networking.log
+bzip2 *
```
jonathan-j-lee: The logs seem to be intact, but the structure seems really weird. Try `tar`-ing the directory first (turn many files into a single one), then `bzip2` that file, rather than the reverse order. Alternatively, use `zip`.

jonathan-j-lee (dismissed):
My notes so far.

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,32 @@
...
+
+#format: ssh ubuntu 192.168.7.2 "Linux commands"
+#makes a temp directory and goes into the temp directory once made
+#gets the logs from the services in the beaglebone and logs in temp directory
+#zips log files and prints the zip file name if it has been created
+#exit ssh
+
+ssh $HOME@$USER "
+mkdir -p /tmp/log-fetch
+cd /temp/log-fetch
```
rmandla: fix typo 
/tmp/log-fetch vs /temp/log-fetch

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cd 
+
+tar -cjvf archivelogs.tar.bz2 tmp 
+
+exit "
+
+#copy and send zip files from remote to local computer
+scp $HOME@$USER:archivelogs.tar.bz2 .
+
```
rmandla: you also want to delete the files you made from the beaglebone

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,32 @@
...
+#exit ssh
+
+ssh $HOME@$USER "
+mkdir -p /tmp/log-fetch
+cd /temp/log-fetch
+
+journalctl -e -u update.service > update.log
+journalctl -e -u runtime.service > service.log
+journalctl -e -u networking.service > networking.log 
+
```
rmandla: try adding a time-stamp to the name of each log so that it is easier to recognize

rmandla (changes requested):
looks like it is almost good to go, but a few changes are still needed

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+ssh $HOME@$USER "
+mkdir -p /tmp/log-fetch
+cd /tmp/log-fetch
+
+today=$(date +%Y-%m-%d.%H:%M:%S)
+
+journalctl -e -u update.service > update.log"."$today
+journalctl -e -u runtime.service > service.log"."$today
+journalctl -e -u networking.service > networking.log"."$today
+
```
rmandla: time-stamp doesn't work. Check the update script (create_update.sh) to see how we have been time-stamping files. And you should time-stamp the archivelogs.tar file also.

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+cd 
+
+tar -cjvf archivelogs.tar.bz2 tmp 
+
```
rmandla: you cd back to home, and then try to tar a file tmp within home, which doesn't exist. Should be /tmp/log-fetch (just /tmp would tar the entire /tmp folder)

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+exit "
+
+#copy and send zip files from remote to local computer 
+#if successful, log back into beaglebone and rm file
+scp $HOME@$USER:archivelogs.tar.bz2 . && ssh$HOME@$USER "
```
rmandla: add space between ssh and $home

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+rm archivelogs.tar.bz2 
+exit "
+
```
rmandla: currently this script requires someone to log in to the beaglebone three times, to make the logs, to transfer the logs, then to delete the logs. We can probably cut this down to two using rsync to combine transfering and deleting to one step
https://serverfault.com/questions/160799/scp-files-and-delete-files-on-remote-directory

rmandla (changes requested):

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+#unzip service-logs.zip -d service-logs
```
rmandla: don't need this anymore

DevOps/frankfurter/scripts/tools/sendlogs.sh
```diff
@@ -0,0 +1,37 @@
...
+
+#copy and send zip files from remote to local computer 
+#if successful, log back into beaglebone and rm file
+scp $HOME@$USER:archivelogs.tar.bz2 . && ssh$HOME@$USER "
+rm archivelogs.tar.bz2 
+exit "
+
+#unzip service-logs.zip -d service-logs 
+bzip2 -d archivelogs.tar.bz2 
+tar -xf archivelogs.tar
```
rmandla: you are also going to want to delete the archivelogs.tar file after tar has been run

DevOps/frankfurter/scripts/tools/sendlogs.sh
rmandla: actually, after you tar the archivelogs.tar file, it outputs the logs within the folder log-fetch, which is within the folder tmp if bzip is run on /tmp/log-fetch. But it would be more helpful if only the log-fetch folder was outputted rather than having a tmp folder be outputted, with the log-fetch folder within that tmp folder.

rmandla (changes requested):

DevOps/frankfurter/scripts/tools/sendlogs.sh
rmandla: sorry, you don't need to time-stamp the archivelogs.tar file since it should be deleted after the script runs

rmandla (approved):

PR #457, opened 2017-11-09
[SHEPHERD] Filled in Alliance.py and Goal.py methods

FIREdog5:
Added the bodies of the methods specified in the API for both the Goal and the Alliance classes. Also added a new POWERUP_TYPES enum, and added to the CONSTANTS enum. Also removed @unique from the CONSTANTS enum. Lastly added a temporary LCM.py file.


shepherd/Goal.py
```diff
@@ -47,13 +48,23 @@ def score(self, alliance):
...
+        if self.owner is alliance:
+            alliance.change_score(self.value * self.calc_multiplier() *\
+                alliance.alliance_multiplier)
 
     def calc_multiplier(self):
         """ returns an integer that represents the overall multiplier for this
             goal, by checking each timer in the goal
         """
-        pass
+        if (self is None or self.gold_two_x_timer is None or
```
bodiddles: Is it possible for any of these values to be None? It seems like they are all initialized in the __init__ method. Additionally, I don't think it is possible for self to be None.

shepherd/Goal.py
```diff
@@ -47,13 +48,23 @@ def score(self, alliance):
             scorring Alliance is the owning alliance.
                 alliance - Alliance object representing the scoring team
```
brandonxxlee: This comment is indented too far

shepherd/Goal.py
```diff
@@ -47,13 +48,23 @@ def score(self, alliance):
...
         """
-        pass
+        if self.owner is alliance:
+            alliance.change_score(self.value * self.calc_multiplier() *\
```
brandonxxlee: I think this would be better as multiple lines instead of one extra long line.

shepherd/Goal.py
brandonxxlee: How could self be None. Also, is this check even necessary? I think the second case does/should just handle no active timers.

shepherd/Goal.py
```diff
@@ -67,4 +78,50 @@ def apply_powerup(self, effect, alliance):
             effect - enum representing an effect to be applied to the goal
             alliance - the alliance trying to apply the effect to the goal
         """
-        pass
+        if effect == POWERUP_TYPES.TWO_X:
+            if alliance.name == CONSTANTS.BLUE:
+                if self.blue_two_x_timer.is_running():
+                    return -1
```
bodiddles: Not really a part of this pull request but is there a reason that we are returning integers defined in the above docstring instead of potential global constants. I think changing this would help with the readability of this code and would make the code that calls this method more clear as well.

shepherd/Goal.py
```diff
@@ -47,13 +48,23 @@ def score(self, alliance):
...
         """ returns an integer that represents the overall multiplier for this
             goal, by checking each timer in the goal
         """
-        pass
+        if (self is None or self.gold_two_x_timer is None or
+                self.blue_two_x_timer is None or self.gold_zero_x_timer is None or
+                self.blue_zero_x_timer is None):
+            return 1
+
+        return (self.gold_two_x_timer.is_running() and 2 or 1) *\
```
brandonxxlee: I don't like everything in one line, what does @JHGZhou thing?

bodiddles: Even though this works fine, I think there should be a more clear way to incorporate this behavior instead of relying on the reader to do boolean logic in their head.

shepherd/Goal.py
```diff
@@ -67,4 +78,50 @@ def apply_powerup(self, effect, alliance):
...
-        pass
+        if effect == POWERUP_TYPES.TWO_X:
+            if alliance.name == CONSTANTS.BLUE:
+                if self.blue_two_x_timer.is_running():
+                    return -1
+                self.blue_two_x_timer.start_timer(CONSTANTS.TWO_X_COOLDOWN)
+                lcm_send(CONSTANTS.LCM_TARGET_SCOREBOARD,
+                         SCOREBOARD_HEADER.POWERUPS,
+                         [self.name, alliance.name, POWERUP_TYPES.TWO_X])
+                return 1
```
brandonxxlee: Why are we returning anything? @JHGZhou

bodiddles (changes requested):
I made some comments that should probably be changed but otherwise, looks pretty good.

shepherd/Goal.py
```diff
@@ -67,4 +78,50 @@ def apply_powerup(self, effect, alliance):
             effect - enum representing an effect to be applied to the goal
             alliance - the alliance trying to apply the effect to the goal
         """
-        pass
+        if effect == POWERUP_TYPES.TWO_X:
+            if alliance.name == CONSTANTS.BLUE:
+                if self.blue_two_x_timer.is_running():
```
brandonxxlee: I feel like these blocks can be added into a helper function. It might be helpful to create a map from constant to object (e.g. POWERUP_TYPES.TWO_X and CONSTANT.BLUE to self.blue_two_x_timer somehow)

shepherd/Goal.py
```diff
@@ -67,4 +78,50 @@ def apply_powerup(self, effect, alliance):
             effect - enum representing an effect to be applied to the goal
             alliance - the alliance trying to apply the effect to the goal
         """
-        pass
+        if effect == POWERUP_TYPES.TWO_X:
+            if alliance.name == CONSTANTS.BLUE:
```
brandonxxlee: I think the `CONSTANTS` name space should be reserve for things with numerical values (e.g. multipliers and things). Perhaps add `CONSTANTS` into a new class.

shepherd/Goal.py
```diff
@@ -67,4 +78,50 @@ def apply_powerup(self, effect, alliance):
...
+            elif alliance.name == CONSTANTS.GOLD:
+                if self.gold_zero_x_timer.is_running():
+                    return -1
+                self.gold_zero_x_timer.start_timer(
+                    CONSTANTS.ZERO_X_COOLDOWN)
+                lcm_send(CONSTANTS.LCM_TARGET_SCOREBOARD,
+                         SCOREBOARD_HEADER.POWERUPS,
+                         [self.name, alliance.name, POWERUP_TYPES.ZERO_X])
+                return 1
+            return -3
```
brandonxxlee: I think `return -3` can be generalized into just one return statement at the end.

shepherd/Utils.py
```diff
@@ -33,13 +33,18 @@ class SCOREBOARD_HEADER(Enum):
...
-    ZERO_X_COOLDOWN = "zero_x_cooldown"
-    STEAL_COOLDOWN = "steal_cooldown"
-    CODE_COOLDOWN = "code_cooldown"
-    BID_INCREASE_CONSTANT = "bid_increase_constant"
+    TWO_X_COOLDOWN = 5
+    ZERO_X_COOLDOWN = 5
+    STEAL_COOLDOWN = 5
+    CODE_COOLDOWN = 5
+    BID_INCREASE_CONSTANT = 5
+    GOLD = "gold"
```
brandonxxlee: As mentioned before, I think `CONSTANTS` should just be for things with numerical values.

shepherd/.pylintrc
```diff
@@ -1,2 +1,2 @@
 [MESSAGES CONTROL]
-disable=missing-docstring, no-self-use, too-many-instance-attributes, wildcard-import, unused-wildcard-import, too-many-arguments, too-few-public-methods
+disable=missing-docstring, no-self-use, too-many-instance-attributes, wildcard-import, unused-wildcard-import, too-many-arguments, too-few-public-methods, too-many-return-statements
```
brandonxxlee: I think we don't need to ignore `too-many-return-statements`, we should just avoid it :smiley:

brandonxxlee (commented):
Left some changes, some kind of just depend on Sam's preferences as well.

shepherd/Goal.py
JHGZhou: scoring spelled wrong :)

shepherd/Goal.py
JHGZhou: -1 to boolean logic

shepherd/Goal.py
JHGZhou: Possibly use an `ENUMS` class for strings an `CONSTANTS` for numerical values

shepherd/Alliance.py
```diff
@@ -1,6 +1,6 @@
 from Utils import *
 from Timer import *
-
+from LCM import *
 
 class Alliance:
     """This is the Alliance class, which holds the state values used to track
```
JHGZhou, line 6: Until spelled wrong a million times in this comment :)

shepherd/Goal.py
JHGZhou: Its mostly for feedback on bidding station. Will only need 3 results: successful powerup, incorrect code, incorrect powerup application.

shepherd/Goal.py
JHGZhou: Are these timer variables not already kinda a map from a timer to the constants since they take in the constant when initialized?

JHGZhou: I do agree the blocks are a bit long and could be put into a helper function however for readability. Functions such as "check_apply(effect, alliance, timer)" or something of that kind.

shepherd/Goal.py
JHGZhou: Actually in reference to this method, there should only be two return values, successful powerup or incorrect powerup application. The check of the code will happen before this method is called and this should assume the code is already correct.

JHGZhou (changes requested):
Address the return values in apply powerup, the cleanup with helper methods in apply powerup, and the tidying up of constants vs enums. Besides that things look good!

bodiddles (approved):

JHGZhou (approved):

PR #464, opened 2017-11-18
[RUNTIME] Fix protobuf float precision

baby-bell:
Timestamps are being truncated to 32-bit floats instead of 64-bit doubles, which causes accuracy issues. This PR fixes the protobuf so that doesn't happen.


bodiddles (approved):
lgtm

PR #470, opened 2017-12-02
[SHEPHERD] Code generation implemented

SilentCtrl:
Created a class that can generate a code, give unique solutions to each RFID and check student solutions.

Implemented staff solutions to each software challenge function (next power, double caesar cipher, most common digit, valid isbn ten, simd four square, silly base two, smallest prime factor, reverse digits)


shepherd/Codegen.py
```diff
@@ -0,0 +1,368 @@
+import random
+
+class Codegen:
+
+    def __init__(self, rfids, rand_seed=None):
```
brandonxxlee: This might be mentioned somewhere, but do you have requirements on how long/large the RFIDs taken in should be.

shepherd/Codegen.py
```diff
@@ -0,0 +1,368 @@
...
+        '''
+        Manages the generated code and rfid-answer pairs.
+        `rfids` must be an iterable of RFID ints.
+        Duplicate RFIDs in the provided iterable are ignored.
+        '''
+
+        # Seed our local RNG
+        self._random = random.Random(rand_seed)
+        rfids = set(rfids) # Ignore duplicates
```
brandonxxlee: I actually don't think we need to handle cases where we have duplicate RFIDs (if anything, if there exist duplicates we've already screwed up). A set might still be the easiest thing to use, but is not strictly necessary.

shepherd/Codegen.py
```diff
@@ -0,0 +1,368 @@
...
+            for rfid in rfids:
+                solution = staff_decode(self._code, rfid)
+                if solution in self._ans_to_rfid:
+                    duplicate_answers = True
+                    break
+                self._ans_to_rfid[solution] = rfid
+                self._rfid_to_ans[rfid] = solution
+            if not duplicate_answers:
+                break
+        assert len(self._rfid_to_ans) == len(self._ans_to_rfid)
```
brandonxxlee: What behavior do you want if these lengths actually didn't match up.

shepherd/Codegen.py
```diff
@@ -0,0 +1,368 @@
...
+        '''
+        return self._ans_to_rfid.get(solution, None)
+
+    def get_code(self):
+        '''
+        Get the challenge code to be sent to students
+        '''
+        return self._code
+
+    def get_solution(self, rfid):
```
brandonxxlee: Can you clarify to me again what the use case of this function is?

shepherd/Codegen.py
```diff
@@ -0,0 +1,368 @@
...
+
+# Various ideas for functions for the students
+# Included are example student implementations
+# Of course, the example solutions are optimized
+# for pedagogical simplicity, not runtime speed or memory
+
+# Functions presented in no particular order of difficulty
+# Both inputs and outputs guaranteed to be nonnegative integers
+
+def next_power(num):
```
brandonxxlee, line 300: Are these functions going to be included in the Shepherd code base? I presumes yes, because it allows us to generate/check solutions. If not, these should be included as a separate file as "reference solutions". This perhaps could also just be precalculated and stored into a JSON or something, so that on the field these functions don't have to be run again (i.e. PSEUDOrandom). I think a next thing to consider is what the student skeleton code looks like.

brandonxxlee: Actually all these functions are probably really fast to run, so don't need to precompute.

brandonxxlee (changes requested):
Great job on this feature! Honestly no one had any clue on how this would look at first, but ya'll have done an amazing job designing not only the questions but also the framework around it. I left some general comments. I think the big next step is to incorporate this into the rest of Shepherd, the game, and student code.

brandonxxlee:
Also you should probably clean up the PR so that it only includes the file that you changed. Looks like you accidentally pulled in master.

Naftoreiclag:
Closing pull request, as this commit branches from too stale of a copy of shepherd/master. Will reopen with changes properly merged in from shepherd/master.

PR #471, opened 2017-12-02
[RUNTIME] Created new mappings for studentAPI

greenkee:
Allowed for new studentAPI commands which do not make use of objects
in order to invoke them (e.g. get_gamepad_value instead of
Gamepad.get_value). Responding to feedback from edu.


baby-bell (approved):
Looks good to me. Might want to expand on the tests later with some virtual devices.

PR #472, opened 2017-12-08
Improve Usability

scottnuma:
- implement arcade mode drive to be more guest-friendly
- update IDs for yogibears and servo controllers
- increase wave and high five speed


scottnuma:
This software is currently running on Staff Curry

brandonxxlee (approved):
lgtm, I'll assume UIDs are all correct.

PR #477, opened 2018-01-18
[DEVOPS] offline api

rmandla:
put robot api documentation on the beaglebones


jonathan-j-lee (approved):
LGTM

DevOps/frankfurter/scripts/update/create_update.sh
```diff
@@ -23,6 +25,17 @@ fi
 mkdir -p $TMP_DIR
 echo "Done."
 
+if [ -d "$DOC_DIR" ]; then
+    echo "Cleaning old documentation ... "
+    rm -rf $DOC_DIR/*
```
kevinji, line 30: These `$VAR` variables should be quoted; otherwise, the script will break when paths with spaces are used.

jonathan-j-lee: Thanks for the catch. Will correct this.

PR #482, opened 2018-01-20
Toggle Console Editor Resize

ArjunSridhar1:
The editor was not resizing correctly when the console was toggled, raised or lowered. These issues are now fixed.


kevinmasd (approved):
Seems fine.

PR #483, opened 2018-01-20
Updating Dawn Dependencies and Complying with new React-Bootstrap Panel API

kevinmasd:
* Getting the Dependencies Up to Date before we get too deep this season
* Updates to React-Bootstrap means some things are gone, some things are augmented
  * `Panel` API uses `<Panel.Heading>` instead of `header` tag
  * `Accordion` moved into `Panel`


ArjunSridhar1 (approved):
Ran it locally and it looks good. The drop downs seem to work.

PR #484, opened 2018-01-21
Fix Dawn Build System; Partial Revert to #414

kevinmasd:
* Clear case of not fully testing/reviewing changes
* #414 tried to move from shell commands to an only-JS implementation. However, the API is not exactly what we thought it was, so the system created empty zip files
* Revert to using shell and non-async commands, called by `execSync`. The heavy lifting is already done by the packager and zipper, so the benefits of making the system non-synchronous or only in JS is not that much actually
* Also updates Electron version, due to a notice of a security warning.


dawn/package.json
```diff
@@ -40,7 +40,7 @@
     "better-npm-run": "0.1.0",
     "chai": "4.1.2",
     "electron-devtools-installer": "2.2.1",
-    "electron-packager": "9.1.0",
+    "electron-packager": "^10.1.1",
```
ArjunSridhar1: why is there a ^?

kevinmasd: https://yarnpkg.com/en/docs/dependency-versions#toc-caret-ranges
(e.g. Keeps things relatively stable)

baby-bell (approved):
Looks alright to me. Is there a reason that you update dependencies in addition to fixing builds?

kevinmasd:
Apparently some people get a notification when packages are known to be vulnerable; might as well attach this update on a relevant change-set.

PR #485, opened 2018-01-21
Dawn/offline api

ArjunSridhar1:
Offline API accessible through the help menu.


kevinmasd (approved):
There are some minor spacing oddities, and the fact it's available in the Help Menu means we should teach people that it exists (e.g. see auto-complete). But these things can be addressed later.

Issue #288, opened 2017-03-23
[DAWN] Dragging a file into Dawn should load it into the editor


kevinji:
Currently, doing so replaces the entire editor with the file, with no way to get back to the editor.

kevinmasd:
For anyone working on this, [dawn/drag_and_drop](https://github.com/pioneers/PieCentral/tree/dawn/drag_and_drop) is a good place to start

PR #487, opened 2018-01-21
Implement Drag and Drop in Dawn

kevinmasd:
* Implement #288 
* Basically a short-cut open operation
* Also fixes bug of replacing Electron window with text, forcing a refresh


ArjunSridhar1 (approved):
Nice job!

PR #488, opened 2018-01-24
Fixing Language Stats Breakdown

kevinmasd:
* PiECentral is currently described as 30% C++, 20.6% CSS, 14% Python, 11.9% JS
  * In reality, the first two percentages are inflated due to large libraries included
* Linguist, the project that provides this breakdown, allows overrides (`linguist-vendored`)
  * This makes things like searching using the Github interface easier


ArjunSridhar1 (approved):
I guess PiE is just software is C++ and CSS.

PR #492, opened 2018-01-25
[SHEPHERD] Code generation implemented (for real for real this time) (?)

Naftoreiclag:
Codegen.py now implemented. Example usage:
```python

### DEPRECATED ###

from Codegen import Codegen

# RIFD, list and numbers can be any non-negative size
rfids = [1234, 2345, 3456, 4567, 5678, 6789]
codegen = Codegen(rfids)
challenge = codegen.generate_challenge(6) # Generate 6-digit-long codes

code = challenge.get_code()

# Send this to the students somehow...
send_code_to_student(code)

# We receive an answer from the students somehow...
answer = get_answer_from_student()

# Check if the student decoded the rfid correctly, and which one
report = challenge.check_rfid_answer(answer)

if report == -1:
    print("Student decoded nothing or decoded incorrectly")
else:
    print("Student decoded RFID " + str(rfids[report]))

# We can continue to generate more challenge codes as needed:
challenge2 = codegen.generate_challenge(20) # Generate 20-digit-long codes
```


shepherd/Codegen.py
```diff
@@ -0,0 +1,688 @@
...
+
+        if len(set(self._rfids)) != len(rfids):
+            raise RuntimeError('The RFID list cannot contain duplicates!')
+
+        # Generate function-related data
+        self._decode_precompute = staff_decode_precompute(self._rfids)
+        # True iff the given function is bijective over the given RFID domain
+        self._is_bijective = {}
+
+        # Apply every function to every rfid. This allows us to
```
brandonxxlee: Sentence incomplete

brandonxxlee (dismissed):
lgtm, make sure to test it too

shepherd/Codegen.py
Naftoreiclag: Fixed:
```python
# Apply every function to every rfid. This precomputation step
# allows us to memoize the potentially expensive example solutions for the
# student functions.
```

SilentCtrl:
Codegen.py is hopefully finished for good now. 

We made sure that the solutions are 4 digits with each digit ranging from 1-5. 

We also wrote the get_original_codes and get_new_code functions, both of which return the rfids, codes, and solutions relevant to the original function call.

brandonxxlee (approved):
lgtm, just want to make sure Sam looks at it too

shepherd/Codegen.py
```diff
@@ -0,0 +1,771 @@
...
+        digit = code % 10
+        code //= 10
+
+        if precompute is None:
+            result = func_map[digit](rfid_seed)
+        else:
+            result = precompute[digit][rfid_seed]
+
+        output += str(result)
+    output = int(output) % 10000
```
brandonxxlee: Actually I think we don't want to mod this number by 10000. This would mean only the least significant digits would be used, so the first few function calls could be useless. Am pretty sure actually the random.seed input can be like any number.

brandonxxlee (changes requested):

shepherd/Codegen.py
```diff
@@ -0,0 +1,771 @@
...
+    filter_map and only if filter_map[key] evaluates to True
+    '''
+
+    retval = {}
+
+    for key in filter_map:
+        if filter_map[key]:
+            retval[key] = data_map[key]
+
+    return retval
```
kevinji: This can be condensed into a single expression:
```py
return { key: val
         for key, val in data_map.items()
         if filter_map.get(key, False) }
```

Or if you want to go through `filter_map`:
```py
return { key: data_map[key]
         for key, to_keep in filter_map.items()
         if to_keep }
```

brandonxxlee (approved):

JHGZhou (approved):
Some things are kinda jank, and we may want to change up how this challenge factory works, but looks fine to me and will approve now

PR #498, opened 2018-01-30
[HIBIKE] Sensor testing interface improvements

baby-bell:
This PR allows writing to float/double parameters using an intuitive slider bar instead of just typing into a text box. Additionally, parameters are now displayed in alphabetical order, as opposed to whatever order they came out of the dictionary before.


tingtingchiang (changes requested):
Starting up sensor_testing leaves all the values on the page as 0. Can we figure out what the values are and set them from the start?
Also, adding/removing devices involves a restart. That's fine but seems like a shame with hotplugging enabled for hibike itself.

tingtingchiang:
Changes work on Linux but not on Mac. This is ok, it's still useable, you just need to restart the process.
Brose has opened a low priority issue to fix Macs.

tingtingchiang (approved):

PR #508, opened 2018-02-08
[RUNTIME] Shepherd Integration 

brandonxxlee:
Adds tests for the decode API function.


baby-bell (changes requested):
I don't really like the way we are copy-pasting functions, and it will probably bite us someday, but there's not much we can do for now. I'll wait until Sam takes a look to approve, otherwise LGTM.

JHGZhou (approved):
I don't fully understand what the arrays in stateManager.py represent (the ones that seem to start with dummy values of 1, 5551, and 1) but I can't really comment on that part very much. Besides that, looks good to me and does what we need it to do

runtime/studentAPI.py
```diff
@@ -280,6 +281,76 @@ def hibike_write_value(self, uid, params):
...
+        func_map = {}
+        func_map[0] = identity
+        func_map[1] = next_power
+        func_map[2] = reverse_digits
+        func_map[3] = compose_funcs(smallest_prime_fact, limit_input_to(1000000))
+        func_map[4] = double_caesar_cipher
+        func_map[5] = silly_base_two
+        func_map[6] = most_common_digit
+        func_map[7] = valid_isbn_ten
+        func_map[8] = simd_four_square
```
kevinji: This could just use a normal dict literal; i.e.
```py
func_map = {
    0: identity,
    1: next_power,
    # etc.
}
```

runtime/studentAPI.py
```diff
@@ -280,6 +281,76 @@ def hibike_write_value(self, uid, params):
...
+                return input_val
+            return retval
+
+        def compose_funcs(func_a, func_b):
+            '''
+            Composes two single-input functions together, A(B(x))
+            '''
+            def composed(input_x):
+                return func_a(func_b(input_x))
+            return composed
```
kevinji: You can shorten this with a lambda:
```py
return lambda input_x: func_a(func_b(input_x))
```

runtime/starter_student_code.py
```diff
@@ -0,0 +1,42 @@
...
+	pass
+
+def valid_isbn_ten(num):
+	pass
+
+def simd_four_square(num):
+	pass
+
+def double_caesar_cipher(key):
+	pass
```
kevinji: These indents should be four spaces instead of tabs.

kevinji (commented):
Minor stylistic comments.

runtime/studentAPI.py
baby-bell: Better yet, an array, since these are integer keys

baby-bell (approved):

PR #514, opened 2018-02-10
[Shepherd] LCM API implemented with dictionaries and JSON

charleshong3:
Added address for LCM objects to use, so LCM works over ethernet.
Uses python dictionaries as input and output for functions.
Uses JSON formatted string to send and receive over LCM.


shepherd/LCM.py
```diff
@@ -1,26 +1,23 @@
 import threading
-import lcm # pylint: disable=import-error
+import json
+import lcm
+
+LCM_address = 'udpm://239.255.76.68:7667?ttl=2'
```
JHGZhou: when was this added and is it necessary?

shepherd/LCM.py
```diff
@@ -32,13 +29,10 @@ def run():
     rec_thread.run = run
     rec_thread.start()
 
-def lcm_send(target_channel, header, *args):
+def lcm_send(target_channel, header, dic):
```
JHGZhou: might be a bit confusing, but api should support not having the `dic` argument filled in. in that case it should just send an empty dictionary

shepherd/LCM.py
```diff
@@ -1,26 +1,23 @@
 import threading
-import lcm # pylint: disable=import-error
+import json
+import lcm
```
JHGZhou: you need to add the pylint disable from before

JHGZhou (changes requested):
made 3 comments, should be quick resolutions

shepherd/LCM.py
charleshong3: You need an address to send messages over ethernet/through a switch. If these methods aren't used for that then it isn't necessary.

PR #530, opened 2018-02-24
[SHEPHERD] shepherd send info to sensors

JHGZhou:


shepherd/Goal.py
```diff
@@ -59,7 +59,7 @@ def reset(self):
                  {"goal" : self.name, "alliance" : None})
         lcm_send(LCM_TARGETS.SCOREBOARD, SCOREBOARD_HEADER.BID_AMOUNT,
                  {"goal" : self.name, "alliance" : self.current_bid_team.name, "bid"
-                  : self.current_bid})
+                  : self.current_bid + CONSTANTS.BID_INCREASE_CONSTANT})
```
brandonxxlee, line 62: I feel like we should document somewhere why we do this.

brandonxxlee (approved):
lgtm

PR #531, opened 2018-02-24
[HIBIKE] Use a watchdog timer in YogiBear firmware

baby-bell:
The YogiBear firmware might have been crashing before, causing
issues with hotplugging. This commit makes use of the watchdog
timer on the Arduino to automatically reset it if the main
loop doesn't get called enough. Even if this isn't the cause of
the issue, it's still probably a good idea for reliability's sake.


JHGZhou (approved):
Looks good to me

PR #533, opened 2018-02-28
Dawn NameEdit Removal and Snippet Error Fix

kevinmasd:
* NameEdit is pretty dead at this point, so might as well remove this dead code
  * Removing `riek` also reduces wording in `yarn` calls
* Snippets were enabled, but the file containing this things for it is missing and thus throw an exception on Dawn load. This is to be removed.


ArjunSridhar1 (approved):
RIP NameEdit

PR #540, opened 2018-03-01
[HIBIKE] Make RFID id Field 24 bits

### LINKED ISSUE #535: [HIBIKE] Read more digits from the RFID card
It needs to be hard for students to brute force RFID codes, so we should expand the number of digits we read.

rohityelamanchili:
Closes #535


baby-bell (approved):
LGTM, tested with current Dawn.

PR #545, opened 2018-03-05
[DAWN] Upgrade to webpack 4

kevinji:
This isn't tested too much, other than the fact that the build works and Dawn runs.


kevinmasd (approved):
Looks okay

Issue #539, opened 2018-03-01
Clean up scripts to follow style guide


jonathan-j-lee:
Scripts should comply with the [Google shell style guide](https://google.github.io/styleguide/shell.xml).

baby-bell:
Closed by #546

jonathan-j-lee:
Close enough.

Issue #489, opened 2018-01-25
Add kernel module blacklist


jonathan-j-lee:
Avoid loading unused modules that interface with HDMI, LCD, etc.

PR #546, opened 2018-03-06
Script improvements

jonathan-j-lee:
* Added `systemd` `install.service` unit to implement autonomous deployment, even with reboots
* Added blacklist of unused kernel modules (e.g. HDMI, `snd*` for sound, etc.)
* Added sanity (file path) checks to `update.service` and `runtime.service` units
* `create_update` now uses `/tmp` (typically mounted in memory) to build a tarball, not the CWD on disk
* Used `ssh-copy-id` to eliminate need to re-enter SSH passwords
* Removed unused config files and dependencies (e.g. `memcached`, `sshd`)
* Disabled protobuf compilation, by default, when creating updates
* Cleaned up script style and syntax

Addresses #489, #539.


DevOps/frankfurter/scripts/install/start_install
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
```
kevinji: Probably want to make this `/bin/bash` for consistency.

brandonxxlee (approved):

PR #547, opened 2018-03-09
Dawn May Dependency Upgrades

kevinmasd:
* Batch Updating Dependencies
* `ws` removed as `Socket.IO` has replaced it for Field Control usage


ArjunSridhar1 (changes requested):
The build on travis is broken cause of 404 error on some url.

kevinmasd:
End of 2018 Season.

PR #548, opened 2018-03-09
Add code to test motors without controller

scottnuma:
Ideally this code could be released to students because it is of so limited use. It also gives students code that we have written that is bug free --> they can know it's a hardware problem not a their code problem.


brandonxxlee (approved):
sure, lgtm

PR #551, opened 2018-03-10
Show Special Values for Game

kevinmasd:
* For the Game, there are certain values to be displayed. Currently, they are sent as LimitSwitches (probably default); now they are modified as "Game Values" (Dawn sorts alphabetically)
  * If Game Value, display name; otherwise keep uid until further notice
* For better debug, we first stringify what we get from Runtime. (Otherwise looks like `[Object object]`)


bodiddles (approved):
lgtm

PR #554, opened 2018-03-11
Harmonized File Upload/Download; Fix Socket.IO On-Close

kevinmasd:
* Due to the functions being made at different times, File Upload and Download aren't symmetrical, which may lead to differences using them
  * Also getting rid of notification thing (since we don't really actually use it for anything)
  * Also shows modal window on finish
* Socket.IO also doesn't close properly when Dawn ends; fixed


brandonxxlee (approved):
took brief look over it lgtm, just make sure to test

PR #569, opened 2018-03-21
[RUNTIME] Remove solar scramble

brandonxxlee:


bodiddles (approved):
Security through obscurity

PR #571, opened 2018-04-01
[RUNTIME] Add lint to prevent importing studentCode

baby-bell:
This PR adds a lint to prevent `studentCode` from being imported anywhere but `run_student_code` and `runtime_test`. Hopefully, this will prevent issues with reloading.


runtime/lints.py
```diff
@@ -0,0 +1,47 @@
...
+    def leave_functiondef(self, node):
+        self._stack.pop()
+
+    def visit_importfrom(self, node):
+        if node.modname == 'studentCode':
+            if not self._ok_to_import():
+                self.add_message('no-import-student-code', node=node)
+
+    def visit_import(self, node):
+        mods = list(map(lambda x: x[0], node.names))
```
kevinji: This is probably cleaner as a list comprehension; i.e.
```python
mods = [name[0] for name in node.names]
```

jonathan-j-lee (approved):
LGTM

PR #575, opened 2018-04-12
[DEVOPS] usb hotplugging

rmandla:
alias usb hotplugging for testing purposes only. Currently does not work when the beaglebone is plugged into an external power source.


jonathan-j-lee (approved):
LGTM

PR #577, opened 2018-04-13
[SHEPHERD] Integrates Bidding Station and Linebreak Sensors into Shepherd

bodiddles:
Adds automatic discovery and enumeration of bidding stations and linebreak sensors.

Modifies sensors to attach an identification header to every message as well as send a heartbeat message for discoverability by Sensors.py.

Modifies Sensors.py to account for this change in protocol as well.


vkhuang (dismissed):
sgtm

JHGZhou (dismissed):
Sure for scrimmage

JHGZhou (approved):
Is fine after style changes

PR #585, opened 2018-04-23
[HIBIKE] Slow down hotplug polling period to 5 seconds

baby-bell:
It seems as though robots with a lot of devices are having performance issues. This should mitigate them.


rohityelamanchili (approved):

baby-bell:
This doesn't appear to make a substantial difference in processing power, so I'm closing

PR #586, opened 2018-04-26
Dawn 2018 Final Competition Changes

kevinmasd:
* Bump Version to 2018.2.1
* Heartbeat message from Field Control properly used to toggle +/- on version field
  * Does not provide watchdog semantics
* Console Button turns red if there are messages that are unread


brandonxxlee (approved):
lgtm, can you get a built version released only internally.

kevinmasd:
Everything's de facto internal until we publicly release.

PR #590, opened 2018-04-29
[RUNTIME] Fix State Manager data race related to RFID

baby-bell:
For some reason, State Manager sends data directly to student code by default, which can cause data races depending on the state of the student code process. This PR unsets that flag when sending RFID codes to State Manager.


brandonxxlee (approved):
lgtm

PR #606, opened 2018-07-21
Frankfurter Docker image

jonathan-j-lee:


baby-bell:
LGTM. Is this intended to be used on BBB?

baby-bell (approved):

jonathan-j-lee:
As of now, no. If I recall correctly, a compressed image is ~50MB, which is far too heavy.

The main purpose of this image is to simulate the performance constraints of the BBB, with the goal of automated performance testing.

PR #621, opened 2018-09-17
[RUNTIME] Fix StateManager crash, improve error

baby-bell:
This PR fixes a bug that caused state manager to crash when no student code process was running. In addition, when a crash occurs, contextual information related to the crash is now provided.


SilentCtrl (approved):
These changes seem to be helpful

PR #657, opened 2018-11-29
completed text size input button

Funplings:


lawrencechen14:
- Add a tooltip onto the textsize dropdown so that students will know that it is meant to modify text sizes.
- Trim the dropdown so that it does not take up as much whitespace and so that it will appear flush with the other buttons.

lawrencechen14 (approved):
Seems good

PR #658, opened 2018-11-29
Shepherd/launchbuttons

FIREdog5:


jonathan-j-lee:
Pull from `master` to get the Dawn version fix. Note that you have a bunch of conflicting files you'll need to fix. @FIREdog5

shepherd/Code.py
```diff
@@ -0,0 +1,45 @@
...
+
+def assign_code_solution():
+    '''
+    Generate 16 codes and create a code_solution dictionary
+    '''
+    global codes
+    global solutions
+    global code_solution
+    global code_effect
+    global effect_list
```
jonathan-j-lee: Most of these variables don't need `global`. `global` is needed for changing what these variables point to (e.g. if you wanted `codes = 1`), but is not needed for internal mutations (e.g. appending elements to lists or setting key-value pairs).

shepherd/Code.py
```diff
@@ -0,0 +1,45 @@
...
+    for i in range(16):
+        code_solution[codes[i]] = solutions[i]
+    return code_solution
+
+def assign_code_effect():
+    '''
+    Assign each code to a random effect
+    '''
+    for i in range(16):
+        code_effect[codes[i]] = effect_list[random.randint(0, len(effect_list))]
```
jonathan-j-lee: Use `random.choice(effect_list)`.

shepherd/Code.py
```diff
@@ -0,0 +1,45 @@
+import random
+from Utils import *
+
+codes = []
+solutions = []
+code_solution = {}
+code_effect = {}
+effect_list = []
+effect_list = list(EFFECTS.ALL_EFFECTS)
```
jonathan-j-lee, line 8: Is there a reason this state and code isn't wrapped in a class?

shepherd/Code.py
```diff
@@ -0,0 +1,45 @@
+import random
+from Utils import *
+
+codes = []
+solutions = []
+code_solution = {}
+code_effect = {}
+effect_list = []
```
jonathan-j-lee: Doesn't seem to be used.

jonathan-j-lee (changes requested):
Comments so far.

shepherd/Code.py
FIREdog5: Done.

jonathan-j-lee (approved):

PR #666, opened 2019-02-02
Hibike/asyncio timer

benliao1:
Logic in runtime.py completed for the 15-second timeout extension

Large comment indicating where we need to write a statement that returns whether the robot has the perk or not. Protobufs have been scrapped and we need to figure out how the signal that the robot has the perk will be sent to runtime. Once that's been figured out, we can fill in that line of code.


jonathan-j-lee:
Closed. Perk timers were handled by Shepherd.

PR #672, opened 2019-03-09
Update scripts for RPis

jonathan-j-lee:


SilentCtrl (approved):
Some firmware commits are in here, but otherwise everything seems ok

PR #693, opened 2019-04-18
Master robot designation fixed

ewc340:
Change master robot designation so that it displays on multiple computers


lawrencechen14 (approved):
Looks good.

PR #694, opened 2019-04-18
Remove old name references

lawrencechen14:
- Removing old yogi bear motor references


jonathan-j-lee:
Please keep this in; per our discussion, Dawn should be backwards-compatible.

PR #733, opened 2020-02-13
Shepherd/recipemanager

ochan1:
Completion of Recipe Manager

Changes to Utils.py and RecipeManager.py


shepherd/Utils.py
```diff
@@ -139,3 +139,54 @@ class STATE():
...
+                    RECIPIES.CHEESEBURGER: {Ingredients:[INGREDIENT.BREAD, \
+                                            INGREDIENT.MEAT, INGREDIENT.CHEESE], Cooked:1}
+                  }
+    VERY_HARD_RECIPE = {RECIPIES.SHEPHERDS_PIE: {Ingredients:[INGREDIENT.MEAT,\
+                                                INGREDIENT.VEGETABLES,\
+                                                INGREDIENT.CHEESE], Cooked:2}}
+
+class TELEOP_RECIPE_MAP():
+    MAPPING = {0:[1, 0, 0, 0], 1:[0.1, 0.8, 0.1, 0], 2:[0.1, 0.1, 0.8, 0], 3:[0.1, 0.2, 0.3, 0.4]}
+    # 0: Easy, 1: Medium, 2: Hard, 3: Very Hard
```
FIREdog5: This comment seems misleading.. should read 0: first recipe, 1: second recipe... the indices in each of the lists are what indicates the difficulty like in this comment.

shepherd/Utils.py
```diff
@@ -139,3 +139,54 @@ class STATE():
...
+    GRILLED_CHEESE = "Grilled Cheese"
+    JALAPENO_POPPERZ = "Jalapeno Popperz"
+    CHEESEBURGER = "Cheeseburger"
+    SHEPHERDS_PIE = "Sherpherd's Pie"
+    CHEESE_AND_CRACKERS = "Cheese and Crackers"
+
+class RECIPE_PROBABILITIES():
+    Ingredients = "Ingredients"
+    Cooked = "Cooked"
+    RECIPE_PROBABILITIES = {0: [], 1: [], 2:[], 3:[]}
```
FIREdog5: This seems like the same variable as the "mappings" variable in teleop_recipe_map. Perhaps that class should be removed and this variable be used instead

shepherd/Utils.py
```diff
@@ -139,3 +139,54 @@ class STATE():
...
+class INGREDIENT():
+    BREAD = "Bread"
+    VEGETABLES = "Vegetables"
+    MEAT = "Meat"
+    CHEESE = "Cheese"
+    # COOKED_MEAT = "COOKED " + MEAT
+    # COOKED_BREAD = "COOKED " + BREAD
+    # COOKED_CHEESE = "COOKED " + CHEESE
+
+class RECIPIES():
```
kaipinryankoh: The word "recipes" doesn't have an 'i', BIG SAD :'(

shepherd/RecipeManager.py
```diff
@@ -0,0 +1,77 @@
+from Utils import *
+
+
+class RecipeManager():
+
+    master_recipes = [RECIPE_PROBABILITIES.EASY_RECIPE, RECIPE_PROBABILITIES.MEDIUM_RECIPE, \
```
FIREdog5: I don't understand why this is a list of references to an empty list... it does give me a good idea of how to manage the recipes tho.

FIREdog5 (changes requested):
Please see the comments for changes, and talk to me in work session cuz i have another change that I don't want to try to explain via github comment

shepherd/RecipeManager.py
```diff
@@ -0,0 +1,77 @@
...
+                      RECIPE_PROBABILITIES.HARD_RECIPE, RECIPE_PROBABILITIES.VERY_HARD_RECIPE]
+    gold_recipes = []
+    blue_recipes = []
+    gold_recipes_completed = 0
+    blue_recipes_completed = 0
+
+    def generate_recipes(self):
```
FIREdog5: self should not be an argument when the class is not going to be instantiated.

shepherd/Utils.py
ochan1: Shoot
I thought it would be funny with "PIE" in the name
(nah, it's the reason why I'm not a Humanities major. My engrish is horribl)

shepherd/RecipeManager.py
ochan1: The idea is to make it easier for us to load and reload the list by having a list that we can read from without the programmer looking back and forth
It's to make our lives and future people lives easier

Also, this is not empty (or I believe it shouldn't be)

shepherd/RecipeManager.py
ochan1: How can we access the class variables above then?

shepherd/RecipeManager.py
```diff
@@ -0,0 +1,77 @@
...
+            recipes = self.blue_recipes
+        else:
+            recipes = self.gold_recipes
+        if len(recipes) != len(ingredients):
+            return False
+        else:
+            for item in ingredients:
+                if item not in recipes:
+                    return False
+        return True
```
kaipinryankoh: put this line inside the else statement to appease the style checker :')

PR #752, opened 2020-05-22
Y13 CODEOWNERS

jonathan-j-lee:
Changing of the guard. PMs will automatically be requested for review if their project's source is modified.

(@benliao1: I hope you're OK with reviewing for DevOps.)


benliao1 (approved):
These look good! I think Charles should also review them, but I guess I can just request him as well to review changes to DevOps.

